PR #562, opened 2025-08-24
Codegen/jov 119 loading replace image based spinner with svgcss skeletons

itstimwhite:


itstimwhite:
@claude fix ⚠️ Issues Found
1. Component Duplication Concern

LoadingSpinner exists in both components/atoms/ and components/ui/
components/atoms/LoadingSpinner.tsx:11-132 has debounce logic and xs size
components/ui/LoadingSpinner.tsx:10-102 lacks debounce and xs size
This creates API inconsistency and potential confusion
2. Theme Handling Inconsistency

components/ui/Spinner.tsx:18 uses useTheme() hook properly
components/atoms/LoadingSpinner.tsx and components/ui/LoadingSpinner.tsx use static variant logic
Different color schemes between components (indigo vs gray/white)
3. Animation Performance

SVG <animateTransform> elements may cause performance issues at scale
Consider CSS-only animations for better performance
4. Missing PropTypes/Validation

No runtime prop validation for skeleton component props
height/width props accept any string without validation

PR #8637, opened 2026-05-13
fix(onboarding): wrap /start SSR in try/catch with redirect fallback

itstimwhite:
## Summary

- Wraps `getOrMintOnboardingSessionId()` in `/start/page.tsx` in try/catch with `captureError` + `redirect('/signup')` fallback
- Resolves the prod 500 that renders as a 404 (audit finding #7, P0) — top-of-funnel anon onboarding entry is restored
- Root cause: `SESSION_SECRET` is marked `.optional()` in the env schema; when absent (or < 32 chars), `getSessionSecret()` throws; the unhandled throw causes Next.js to render a 500, which the error boundary chain serves as the `not-found.tsx` component

## Root cause detail

`lib/onboarding/session.ts` L31–38: `getSessionSecret()` explicitly throws when `SESSION_SECRET` is falsy or < 32 chars. The env schema (`env-server-schema.ts` L158) defines it as `z.string().min(32).optional()`, meaning builds without this var succeed silently. On production, if `SESSION_SECRET` was never provisioned in Doppler (or was provisioned after deploy but not yet reloaded), every `/start` request throws → 500.

The fix does not mask the root cause — `captureError` fires to Sentry so the missing env var surfaces as an actionable alert, and the redirect keeps users moving toward signup rather than hitting a dead end.

## Test plan
- [x] Vitest: `StartPage — session error fallback` (2 tests) — error path calls `captureError` + `redirect('/signup')`; happy path renders without redirect
- [x] Typecheck clean (`pnpm --filter @jovie/web run typecheck`)
- [x] Biome clean (`pnpm biome check --write` on both files)
- [ ] `curl https://staging.jov.ie/start` post-merge — should return redirect, not 500

## Files changed
- `apps/web/app/(dynamic)/start/page.tsx` — try/catch + captureError + redirect fallback
- `apps/web/app/(dynamic)/start/page.test.ts` — new unit tests for error + happy paths

Source: 2026-05-13 audit, finding #7 (P0).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during onboarding session creation—users experiencing issues with session setup are now gracefully redirected to signup instead of encountering an error.

* **Tests**
  * Added test coverage for onboarding session error scenarios and successful session creation paths.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JovieInc/Jovie/pull/8637)

### DIFF
--- a/apps/web/app/(dynamic)/start/page.test.ts
+++ b/apps/web/app/(dynamic)/start/page.test.ts
@@ -1,0 +1,75 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Hoisted mocks must be declared before any imports that use them.
+const { getOrMintOnboardingSessionIdMock, captureErrorMock, redirectMock } =
+  vi.hoisted(() => ({
+    getOrMintOnboardingSessionIdMock: vi.fn(),
+    captureErrorMock: vi.fn(),
+    // In real Next.js, redirect() throws a NEXT_REDIRECT error so execution
+    // stops. Simulate that behaviour so the catch branch returns cleanly.
+    redirectMock: vi.fn(() => {
+      throw new Error('NEXT_REDIRECT');
+    }),
+  }));
+
+vi.mock('@/lib/onboarding/session', () => ({
+  getOrMintOnboardingSessionId: getOrMintOnboardingSessionIdMock,
+}));
+
+vi.mock('@/lib/error-tracking', () => ({
+  captureError: captureErrorMock,
+}));
+
+vi.mock('next/navigation', () => ({
+  redirect: redirectMock,
+}));
+
+// OnboardingShell is a UI component we don't need to render in this test.
+vi.mock('@/components/features/onboarding/OnboardingShell', () => ({
+  OnboardingShell: () => null,
+}));
+
+import { APP_ROUTES } from '@/constants/routes';
+
+import StartPage from './page';
+
+describe('StartPage — session error fallback', () => {
+  beforeEach(() => {
+    getOrMintOnboardingSessionIdMock.mockReset();
+    captureErrorMock.mockReset();
+    redirectMock.mockReset();
+    // Re-apply the throw so each test inherits the real-Next.js behaviour.
+    redirectMock.mockImplementation(() => {
+      throw new Error('NEXT_REDIRECT');
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


itstimwhite:
## Review Summary
**Verdict**: ✅ Approve

The PR correctly addresses a P0 production bug by gracefully handling failures in `getOrMintOnboardingSessionId()` caused by a missing/invalid `SESSION_SECRET`. The fix follows Jovie conventions, adds proper error telemetry, and redirects users to `/signup` instead of exposing a 500 error. Tests cover both the error and happy paths.

---

### Critical Issues
None.

---

### Warnings
None.

---

### Suggestions

1. **`apps/web/app/(dynamic)/start/page.tsx` – `return redirect(...)` is unreachable**  
   `redirect()` from Next.js always throws a `NEXT_REDIRECT` error. The `return` statement is dead code. While harmless, removing `return` would make the intent clearer and avoid confusion for future readers.
   ```ts
   // Consider changing
   return redirect(APP_ROUTES.SIGNUP);
   // to
   redirect(APP_ROUTES.SIGNUP);
   ```

2. **`apps/web/app/(dynamic)/start/page.test.ts` – No test for `captureError` rejection**  
   If `captureError` itself throws (e.g., network failure talking to Sentry), the catch block would exit ungracefully and a 500 could still occur. While unlikely in practice, a nested try/catch or a test verifying that the page still redirects even if `captureError` rejects would improve robustness. Since this is a P0 fix, the current coverage is sufficient; consider adding a follow-up edge-case test.

3. **`apps/web/app/(dynamic)/start/page.test.ts` – Mock `redirect` without `return` in the implementation**  
   The mock currently mimics Next.js’s throwing behaviour, which correctly forces the test to use `rejects.toThrow`. If you adopt suggestion #1, ensure the mock still throws (it does), and the test will continue to work.

---

### What Looks Good

- **Targeted fix** – Only the single failure-prone line is wrapped, with clear comments explaining the guard and the expected root cause.
- **Telemetry preserved** – `captureError` fires to Sentry with a descriptive message and route context, so operators can detect the missing env var without relying on 5xx alerts.
- **Clean test setup** – Mocks are hoisted correctly, and the test simulates Next.js’s redirect behaviour faithfully.
- **Uses existing constants** – `APP_ROUTES.SIGNUP` is already defined; no new magic strings introduced.
- **Conforms to conventions** – No emoji, no native dialogs, no decorative motion, TypeScript safe, no `any`.

itstimwhite (commented):

PR #8659, opened 2026-05-14
fix(profile): sync alert prefs across all 3 notification entry points (JOV-1986)

itstimwhite:
## Summary

Fixes [JOV-1986](https://linear.app/jovie/issue/JOV-1986) — inconsistent notification alert modes between subscribe/unsubscribe states.

### Root cause

`alertPrefs` in `ProfileInlineNotificationsCTA` was always initialized to `DEFAULT_ALERT_PREFS` (all `true`), regardless of the user's server-side preferences. This caused divergence between:
- The **notifications drawer** (showed real server prefs via `contentPrefs` prop chain from `ProfileCompactTemplate`)  
- The **inline CTA** and **subscribe drawer** flows (both initialized to hardcoded defaults)

### Fix

Two targeted changes in `ProfileInlineNotificationsCTA.tsx`:

1. **Lazy `useState` initializer for `alertPrefs`** — seeds from server `contentPreferences` at mount when the user is already subscribed, so all three entry points start from the same canonical state instead of `DEFAULT_ALERT_PREFS`.

2. **`syncPreferencesFromStatus()` call in the auto-transition effect** — handles the late-hydration case where subscription state resolves after component mount; server preferences are now merged before switching to manage mode.

`artistEmailOptIn` received the same treatment — seeds from `artistEmail?.optedIn` when subscribed rather than always `false`.

### Tests

Added 3 regression tests in `inline-notifications-cta.test.tsx` covering all three entry points:
- Overlay entry (already-subscribed fan opens manage alerts)
- Inline entry (inline CTA when already subscribed)
- Subscribe-drawer entry (hydration resolves to subscribed mid-flow)

### Verification

- 8/8 new + existing tests pass
- 475/475 profile unit tests pass  
- 0 TypeScript errors
- Biome clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Notification preferences now correctly initialize from saved server settings when managing alert preferences, ensuring users see their actual configured preferences instead of defaults.

* **Tests**
  * Added test coverage verifying notification preference synchronization across different subscription states and component entry points.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JovieInc/Jovie/pull/8659)

### DIFF
--- a/apps/web/components/features/profile/artist-notifications-cta/ProfileInlineNotificationsCTA.tsx
+++ b/apps/web/components/features/profile/artist-notifications-cta/ProfileInlineNotificationsCTA.tsx
@@ -179,8 +179,20 @@ export function ProfileInlineNotificationsCTA({
   const [nameInput, setNameInput] = useState('');
   const [birthdayInput, setBirthdayInput] = useState('');
   const [birthdayHintShown, setBirthdayHintShown] = useState(false);
-  const [alertPrefs, setAlertPrefs] = useState(DEFAULT_ALERT_PREFS);
-  const [artistEmailOptIn, setArtistEmailOptIn] = useState(false);
+  const [alertPrefs, setAlertPrefs] = useState<
+    Record<NotificationContentType, boolean>
+  >(() => {
+    // Seed from server preferences when the user is already subscribed at mount,
+    // so all three entry points (inline CTA, notifications drawer, subscribe
+    // drawer) start from the same canonical state rather than DEFAULT_ALERT_PREFS.
+    if (isSubscribed && contentPreferences) {
+      return { ...DEFAULT_ALERT_PREFS, ...contentPreferences };
+    }
+    return DEFAULT_ALERT_PREFS;
+  });
+  const [artistEmailOptIn, setArtistEmailOptIn] = useState(() =>
+    isSubscribed ? (artistEmail?.optedIn ?? false) : false
+  );
   const [canEditPreferences, setCanEditPreferences] = useState(isSubscribed);
   const subscribedEmailRef = useRef('');
   const hasAutoOpenedRef = useRef(false);
@@ -308,10 +320,21 @@ export function ProfileInlineNotificationsCTA({
       return;
     }
 
+    // Sync server preferences before transitioning to manage mode so that the
+    // preferences step shows the actual saved values, not DEFAULT_ALERT_PREFS.
+    syncPreferencesFromStatus();
     setFlowOrigin('manage');
     setCanEditPreferences(true);
     setStep('preferences');
-  }, [autoOpen, flowOrigin, isFlowOpen, isInline, isSubscribed, step]);
+  }, [
+    autoOpen,
+    flowOrigin,
+    isFlowOpen,
+    isInline,
+    isSubscribed,
+    step,
+    syncPreferencesFromStatus,
+  ]);
 
   const activeEmail = useMemo(
     () =>
--- a/apps/web/tests/unit/profile/inline-notifications-cta.test.tsx
+++ b/apps/web/tests/unit/profile/inline-notifications-cta.test.tsx
@@ -288,4 +288,128 @@ describe('ProfileInlineNotificationsCTA flow', () => {
 
     expect(container).toBeEmptyDOMElement();
   });
+
+  // JOV-1986: alert mode consistency across all three entry points.
+  // The preference labels in ProfileMobileNotificationsFlow map as:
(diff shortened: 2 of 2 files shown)


itstimwhite:
## Review Summary  

The PR correctly fixes the alert preference inconsistency by seeding `alertPrefs` from server `contentPreferences` and syncing on late hydration. Tests cover the three entry points well.  

However, there is a **critical performance concern** with the newly added `syncPreferencesFromStatus` in the `useEffect` dependency array, which could cause infinite re-renders unless it is memoized (e.g., with `useCallback`). This must be verified or fixed before merging.

---

### Critical Issues  

- **apps/web/components/features/profile/artist-notifications-cta/ProfileInlineNotificationsCTA.tsx:335**  
  `syncPreferencesFromStatus` is added to the dependency array of the `useEffect` that transitions to manage mode. If this function is not wrapped in `useCallback` (or defined outside the component), it will be a **new reference on every render**, causing the effect to run on every render. This can lead to repeated calls to `setFlowOrigin`, `setCanEditPreferences`, and `setStep`, potentially causing infinite re-renders or a broken UI.  

  **Why it matters**: Even though React will bail out of state updates if the value is the same, multiple effect executions can still cause performance degradation and unexpected behavior (e.g., multiple transitions).  

  **Suggested fix**:  
  - Confirm that `syncPreferencesFromStatus` is either a stable function (defined with `useCallback`) or that its definition is imported from outside the component.  
  - If not, wrap it in `useCallback` with appropriate dependencies (likely nothing, since it only reads component state/refs – but verify that it doesn’t cause stale closures).  

  ```ts
  const syncPreferencesFromStatus = useCallback(() => {
    // existing implementation
  }, []); // or with necessary deps
  ```

---

### Warnings  

- **apps/web/components/features/profile/artist-notifications-cta/ProfileInlineNotificationsCTA.tsx:183–191**  
  The lazy initializer depends on `isSubscribed` and `contentPreferences`. If these values are `undefined` at mount (e.g., if the hooks fetch asynchronously), the component will fall back to `DEFAULT_ALERT_PREFS` and later correct via `syncPreferencesFromStatus()`. This is intentional, but the effect that calls `syncPreferencesFromStatus` must not be blocked by other conditions. The test for the subscribe-drawer entry covers this correctly.  

  **Edge case**: If `contentPreferences` is an empty object `{}` for a truly new user, `{ ...DEFAULT_ALERT_PREFS, ...{} }` defaults to all `true`. That’s correct for a new user.  
  No action needed, but worth noting.

- **tests/unit/profile/inline-notifications-cta.test.tsx:320**  
  The test `'syncs alertPrefs from server when hydration resolves to subscribed mid-flow'` uses `rerender` after mocking new return values. It correctly validates that the switch is not checked after transition. However, the test does not verify that `syncPreferencesFromStatus()` was actually called (e.g., via spy). This is acceptable for a unit test that checks UI state, but a mock spy would provide stronger regression protection.  

  **Suggestion**: Consider adding a spy on `setAlertPrefs` or `syncPreferencesFromStatus` to ensure the function runs as expected.

---

### Suggestions  

- **Missing test case**: Add a test where `contentPreferences` is `null` or `undefined` at mount but the user is subscribed. This would validate that `alertPrefs` defaults to `DEFAULT_ALERT_PREFS` and later gets updated by `syncPreferencesFromStatus`.  

- **Comments**: The current inline comments are clear and helpful. Consider adding a brief comment explaining why `syncPreferencesFromStatus` is added to the effect deps (stability requirement) to prevent future refactoring mistakes.  

- **Type safety**: The `useState<Record<NotificationContentType, boolean>>` cast via generic is good. However, if `contentPreferences` has extra keys, the spread will include them. This is likely fine given the server response shape, but consider using a strict merge function to avoid accidental property polution.

---

### What Looks Good  

- **Root cause analysis** is spot-on: initializing `alertPrefs` to `DEFAULT_ALERT_PREFS` regardless of server state caused the inconsistency.  
- **Lazy `useState` initializer** correctly seeds from server prefs only when already subscribed, avoiding unnecessary work.  
- **Sync on late hydration** via `syncPreferencesFromStatus()` in the auto-transition effect handles the case where subscription state resolves after mount.  
- **All three entry points** (overlay, inline, subscribe-drawer) are covered by regression tests, and the test organization with nested `describe` is clear.  
- **Code is concise** and follows existing patterns (no emoji, proper imports, TypeScript).  
- **No UI regressions**: the changes are purely internal state initialization logic with no visual modifications.  
- **Commit message** follows conventional commit format, and the PR description includes clear verification metrics.  

Once the stability of `syncPreferencesFromStatus` is confirmed or fixed, this PR is ready to merge.

itstimwhite (commented):

PR #8734, opened 2026-05-16
fix(shell): handle deployment-skew server action errors gracefully (JOV-2192)

itstimwhite:
## Summary

Fixes JOV-2192: Next.js 15 deployment-skew server action errors were leaking into Sentry as noise and presenting users with a broken \"Try again\" loop.

**Root cause (two bugs):**

1. **Sentry filter never matched.** `STALE_SERVER_ACTION_PATTERN` in `sentry/config.ts` was a regex that embedded `unrecognizedactionerror:` as a prefix and tested against the exception `message` field — but that prefix only lives in the `type` field. The regex never fired, so every stale-action Sentry event slipped through as noise.

2. **ErrorBoundary offered the wrong recovery.** When a server action from a stale bundle is called, Next.js throws `UnrecognizedActionError` (client) or a plain `Error` with code E974/E975 (server). The existing `ErrorBoundary` showed \"Try again\" — which re-invokes the same missing action and loops forever. The correct recovery is a hard page reload.

**Fixes:**

- `sentry/config.ts` — replaced the broken regex with a direct `type === 'unrecognizedactionerror'` check plus message substring patterns covering both Next.js error shapes (`was not found on the server`, `failed to find server action`).
- `ErrorBoundary.tsx` — added `isDeploymentSkewError()` helper; skew errors now show \"App Updated / Reload\" with `location.reload()` instead of the retry loop. Sentry capture is also skipped for skew errors (already filtered in `beforeSend`).
- `sentry-config.test.ts` — replaced the old test (which used the wrong error shape) with three accurate regression tests covering client-side `UnrecognizedActionError`, server-side E974/E975 plain Error, and hash-in-message variants.

## Verification

```
✓ tests/unit/lib/sentry/sentry-config.test.ts — 42/42 tests pass
✓ pnpm biome check — clean (3 files)
✓ pnpm typecheck — clean
✓ Pre-commit hooks — all passed
```

## Files Changed

- `apps/web/lib/sentry/config.ts` — fix `isDeploymentTransitionError()` filter
- `apps/web/components/organisms/ErrorBoundary.tsx` — skew-aware UX + skip redundant Sentry capture
- `apps/web/tests/unit/lib/sentry/sentry-config.test.ts` — accurate regression tests

## Test Plan

- [x] Unit tests cover both Next.js error shapes (client `UnrecognizedActionError` + server E974/E975 plain Error)
- [x] Sentry `beforeSend` now correctly drops stale server action events
- [x] ErrorBoundary shows \"Reload\" CTA instead of looping \"Try again\" on deployment skew
- [x] Typecheck, lint, and pre-commit hooks all pass

## Release Notes

* **Bug Fixes**
  * Improved detection and handling of deployment-related errors with a dedicated "Reload" button for faster recovery.
  * Enhanced error filtering to reduce monitoring noise from stale server-action errors.

* **Tests**
  * Updated test coverage for server-action error detection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JovieInc/Jovie/pull/8734?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/web/components/organisms/ErrorBoundary.tsx
+++ b/apps/web/components/organisms/ErrorBoundary.tsx
@@ -14,6 +14,24 @@ interface ErrorBoundaryProps {
   readonly message?: string;
 }
 
+/**
+ * Returns true when the error is a deployment-skew server action mismatch.
+ *
+ * Next.js throws `UnrecognizedActionError` (client) or a plain `Error` with
+ * a specific message (server) when a stale client bundle calls a server action
+ * that no longer exists in the current deployment. Retrying with the same
+ * stale bundle always fails — the correct recovery is a hard page reload.
+ */
+function isDeploymentSkewError(error: Error): boolean {
+  const type = error.constructor?.name?.toLowerCase() ?? '';
+  const msg = error.message?.toLowerCase() ?? '';
+  return (
+    type === 'unrecognizedactionerror' ||
+    msg.includes('was not found on the server') ||
+    msg.includes('failed to find server action')
+  );
+}
+
 export default function ErrorBoundary({
   error,
   reset,
@@ -21,13 +39,21 @@ export default function ErrorBoundary({
   message = "We couldn't load this page. Give it another try, or head home.",
 }: ErrorBoundaryProps) {
   const router = useRouter();
+  const isSkewError = isDeploymentSkewError(error);
 
   useEffect(() => {
     console.error(`[${context} Error]`, error);
-    captureErrorInSentry(error, context.toLowerCase(), {
-      digest: error.digest,
-    });
-  }, [error, context]);
+    // Deployment-skew errors are filtered in Sentry's beforeSend — skip capture.
+    if (!isSkewError) {
+      captureErrorInSentry(error, context.toLowerCase(), {
+        digest: error.digest,
+      });
+    }
+  }, [error, context, isSkewError]);
+
+  const displayMessage = isSkewError
+    ? 'The app was just updated. Reload to continue.'
+    : message;
 
   return (
     <div
@@ -44,21 +70,33 @@ export default function ErrorBoundary({
 
         <div className='space-y-1.5'>
           <h3 className='text-sm font-medium text-secondary-token'>
-            Something went wrong
+            {isSkewError ? 'App Updated' : 'Something went wrong'}
           </h3>
-          <p className='text-app text-tertiary-token'>{message}</p>
+          <p className='text-app text-tertiary-token'>{displayMessage}</p>
         </div>
 
         <div className='flex justify-center gap-3'>
-          <Button variant='primary' size='sm' onClick={reset}>
-            Try again
-          </Button>
+          {isSkewError ? (
+            <Button
+              variant='primary'
+              size='sm'
+              onClick={() => globalThis.location.reload()}
+            >
+              Reload
+            </Button>
+          ) : (
+            <Button variant='primary' size='sm' onClick={reset}>
+              Try again
+            </Button>
+          )}
           <Button variant='outline' size='sm' onClick={() => router.push('/')}>
             Go home
           </Button>
         </div>
 
-        <ErrorDetails error={error} extraContext={{ Context: context }} />
+        {!isSkewError && (
+          <ErrorDetails error={error} extraContext={{ Context: context }} />
+        )}
       </div>
     </div>
   );
--- a/apps/web/lib/sentry/config.ts
+++ b/apps/web/lib/sentry/config.ts
@@ -190,14 +190,28 @@ const SENSITIVE_QUERY_PARAMS =
 const DEPLOYMENT_TRANSITION_PATTERN = /^.+ is not defined$/;
 
 /**
- * Pattern matching stale server action IDs after deployment.
+ * Message patterns matching stale server action errors after deployment.
  *
- * When a deployment changes server action manifests, clients with stale
- * JavaScript can reference an old action hash and trigger
- * `UnrecognizedActionError` until refresh.
+ * Next.js uses two error shapes for this condition:
+ *
+ * 1. Client-side `UnrecognizedActionError` (type field) — thrown by the
+ *    App Router when the server returns NEXT_ACTION_NOT_FOUND_HEADER.
+ *    Message: `Server Action "<hash>" was not found on the server.`
+ *
+ * 2. Server-side `Error` — thrown by action-handler.js when the action
+ *    manifest lookup fails (E974/E975 error codes).
+ *    Message: `Failed to find Server Action. This request might be from
+ *              an older or newer deployment.`
+ *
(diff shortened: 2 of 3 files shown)


itstimwhite:
## Review Summary
**Verdict**: Approve

This is a well-structured, targeted fix for a real production issue. The root cause analysis is accurate, the two-bug diagnosis is correct, and the solution addresses both the Sentry noise and the broken UX loop. The code is clean, well-documented, and follows Jovie conventions.

---

### Critical Issues
None found.

---

### Warnings

- **`apps/web/lib/sentry/config.ts:311`** — The `type === 'unrecognizedactionerror'` check is case-sensitive. While Next.js consistently uses `UnrecognizedActionError` (PascalCase), the `toLowerCase()` call in `isDeploymentSkewError` suggests the team is aware of potential casing variance. Consider normalizing the type to lowercase before comparison here too, for consistency and defense-in-depth:
  ```ts
  type?.toLowerCase() === 'unrecognizedactionerror'
  ```
  This is a minor nit — the current code works with the known Next.js shape — but aligning with the helper's approach would be more robust.

---

### Suggestions

- **`apps/web/components/organisms/ErrorBoundary.tsx:26`** — The `isDeploymentSkewError` helper uses `error.constructor?.name` which is reliable for class instances but could be `undefined` for errors created with `{ message: '...' }` plain objects (unlikely in practice, but possible if caught from third-party code). Consider adding a fallback to `error.name` (the standard `Error.prototype.name` property) as a secondary check:
  ```ts
  const type = (error.constructor?.name ?? error.name ?? '').toLowerCase();
  ```

- **`apps/web/tests/unit/lib/sentry/sentry-config.test.ts`** — The three new tests are excellent. Consider adding one more edge case: a plain `Error` with the message `"was not found on the server"` but *without* the `UnrecognizedActionError` type (simulating a future Next.js version that might change the error shape). This would verify the message-pattern fallback works independently of the type check.

- **`apps/web/components/organisms/ErrorBoundary.tsx:49`** — The `captureErrorInSentry` skip for skew errors is correct (since `beforeSend` already filters them), but the comment says "skip capture" while the code just skips the call. This is fine — no double-send. However, if Sentry's `beforeSend` filter ever changes or is bypassed (e.g., via `captureException` with a different scope), these errors would silently disappear from Sentry *and* not be logged. Consider adding a `console.warn` or `console.debug` for skew errors so they're still visible in dev tools:
  ```ts
  if (isSkewError) {
    console.debug('[ErrorBoundary] Deployment skew detected — skipping Sentry capture');
  } else {
    captureErrorInSentry(...);
  }
  ```

---

### What Looks Good

1. **Accurate root cause analysis** — The PR correctly identifies that the old regex never matched because it checked `message` for a prefix that only exists in `type`. This is a classic Sentry config bug and the fix is precise.

2. **Clean separation of concerns** — The `isDeploymentSkewError` helper is defined in the ErrorBoundary component (where UX logic lives) and the Sentry filter uses its own pattern matching in `config.ts`. No cross-cutting concerns tangled.

3. **Comprehensive test coverage** — Three regression tests cover the exact shapes seen in production: client-side `UnrecognizedActionError`, server-side plain `Error` with E974/E975 message, and a hash-in-message variant. This is exactly the kind of test that prevents regression.

4. **Correct UX recovery** — Replacing "Try again" with "Reload" is the right fix. The old behavior was a user-hostile infinite loop; the new behavior matches how most frameworks handle deployment skew (hard refresh to pick up new JS bundles).

5. **Minimal diff** — Only 3 files changed, no refactoring beyond the fix, no unnecessary formatting changes. Easy to review and low risk.

6. **Documentation** — The JSDoc comments on both `isDeploymentSkewError` and the Sentry config patterns are thorough and explain the two error shapes clearly. This will help future maintainers understand why these patterns exist.

7. **Convention compliance** — No emoji in UI (uses "App Updated" text), no native browser dialogs, no decorative motion, TypeScript throughout, no `any` types. Clean.

itstimwhite (commented):

itstimwhite:
## Review Summary

**Verdict:** Approve

This PR correctly fixes two related bugs in handling deployment‑skew server action errors. The root cause analysis is accurate, the Sentry filter now matches on both the error type and message substrings, and the ErrorBoundary offers the correct recovery (hard reload) instead of an infinite “Try again” loop. The code is clean, TypeScript‑strict, and follows Jovie conventions.

---

### What Looks Good

- **Accurate root cause identification** – Both bugs (Sentry regex never matching, wrong recovery UX) are clearly explained and addressed.
- **Robust error detection** – The new `isDeploymentSkewError` helper checks both the constructor name and message substrings, covering both client‑side (`UnrecognizedActionError`) and server‑side (plain `Error`) shapes.
- **Clean separation of concerns** – Sentry filtering lives in `config.ts`, while the UX change is isolated in `ErrorBoundary.tsx`. The two layers don’t repeat logic.
- **Comprehensive tests** – Three regression tests cover the exact error shapes that were previously missed.
- **Convention compliance** – No native browser dialogs, no emoji, no `any` (except test casts which are acceptable), proper TypeScript, and a well‑structured commit message.

---

### Warnings

No critical issues. The following are minor observations that do not block approval.

- **`apps/web/components/organisms/ErrorBoundary.tsx` (line 21)** – The check `error.constructor?.name?.toLowerCase()` relies on the constructor name surviving minification. In production Next.js builds, `UnrecognizedActionError` might be mangled (e.g., to `t`). However, the message‑based fallback (lines 24–25) will catch the error regardless, so this is not a functional bug. Consider adding a comment to note the fallback, or simply remove the constructor check since the message patterns are sufficient. (Suggestion)

- **`apps/web/tests/unit/lib/sentry/sentry-config.test.ts`** – The test file uses `scrubPii(event as any)`. While `as any` is acceptable for test flexibility, the test would be stronger if it used a well‑typed factory for Sentry event objects. This is a minor improvement, not a blocker.

---

### Suggestions

1. **Add ErrorBoundary unit tests** – The PR includes Sentry config tests but does not test the new `isDeploymentSkewError` helper or the conditional rendering in `ErrorBoundary`. Adding a simple test (e.g., rendering the component with a mock skew error and asserting the “Reload” button is shown) would improve coverage and prevent regressions.

2. **Consider exposing the dependency on `globalThis`** – The use of `globalThis.location.reload()` is fine, but many codebases prefer `window.location.reload()` for clarity. Both work in modern environments; this is only a style suggestion if the rest of the codebase uses `window`.

3. **Simplify the Sentry filter logic** – The `STALE_SERVER_ACTION_MESSAGE_PATTERNS` array and the `type === 'unrecognizedactionerror'` check could be combined into a single helper function, but the current inline logic is clear enough. No change required.

---

### Final Verdict

**Approve.** The PR correctly solves the reported issues, follows conventions, and includes adequate tests for the Sentry filtering portion. No security, performance, or design drift concerns. The minor suggestions above are optional improvements.

itstimwhite (commented):

PR #8738, opened 2026-05-16
v26.5.13 fix(security): drop scanner probes at the edge (JOV-2189)

itstimwhite:
## Summary

The public profile catch-all route was matching nested third-party paths that no Jovie route ever serves, letting them reach the page handler and surface as observability warnings. Add a path-pattern detector to the proxy middleware so those requests get a quiet 404 before any auth, DB lookup, or rendering work runs.

- `apps/web/lib/security/probe-detection.ts` — pattern detector + canonical drop response. Edge-compatible, O(path length), no allowlist required.
- `apps/web/proxy.ts` — earliest exit in `middleware()`, before investor portal, before Clerk proxy, before `handleRequest`.

Closes JOV-2189.

## Risk

Low. The detector is conservative — patterns are scoped to substrings that the Jovie app could never legitimately produce. False positives are covered by an explicit "lets legitimate paths pass through" suite enumerating every public Jovie route.

## Cost Impact

Negative. The fix removes serverless invocations and Sentry events for traffic that should never have been processed.

## Verification

- `pnpm --filter web exec vitest run lib/security/probe-detection.test.ts` — 83 passed
- `pnpm --filter web exec vitest run tests/unit/middleware/` — 73 passed (existing + new probe-drop integration tests asserting probes never invoke Clerk/auth/DB mocks)
- `pnpm --filter @jovie/web run typecheck` — clean
- `pnpm biome check` — clean
- `pnpm --filter web lint:server-boundaries` — clean
- Local dev QA via curl + browse (21/21 PASS): 11 probe variants all return 404 with `Cache-Control: no-store` + `X-Robots-Tag: none`; 10 legitimate paths (`/`, `/signin`, `/changelog`, `/api/health`, `/api/profiles/lookup`, `/timwhite`, `/timwhite/listen`, `/timwhite/shop`, `/timwhite/about`, `/timwhite/pay`, `/timwhite/notifications`) all unchanged
- pre-push hook: Biome, proxy guard, typecheck, lint

Linear: JOV-2189

## Known gap (out of scope)

Local QA surfaced a pre-existing middleware matcher bug: paths like `/wp-json/wp/v2/users` bypass middleware entirely because the matcher's static-asset exclusion uses unescaped `\.` in a single-quoted JS string, so `.json` actually matches `<any>json` in the compiled regex. Out of scope for this PR. Tracked in JOV-2236.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security by implementing edge-layer detection and silent dropping of unauthenticated scanner and probe requests, preventing malicious traffic from reaching application handlers.

* **Tests**
  * Added extensive test coverage for security probe detection, including malicious path patterns, case-insensitive matching, defensive input handling, and legitimate route validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JovieInc/Jovie/pull/8738?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


itstimwhite:
## Review Summary
**Verdict**: Approve  

This PR introduces a well-architected early-exit layer in the proxy middleware that drops obviously malicious probe traffic before any downstream processing. The implementation is conservative, edge-compatible, and includes thorough test coverage. I found no blocking issues.

---

### Critical Issues
- **None detected.** The logic is sound, the patterns are appropriate for the Jovie routing domain, and the early-return placement in `proxy.ts` correctly intercepts probes before Clerk, DB, or page handling.

---

### Warnings
- **None.** All code adheres to Jovie conventions: TypeScript strictness, no emoji, no native dialogs, no decorative motion. The probe detector is purely pattern-based and does not act as a WAF (as documented).

---

### Suggestions
1. **`apps/web/lib/security/probe-detection.ts:45-46`** — Consider adding a brief inline comment explaining why some entries include a leading slash (e.g., `'/.env'` vs `'.env'`). The current comment says “Matched as a case-insensitive substring”, but a reader might wonder why a leading slash is part of some patterns. Adding a note like “Leading slash prevents false matches inside legitimate path segments (e.g., `/.env` won’t match `someenv`)” would improve maintainability.

2. **`apps/web/tests/unit/middleware/proxy-behavioral.test.ts:213-226`** — The test that checks “still serves a legitimate profile path” asserts `res.status !== 404`. While this is sufficient for the probe gate, a more explicit assertion like `expect(res.status).toBeOneOf([200, 301, 401, ...])` could make the test more robust against future changes. Not blocking, but a minor reliability improvement.

3. **`apps/web/lib/security/probe-detection.ts:134`** — The `createProbeDropResponse` sets `Content-Length: '0'` explicitly. In most edge runtimes, `new Response(null, …)` already omits a body and sets `Content-Length: 0`. This header is redundant but harmless. Consider removing it for idiomatic brevity, or keep it as an explicit signal. Both fine.

---

### What Looks Good
- **Clear separation of concerns** — A dedicated module for detection and drop response, with a single entry point in middleware.  
- **Conservative patterns** — Every blocked substring is clearly not a Jovie route, reducing false-positive risk.  
- **Thorough test coverage** — 83 unit tests passing, with integration tests in proxy-behavioral.test.ts that verify mocks are not invoked for probes.  
- **Security posture** — 404 (not 403), empty body, `no-store`, and `X-Robots-Tag: none` give attackers minimal signal.  
- **Performance-focused** — Early exit avoids auth, DB, and rendering costs for malicious traffic.  
- **Documented gap** — The known matcher bug (JOV-2236) is acknowledged and tracked separately, keeping scope tight.

itstimwhite (commented):

itstimwhite:
## Review Summary

The PR is well-structured and addresses a real security/operations concern. However, there is a pattern-match risk that could block legitimate user profile pages, violating the stated design rule of being conservative. See Critical Issues below.

---

### Critical Issues

**`apps/web/lib/security/probe-detection.ts:34-56` — Unanchored substring patterns risk false positives on legitimate usernames**

Several `PROBE_PATH_SUBSTRINGS` are general substrings that could match a creator’s username or a legitimately nested `[username]/[...slug]` path:

- `'phpmyadmin'` → matches `/phpmyadmin` (a valid top-level profile route) and `/phpmyadmin-fan/listen`  
- `'phpinfo'` → matches `/phpinfo-fan/`  
- `'wp-admin'`, `'wp-content'`, etc. → matches `/wp-admin-fan/`  
- `'wordpress/'` (with trailing slash) → matches `/wordpress/listen` if a user named `wordpress` exists  

The same applies to `'xmlrpc.php'` — while extremely unlikely, a slug segment could theoretically be named `xmlrpc.php`.

The PR description claims “patterns are scoped to substrings that the Jovie app could never legitimately produce,” yet a creator named “phpmyadmin” or “wordpress” is plausible in a public platform. The test suite’s “lets legitimate paths pass through” block does not include any username containing these substrings, so false positives are untested.

**Suggestion**:  
Either segment-bound these patterns (e.g., `/\bphpmyadmin\b/` anchored to path segments), or explicitly document in code and tests that these patterns are known to risk blocking edge-case usernames, and include a follow-up tracking issue. Since you already require trailing slashes for `'/myadmin/'` and `'/pma/'`, the same approach should be applied to other ambiguous patterns. For example, `'phpmyadmin'` could be `'/phpmyadmin/'` or `'/phpmyadmin'` with an optional trailing slash / EOS check.

---

### Warnings

**None critical beyond the false-positive concern.**

---

### Suggestions

1. **`apps/web/lib/security/probe-detection.ts:141` — Consider using `new Response('', ...)` instead of `new Response(null, ...)`**  
   While `null` is acceptable, an empty string body is more explicit and avoids potential edge-case quirks in some runtime environments.

2. **`apps/web/lib/security/probe-detection.ts:12-13` — Add a test case for a legitimate path that contains a probe substring that is *not* a word boundary (e.g., `/timmyadmin/listen`)**  
   This would prove the pattern does not accidentally match when the substring appears as part of a larger word. Currently, only exact substrings are tested for blocking, not for safe non-matching.

3. **`apps/web/lib/security/probe-detection.ts:9` — `'phpmyadmin'` could also be anchored as `'/phpmyadmin'` (with leading slash) to avoid matching `/someuser/phpmyadmin/page` if that were ever legitimate.**  
   If you decide to segment-bound, apply consistently across all ambiguous tokens.

4. **`apps/web/lib/security/probe-detection.ts` — Consider making the 2KB limit a named constant** for clarity and ease of future adjustment, e.g., `MAX_PATHNAME_LENGTH = 2048`.

---

### What Looks Good

- **Early exit location**: The probe detection is correctly placed at the very top of `middleware()`, before any auth, DB lookup, or request analysis, minimizing serverless invocation cost and observability noise.
- **Conservative response design**: Returning `404` (not `403`) with `no-store` and `X-Robots-Tag: none` denies scanners any signal that the path was specifically blocked.
- **Edge compatibility**: The module uses only plain JS string methods — no Node APIs, no fetch — ensuring it works in Edge runtime.
- **Thorough test coverage**: 83 unit tests + integration tests that verify Clerk/auth mocks are *not* called after a probe match, proving the early exit works.
- **Defensive input handling**: Null/undefined/non-string inputs return `false` safely, preventing middleware crashes from unexpected upstream callers.
- **Clear documentation**: JSDoc explains the rationale, design rules, and what the function must NOT do (act as a general WAF).
- **CHANGELOG and versioning follow conventional format**.
- **Follows Jovie conventions**: no emoji, no native dialogs, no decorative hover motion, TypeScript strict mode, etc.

itstimwhite (commented):

PR #8907, opened 2026-05-17
fix: add production signup canary

itstimwhite:
## Summary
- Added a redacted signup/onboarding readiness contract plus CLI guard for required runtime keys.
- Added a production signup/onboarding canary path with Gmail OTP support, narrow synthetic cleanup, and blocking synthetic workflow wiring.
- Hardened auth route/copy/waitlist/front-door checks, Turnstile missing-key tests, and public screenshot/demo asset checks.
- Split auth page client components behind Suspense so source production builds no longer fail on client search params.

## Production repair
- Copied the staging Turnstile key pair into Doppler production as requested, without printing values.
- Fixed the production Vercel public env classification for the Turnstile site key and promoted the repaired production deployment.
- Updated the Cloudflare Turnstile staging widget hostnames so production domains are allowed; dashboard confirms the widget now has 3 hostnames.

## Verification
- `pnpm --filter @jovie/web exec vitest run tests/unit/app/signin-page.test.tsx tests/unit/app/signup-page.test.tsx tests/unit/app/auth-front-door-contract.test.ts tests/unit/lib/readiness tests/unit/onboarding/OnboardingTurnstile.test.tsx --config vitest.config.mts`
- `pnpm --filter @jovie/web exec vitest run tests/unit/e2e/production-signup-canary.test.ts tests/unit/product-screenshots/marketing-export.test.ts --config vitest.config.mts`
- `doppler run --project jovie-web --config prd -- pnpm --filter @jovie/web exec tsx scripts/check-signup-readiness.ts --target=prd --source=env`
- `doppler run --project jovie-web --config dev -- pnpm --filter @jovie/web run build`
- Live production browser smoke: `/`, `/start`, `/signup`, `/signin` all returned 200 with no Turnstile config error, no 110200, and no auth-unavailable text.
- Pre-push hook passed: repo Biome check, web proxy guard, Tailwind guard, typecheck, and Biome lint.

## Operational note
Production and staging now share the same Turnstile widget because that was the requested incident hotfix. Changes to that widget affect both environments.

## Summary by CodeRabbit

* **New Features**
  * Cloudflare Turnstile onboarding protection documented and front-door readiness checks added.
  * Production signup readiness verification and a synthetic production signup canary for end-to-end validation.

* **Bug Fixes**
  * Stricter detection/failure for front-door / Turnstile configuration errors during health checks.

* **Refactor**
  * Server/client split for auth pages and streamlined modal flows.

* **Documentation**
  * Updated setup and synthetic monitoring runbooks.

* **Tests**
  * New E2E canary and expanded unit coverage for auth and readiness.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JovieInc/Jovie/pull/8907?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


itstimwhite:
## Review Summary

### Critical Issues

- **`apps/web/app/(auth)/signup/SignUpPageClient.tsx:55`** — The `useEffect` dependency array includes `searchParams` (an object). Because `useSearchParams()` returns a new object reference on every render, this effect will re-run on every render, causing unnecessary API calls (`/api/handle/check`) and state updates.  
  **Fix**: Depend on `searchParams.toString()` or individual param values (`spotifyUrl`, `artistName`, `plan`, `handle`) instead of the object itself.

- **`apps/web/app/(auth)/signup/SignUpPageClient.tsx:1`** — `AuthLayout` is imported from `@/features/auth`. If `AuthLayout` is a server component (e.g., uses `cookies()`, `headers()`, or database queries), it cannot be rendered inside a client component. The PR moves the `AuthLayout` wrapper into the client component, which would break if `AuthLayout` is server-only.  
  **Fix**: Verify that `AuthLayout` is a client component or does not use any server-only features. If it is a server component, keep it in the server page and only move the client-specific parts (e.g., `SignUpPageClient` should not wrap `AuthLayout`).

### Warnings

- **`apps/web/app/(auth)/signup/SignUpPageClient.tsx:55`** — The effect also lists `handle` as a dependency, but `handle` is derived from `searchParams`. Since `searchParams` already triggers re-runs when `handle` changes, the extra dependency is redundant. Consider removing it for clarity.

- **`apps/web/app/(auth)/signin/SignInPageClient.tsx:18-48`** — The `isValidEmail` function is duplicated from the original `page.tsx`. This creates a maintenance risk if the validation logic needs to change.  
  **Suggestion**: Extract `isValidEmail` into a shared utility (e.g., `lib/auth/email-validation.ts`) and import it in both places.

- **`apps/web/app/(auth)/signup/SignUpPageClient.tsx:55`** — The first `useEffect` uses `searchParams` to read `spotify_url`, `artist_name`, and `plan`, but also uses `handle` in the `source` assignment. However, `handle` is not read from `searchParams` inside that effect (it is read outside). This inconsistency could lead to stale `handle` values if the effect doesn't re-run when `handle` changes.  
  **Fix**: Either read `handle` from `searchParams` inside the effect, or add `handle` to the dependency array (which is already done, but the effect body should use the local `handle` variable from the closure, which is fine). The current code is correct but confusing; consider refactoring for clarity.

### Suggestions

- **Add unit tests for `SignUpPageClient`** — The new client component contains non-trivial logic (claim data persistence, OAuth error banner, handle availability check). While the existing tests cover the page-level behavior, adding a focused test for `SignUpPageClient` (e.g., mocking `useSearchParams` and verifying the rendered output) would improve coverage.

- **Consider a skeleton fallback for modal Suspense** — The modal pages (`@auth/(.)signin/page.tsx`, `@auth/(.)signup/page.tsx`) use `Suspense fallback={null}`. This may cause a flash of empty content while the client component loads. A minimal skeleton (e.g., a spinner or placeholder) would improve perceived performance.

- **Verify `connection()` usage** — The `await connection()` call in modal pages is correct for preventing dynamic rendering, but ensure that the modal still renders correctly during SSR (e.g., when the connection is not yet established). The Suspense boundary should handle this, but double-check that the modal does not rely on client-only APIs before the connection is ready.

### What Looks Good

- **Extraction of client components behind Suspense** — Moving `useSearchParams`-dependent code into separate client components and wrapping them with `Suspense` fixes the production build failure caused by client-side search params in server components. This is a clean architectural improvement.

- **Turnstile integration** — Adding Cloudflare Turnstile bot protection on sign-up and onboarding pages improves security without degrading UX. The env var naming follows conventions (`NEXT_PUBLIC_TURNSTILE_SITE_KEY`, `TURNSTILE_SECRET_KEY`).

- **Comprehensive test and CI updates** — The PR adds unit tests for readiness contracts, Turnstile, and production signup canary, plus updates to CI workflows (canary health gate, synthetic monitoring, e2e matrix). This demonstrates a strong commitment to reliability.

- **Operational transparency** — The PR description clearly documents the shared Turnstile widget between staging and production, and the scoped synthetic cleanup approach. This reduces confusion for future incidents.

- **Homepage CTA contract** — Using `getHomepageFrontDoorCtaContract` to conditionally render the secondary CTA based on feature flags is a clean pattern that keeps the homepage static while allowing dynamic behavior.

itstimwhite (commented):

itstimwhite:
## Review Summary

The PR introduces valuable production resilience mechanisms (synthetic canary, readiness checks, Turnstile hardening). However, several issues need attention before merging.

---

### Critical Issues

- **apps/web/app/(home)/page.tsx:227**  
  `getHomepageFrontDoorCtaContract(FEATURE_FLAGS.WAITLIST_ENABLED).secondary`  
  This call uses `FEATURE_FLAGS.WAITLIST_ENABLED` which is typically derived from `process.env`. In a server component, reading runtime env at the top level is safe, but the function `getHomepageFrontDoorCtaContract` is imported from `@/data/homepageFrontDoorCta`. If that module reads `FEATURE_FLAGS` at module scope, it may not reflect runtime overrides in a static export. **Fix**: Confirm the function is a pure sync function that accepts the flag as argument (as done) **and** that no module-level env reads exist. Also ensure `secondaryCta` matches the shape used later (`href`, `label`). If the function returns `undefined` when waitlist is enabled (as the old logic implied), the new code is equivalent—but verify the contract returns exactly `{ href: string, label: string }` or `null`.

- **apps/web/app/(auth)/signin/page.tsx & signup/page.tsx**  
  `AuthLayout` is now inside the client component (`SignInPageClient`, `SignUpPageClient`), whereas previously it lived outside `Suspense` in the server page. This forces `AuthLayout` (which is likely pure React markup) to be bundled as client code even if it uses no hooks. **Fix**: Either keep `AuthLayout` in the server page (it can wrap the `Suspense` boundary) or move only the `useSearchParams`-dependent fragments into the client component. Consider:
  ```tsx
  // page.tsx
  export default function SignInPage() {
    return (
      <AuthLayout formTitle="Sign in" ...>
        <Suspense fallback={<AuthFormSkeleton />}>
          <SignInPageClient />
        </Suspense>
      </AuthLayout>
    );
  }
  ```

- **apps/web/.github/workflows/synthetic-monitoring.yml** – Cleanup removal  
  The step `Cleanup Synthetic E2E Users` was replaced with a mere echo. The PR description claims cleanup is “scoped inside synthetic-golden-path.spec.ts”, but the test file is not shown in the diff. If the test itself does not clean up on failure, these users will accumulate in production. **Fix**: Either confirm the spec has robust `afterAll`/error handling, or restore a safe cleanup step that deletes only canary accounts (e.g., filtering by email base `E2E_PROD_SIGNUP_EMAIL_BASE`).

- **apps/web/app/@auth/(.)signin/page.tsx** – `await connection()` usage  
  `connection()` from `next/server` is an experimental API (Next.js 15+) and may cause runtime errors in older versions or in static generation contexts. The PR does not update `next.config.js` or `package.json` to pin a compatible version. **Fix**: Add a version constraint (e.g., `"next": ">=15.0.0-rc"`) or use a polyfill/alternative if the app is not yet on Next.js 15. Also ensure `next.config.js` exports are compatible.

---

### Warnings

- **apps/web/app/(auth)/signup/SignUpPageClient.tsx** – Missing error boundary for OAuth error banner  
  The `SignUpOauthErrorBanner` component uses `globalThis.history.replaceState`. If SSR/prerendering runs this code (it shouldn't, it's a client component, but there is no explicit `'use client'` guard inside the function), it could throw. **Fix**: Wrap the component in a check for `typeof window !== 'undefined'` or ensure it's only rendered client-side.

- **apps/web/scripts/check-signup-readiness.ts** – No diff provided  
  This critical script is not visible. It must validate every required key (`TURNSTILE_SECRET_KEY`, `NEXT_PUBLIC_TURNSTILE_SITE_KEY`, `CLERK_SECRET_KEY`, etc.) and fail fast with a clear message. Ensure it doesn’t rely on `.env` loading that may be incomplete in CI.

- **.env.example** – New prod canary variables  
  `E2E_PROD_SIGNUP_PASSWORD` is listed as a placeholder. If a developer accidentally fills these with real values and pushes, the secrets are exposed. **Fix**: Add a clear comment above these vars: `# DO NOT COMMIT REAL VALUES - CONFIGURE ONLY IN DOPPLER`.

- **apps/web/app/(home)/page.tsx** – Possible duplicate or missing import  
  `getHomepageFrontDoorCtaContract` is imported from `@/data/homepageFrontDoorCta`; ensure this file actually exports a sync function (not async) and does not have side effects.

---

### Suggestions

- **Extract `isValidEmail` to a shared utility**  
  It’s duplicated in `SignInPageClient.tsx` and may have been copied from the old file. Consider moving to `@/lib/auth/email-validation.ts` and reusing in both the client and any server-side validation.

- **Consider type‑safe env for canary credentials**  
  In `production-signup-canary.ts`, use the validated server environment schema (`env-server-schema.ts`) to ensure all required vars exist at runtime rather than ad‑hoc `process.env` reads.

- **Add a comment linking canary cleanup logic**  
  In `synthetic-golden-path.spec.ts`, include a comment at the top: “User cleanup is handled here; see DESIGN.md for canary account lifecycle.” This prevents future confusion.

- **Prefer `next/navigation` stable API**  
  If `connection()` is not yet stable, a workaround is to export the page as dynamic: `export const dynamic = 'force-dynamic'` and avoid `await connection()`.

---

### What Looks Good

- **Thorough verification steps** – The PR includes typecheck, unit tests, Doppler readiness script, and live smoke tests across multiple routes. This is the right level of confidence for a production-facing change.

- **Split auth pages behind Suspense** – Separating `useSearchParams` into dedicated client components fixes the original build failure while keeping the skeleton fallback SSR-friendly. The structure is clean and modular.

- **CI hardening** – Adding `/start` to the canary health gate and checking for “turnstile is not configured” text improves early detection of misconfiguration.

- **Doppler-first secrets** – Canary credentials are explicitly scoped to Doppler `prd` and never hardcoded. The `.env.example` placeholders reduce accident risk.

- **Readiness contract** – The `scripts/check-signup-readiness.ts` (assuming it’s well-written) and the unit tests provide a safety net before every production deployment.

---

**Action Items** (address critical issues before merging):
1. Verify `getHomepageFrontDoorCtaContract` contract and move `AuthLayout` back to server page.
2. Confirm canary cleanup in test file or reinstate safe cleanup step.
3. Pin Next.js version or replace `connection()` to avoid runtime breakage.
4. Review `check-signup-readiness.ts` source (not visible) for completeness.

itstimwhite (commented):

PR #8919, opened 2026-05-17
fix: resolve signup readiness env file from repo root

itstimwhite:
## Summary
- Fixes the production deploy gate crash from `pnpm --filter=@jovie/web exec` running the readiness script from `apps/web` while Vercel writes `.vercel/.env.production.local` at the repo root.
- Adds focused script coverage for repo-root env file resolution from the web package cwd.

## Verification
- `pnpm biome check --write apps/web/scripts/check-signup-readiness.ts apps/web/scripts/check-signup-readiness.test.ts`
- `pnpm --filter @jovie/web exec vitest run scripts/check-signup-readiness.test.ts tests/unit/lib/readiness/signup-onboarding.test.ts`
- `pnpm --filter=@jovie/web exec tsx scripts/check-signup-readiness.ts --target=prd --source=vercel-file --file=.vercel/.env.production.local`
- pre-push hook: `biome check . && pnpm --filter=@jovie/web run pre-push`

### DIFF
--- a/apps/web/scripts/check-signup-readiness.test.ts
+++ b/apps/web/scripts/check-signup-readiness.test.ts
@@ -1,0 +1,33 @@
+import { mkdirSync, rmSync, writeFileSync } from 'node:fs';
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { afterEach, describe, expect, it } from 'vitest';
+import { resolveEnvFilePath } from './check-signup-readiness';
+
+const scriptsDir = dirname(fileURLToPath(import.meta.url));
+const webRoot = resolve(scriptsDir, '..');
+const repoRoot = resolve(webRoot, '..', '..');
+const testDirName = `.tmp-signup-readiness-env-file-test-${process.pid}`;
+const testDir = resolve(repoRoot, testDirName);
+
+afterEach(() => {
+  rmSync(testDir, { force: true, recursive: true });
+});
+
+describe('signup readiness env file resolution', () => {
+  it('finds repo-root Vercel env files when invoked from the web package cwd', () => {
+    const relativeEnvPath = `${testDirName}/prod.env`;
+    const repoEnvPath = resolve(repoRoot, relativeEnvPath);
+    mkdirSync(dirname(repoEnvPath), { recursive: true });
+    writeFileSync(repoEnvPath, 'NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=pk_test\n');
+
+    const originalCwd = process.cwd();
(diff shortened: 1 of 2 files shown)


itstimwhite:
## Review Summary

### Critical Issues

- **apps/web/scripts/check-signup-readiness.ts:134-137** – The `main()` guard uses `pathToFileURL(process.argv[1])` which **throws** when `process.argv[1]` is a relative path not starting with `./` or `../`.  
  When the script is invoked as `tsx scripts/check-signup-readiness.ts`, Node sets `process.argv[1]` to `scripts/check-signup-readiness.ts` (a bare relative path). `pathToFileURL` requires either an absolute path or a path starting with `./` or `../`; it will throw `ERR_INVALID_ARG_TYPE` or `ERR_INVALID_ARG_VALUE`. This causes the script to crash with an unhandled error instead of running `main()`.  
  **Fix**: Replace the guard with a safe check that resolves both paths to absolute before comparing:  
  ```typescript
  if (process.argv[1] && fileURLToPath(import.meta.url) === resolve(process.argv[1])) {
    main().catch(error => { ... });
  }
  ```  
  (Import `fileURLToPath` from `'node:url'` – it’s already imported in the file.)

### Warnings

- **apps/web/scripts/check-signup-readiness.test.ts:1-12** – The test creates a temporary directory in the repo root using `process.pid`. In CI or parallel runs this could clash, though unlikely. Consider using `fs.mkdtempSync` with a prefix (e.g., `resolve(repoRoot, '.tmp-signup-readiness-test-')`) to guarantee uniqueness.

### Suggestions

- **apps/web/scripts/check-signup-readiness.ts:37-43** – The candidate order (cwd → webRoot → repoRoot) could mask a misconfiguration if a file with the same relative path exists at cwd but the intended env file is at repo root. Since the root cause is Vercel writing at repo root, consider checking repoRoot first, then webRoot, then cwd. Not a blocker, but would align the resolution strategy with the deployment environment expectation.

- **apps/web/scripts/check-signup-readiness.test.ts:5** – The test imports the internal function `resolveEnvFilePath` but does not test any edge cases (absolute path, file not found, file at webRoot only). Consider adding a test for the “not found” error path (expect throw) to guarantee the error message includes all candidates.

### What Looks Good

- The `resolveEnvFilePath` extraction is clean and makes the logic testable.
- The test file restores `process.cwd()` in a `finally` block – good practice.
- The exported `resolveEnvFilePath` allows unit testing without running the full CLI.
- The added guard prevents side effects when the script is imported as a module.
- The overall approach (checking multiple roots) directly addresses the deployment crash described in the PR summary.

itstimwhite (commented):

PR #9259, opened 2026-05-19
refactor(contacts): use shared loading table skeleton

itstimwhite:
## Summary
- replaces the bespoke desktop contacts loading table with UnifiedTableSkeleton
- keeps the existing mobile contacts loading layout and footer intact
- keeps the route loading state aligned with shared shell table primitives

## Verification
- pnpm biome check --write apps/web/app/app/\(shell\)/dashboard/contacts/loading.tsx
- pnpm --filter @jovie/web exec vitest run components/organisms/table/organisms/UnifiedTableSkeleton.test.tsx
- pnpm --filter @jovie/web run typecheck -- --pretty false
- pre-push: biome check ., next proxy guard, tailwind guard, web typecheck, web lint

Linear: JOV-2482

## Summary by CodeRabbit

* **Refactor**
  * Updated the contacts page loading skeleton UI with an improved, unified architecture that provides consistent loading states across desktop and mobile views.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JovieInc/Jovie/pull/9259?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/web/app/app/(shell)/dashboard/contacts/loading.tsx
+++ b/apps/web/app/app/(shell)/dashboard/contacts/loading.tsx
@@ -1,34 +1,81 @@
+'use client';
+
+import { type ColumnDef, createColumnHelper } from '@tanstack/react-table';
 import { LoadingSkeleton } from '@/components/molecules/LoadingSkeleton';
+import { UnifiedTableSkeleton } from '@/components/organisms/table';
 import { SKELETON_ROW_COUNT } from '@/lib/constants/layout';
 
-const CONTACTS_HEADER_KEYS = [
-  'col-role',
-  'col-name',
-  'col-territories',
-  'col-email',
-  'col-phone',
-  'col-actions',
-] as const;
+type ContactsLoadingRow = {
+  readonly role: string;
+  readonly name: string;
+  readonly territories: string;
+  readonly email: string;
+  readonly phone: string;
+  readonly actions: string;
+};
 
-const CONTACTS_ROW_KEYS = Array.from(
-  { length: SKELETON_ROW_COUNT.TABLE },
-  (_, i) => `contacts-row-${i + 1}`
-);
+const contactLoadingColumnHelper = createColumnHelper<ContactsLoadingRow>();
+
+function createLoadingHeader(width: string) {
+  function ContactsLoadingHeader() {
+    return <LoadingSkeleton height='h-4' width={width} rounded='md' />;
+  }
+
+  return ContactsLoadingHeader;
+}
+
+const CONTACTS_LOADING_COLUMNS = [
+  contactLoadingColumnHelper.accessor('role', {
+    id: 'role',
+    header: createLoadingHeader('w-[112px]'),
+    size: 180,
+    minSize: 180,
+  }),
+  contactLoadingColumnHelper.accessor('name', {
+    id: 'name',
+    header: createLoadingHeader('w-[128px]'),
+    size: 200,
+    minSize: 200,
+  }),
+  contactLoadingColumnHelper.accessor('territories', {
+    id: 'territories',
+    header: createLoadingHeader('w-[96px]'),
+    size: 140,
+    minSize: 140,
+  }),
+  contactLoadingColumnHelper.accessor('email', {
+    id: 'email',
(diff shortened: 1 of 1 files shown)


itstimwhite:
## Review Summary

The PR replaces a custom desktop contacts loading table with `UnifiedTableSkeleton` and preserves the mobile layout. The approach is correct, but there are two issues that should be addressed before merging.

---

### Critical Issues

1. **Unnecessary `'use client'` directive makes loading.tsx a client component**  
   `apps/web/app/app/(shell)/dashboard/contacts/loading.tsx:1`  
   The file now starts with `'use client'`. This forces the entire loading state to be executed on the client side, adding unnecessary JavaScript to the route’s loading chunk.  
   - The only client dependency is `UnifiedTableSkeleton` (a client component) and the `@tanstack/react-table` helpers (`createColumnHelper`, `ColumnDef`) are pure functions that can run on the server.  
   - A server component can render a client component as a child without itself being a client component.  
   - **Why it matters**: Loading states are supposed to be lightweight; shipping a client component bundle for a skeleton increases the time to interactive and adds to the main thread workload. This violates the project’s performance guideline “Performance must not replace route UIs (same design, faster)”.  
   - **Fix**: Remove `'use client'`. The `ContactsLoading` default export can remain a server component.

2. **Skeleton variant mismatch for the “actions” column**  
   `apps/web/app/app/(shell)/dashboard/contacts/loading.tsx:78`  
   The config uses `variant: 'avatar'` for the `actions` column. In the actual contacts table the actions column is a kebab menu (three-dot icon) – not an avatar. Using an avatar skeleton (circular) will create a visual mismatch between loading and final UI, which could confuse users.  
   - **Why it matters**: According to design drift rules, loading placeholders should match the final layout as closely as possible.  
   - **Fix**: Change the variant for actions to `'text'` or a small `'button'` variant (if `UnifiedTableSkeleton` supports it), or pass a custom skeleton renderer that mimics a 24px wide kebab icon.

---

### Warnings

- **No dedicated test for the new loading component**  
  The PR verifies existing tests for `UnifiedTableSkeleton` but does not add a test for `ContactsLoading` itself. While not strictly required because the component is thin, adding a snapshot or render test would increase confidence that the skeleton structure matches the final table (e.g., column count, row count, footer presence). This is a minor gap.

- **Pixel-based widths vs. Tailwind utility classes**  
  The old implementation used Tailwind width utilities (`w-28`, `w-32`, etc.) while the new one uses arbitrary values (`w-[112px]`, `w-[128px]`). This is acceptable but could cause slight visual differences. Confirm that `UnifiedTableSkeleton`’s column sizing matches the actual contacts table columns exactly.

---

### Suggestions

- **Consider extracting column definitions to a shared constant file**  
  The column definitions (`CONTACTS_LOADING_COLUMNS`) and the skeleton config (`CONTACTS_LOADING_SKELETON_CONFIG`) could be reused if the contacts data table ever needs a similar loading state. Moving them to `@/lib/contacts/columns` or similar would reduce duplication.

- **Use `as const` on skeleton config for better type inference**  
  The config object `CONTACTS_LOADING_SKELETON_CONFIG` is typed as `{ width: string; variant: 'text'|'badge'|'avatar' }[]`. Adding `as const` to the array would allow the type system to narrow the tuple length, which could help with future validation.

- **Review `rowHeight` versus actual row height in the data table**  
  The old code set `height: 44` via inline style. The new code passes `rowHeight={44}`. Ensure the data table also uses the same height (44px) to prevent a layout shift when data loads.

---

### What Looks Good

- The refactor correctly centralises the loading skeleton into the shared `UnifiedTableSkeleton` component, reducing bespoke markup.
- Mobile layout remains untouched, preserving responsive behavior.
- The footer is kept intact, matching the actual contacts page structure.
- Type safety is enforced with the `ContactsLoadingRow` type and `createColumnHelper`.
- No raw DOM manipulation or `any` types are introduced.
- The commit message follows conventional commit format.

---

**Summary**: The PR is a solid refactoring step. The two issues (unnecessary `'use client'` and wrong skeleton variant) are easy to fix and should be addressed before merging. Once resolved, the changes will improve maintainability without regressing performance or visual accuracy.

itstimwhite (commented):

PR #9283, opened 2026-05-19
JOV-2494: seal releases warm navigation path for shell persistence

itstimwhite:
## Summary
- warm the releases route across shell surfaces so warm navigation carries both route and releases query state
- skip the releases pending-shell overlay once the route is warm and clear residual pending state before paint
- tighten the shell persistence regression to fail if the releases overlay activates during warm navigation

## Verification
- `pnpm exec vitest run apps/web/tests/unit/app/shell-route-matches.test.ts apps/web/tests/unit/dashboard/ReleasesClientBoundary.test.tsx`
- `pnpm exec playwright test apps/web/tests/e2e/shell-route-persistence.spec.ts` (skipped in this environment)
- `git push` pre-push hook: repo Biome plus `@jovie/web` pre-push checks

## Summary by CodeRabbit

* **Refactor**
  * Optimized releases route prefetching and navigation shell behavior for improved responsiveness.

* **Tests**
  * Added detection mechanism to verify overlay visibility during releases route navigation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JovieInc/Jovie/pull/9283?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/web/app/app/(shell)/dashboard/releases/ReleasesClientBoundary.tsx
+++ b/apps/web/app/app/(shell)/dashboard/releases/ReleasesClientBoundary.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import type { ReactNode } from 'react';
-import { useEffect } from 'react';
+import { useLayoutEffect } from 'react';
 import { usePendingShell } from '@/components/organisms/PendingShellContext';
 
 interface ReleasesClientBoundaryProps {
@@ -13,7 +13,7 @@ export function ReleasesClientBoundary({
 }: ReleasesClientBoundaryProps) {
   const { clearPendingShell } = usePendingShell();
 
-  useEffect(() => {
+  useLayoutEffect(() => {
     clearPendingShell('releases');
   }, [clearPendingShell]);
 
--- a/apps/web/components/features/dashboard/dashboard-nav/DashboardNav.tsx
+++ b/apps/web/components/features/dashboard/dashboard-nav/DashboardNav.tsx
@@ -252,6 +252,7 @@ export function DashboardNav(_: DashboardNavProps) {
   // Debounced prefetch: avoid firing on fast mouse sweeps across nav items
   const prefetchTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const releasesPrefetchedProfileIdRef = useRef<string | null>(null);
+  const releasesWarmReadyProfileIdRef = useRef<string | null>(null);
   useEffect(
     () => () => {
       if (prefetchTimerRef.current) clearTimeout(prefetchTimerRef.current);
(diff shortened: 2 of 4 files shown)


itstimwhite:
## Review Summary
**Verdict**: Approve

The PR correctly implements warm navigation for the releases route, skips the pending shell overlay when the route is already warm, and adds a robust E2E probe to detect unwanted overlay activations. The logic is sound, the refactoring is clean, and the changes align with Jovie conventions.

---

### Critical Issues
None.

---

### Warnings
- **`apps/web/components/features/dashboard/dashboard-nav/DashboardNav.tsx` (line ~280)** — The `warmReleasesRoute` function sets `releasesPrefetchedProfileIdRef.current = profileId` twice (once at the top of the function and again after successful warm-ready). This is redundant but harmless. Consider removing the early assignment to avoid confusion.

- **`releasesPrefetchedProfileIdRef.current = profileId;` before the async block is unnecessary because the catch block already resets it on failure. If the warm succeeds, the second assignment overwrites it anyway. Not a bug, but could be simplified.

- **`apps/web/components/features/dashboard/dashboard-nav.tsx` (line ~325)** — The `handlePrefetch` callback for the releases item calls `warmReleasesRoute()` and then returns early. However, the original code also imported `@/lib/queries/prefetch-dashboard` for other items. The new code returns early for releases, which is correct because `warmReleasesRoute` already handles both the component and data prefetch. No issue.

---

### Suggestions
- **`apps/web/app/app/(shell)/dashboard/releases/ReleasesClientBoundary.tsx`** — Changing `useEffect` to `useLayoutEffect` is justified to avoid a flash of the pending shell overlay. However, consider adding a comment explaining comment why `useLayoutEffect` is necessary (e.g., "Clear pending shell synchronously before paint to prevent overlay flash"). This helps future maintainers understand the deviation from typical React patterns.

- **`apps/web/components/features/dashboard/dashboard-nav.tsx** — The `useEffect` that resets `releasesPrefetchedProfileIdRef` and `releasesWarmReadyProfileIdRef` on `profileId` change could be simplified by using a single `useRef` for the profileId and reset. Currently it's a separate `useEffect` with an empty cleanup. Consider merging it into the existing `useEffect` that handles the debounced warm trigger, or using a `useRef` that is reset inline. Not a blocker.

- **E2E test** — The `attachReleasesOverlayProbe` function uses `globalThis.getComputedStyle`. While correct in Playwright, using `window.getComputedStyle` is more idiomatic. Also, the probe attaches to `[data-testid="releases-shell-ready"]` but the performance manifest now uses `[data-app-shell-frame="true"]`. Ensure the test selector matches the actual overlay element. If the overlay uses a different testid, the probe may never detect activations. Consider using the same selector as the manifest or a more generic one.

- **Performance route manifest** — Changing the ready selector from `[data-testid="releases-shell-ready"]` to `[data-app-shell-frame="true"]` broadens the match. Verify that `[data-app-shell-frame="true"]` is present on the shell frame during warm navigation and that it does not match prematurely (e.g., before the shell is fully rendered). If the shell frame is always present, this could cause false positives in performance measurements.

---

### What Looks Good
- **Clear separation of concerns** — The warm logic is extracted into a `warmReleasesRoute` callback, making it reusable and testable.
- **Correct use of refs** — The `releasesWarmReadyProfileIdRef` prevents redundant warm attempts and allows the shell overlay to be skipped when the route is already warm.
- **Debounced prefetch debouncing** — The 300ms debounce for the initial warm trigger on pathname change is preserved, preventing unnecessary work on fast route changes.
- **E2E overlay probe** — The new probe is well-designed: it uses a `MutationObserver` to track visibility changes and provides a clean API for asserting API. This will catch regressions where the overlay appears during warm navigation.
- **Consistent cleanup** — All timers and refs are properly cleaned up on unmount or profile change.
- **No breaking changes** — The public API of `DashboardNav` remains unchanged; only internal logic is refactored.

itstimwhite (commented):

PR #9302, opened 2026-05-19
feat(shell): integrate lyrics/screening-room into unified app shell (Wave 7)

itstimwhite:
## Summary
Smallest correct integration of the lyrics (screening-room) route into the converged shell-v1 frame:
- Lyrics now renders cleanly inside the shell content surface without duplicate headers (shell DashboardHeader suppressed on /app/lyrics/*, matching the settings pattern).
- `isLyricsRoute` plumbed through `useAuthRouteConfig` → `AuthShell` for route-aware chrome decisions.
- Preserves: global audio player state (no restart on warm nav), Esc/keyboard (existing handlers in `PersistentAudioBar` + `LyricsPageClient`), `?from=` return routing from releases/library/chat, cinematic shell tokens, reduced-motion parity, mobile bottom nav, sidebar.
- Entry via audio bar lyrics button (L key), play from library/releases rows, direct URL all feel native.

## Scope (HOT ZONE only)
- `apps/web/hooks/useAuthRouteConfig.ts`
- `apps/web/components/organisms/AuthShell.tsx`
- `apps/web/components/organisms/AuthShellWrapper.tsx`
- Related test mocks

## Verification
- All lyrics + audio + shell-route unit tests pass.
- `pnpm --filter @jovie/web run typecheck`
- `pnpm biome check --write apps/web`
- Pre-commit hooks (biome, eslint, a11y, typecheck-singleflight) passed on commit.
- Rebased on main.

## Follow-up (if any)
None — this completes the minimal Wave 7 lyrics cohesion per handoff.

Refs: shell-v1-unified-app-shell-agent-handoff-20260518.md (Wave 7)
JOVIE_AGENT_PROFILE=coder (isolated worktree)

## QA Plan
- Visual: desktop/tablet/mobile — confirm single immersive lyrics header at top of rounded content pane, audio bar below timeline, no overlap or cut-off.
- Interaction: enter lyrics from release row play → audio L, from library, from chat context if any; Esc closes to correct return (library or explicit ?from); space toggles playback; L toggles close; warm nav no audio restart.
- Keyboard: predictable, no double handlers.
- Exhaustive via /qa --exhaustive + design review.

(Opened as draft per incremental shipping.)

### DIFF
--- a/apps/web/components/organisms/AuthShell.tsx
+++ b/apps/web/components/organisms/AuthShell.tsx
@@ -26,6 +26,7 @@ export interface AuthShellProps {
   readonly headerAction?: ReactNode;
   readonly showMobileTabs?: boolean;
   readonly isTableRoute?: boolean;
+  readonly isLyricsRoute?: boolean;
   readonly onSidebarOpenChange?: (open: boolean) => void;
   readonly sidebarDefaultOpen?: boolean;
   readonly children: ReactNode;
@@ -43,6 +44,7 @@ function AuthShellInner({
   headerAction,
   showMobileTabs = false,
   isTableRoute = false,
+  isLyricsRoute = false,
   children,
 }: Readonly<Omit<AuthShellProps, 'children'> & { children: ReactNode }>) {
   const { isMobile } = useSidebar();
@@ -54,6 +56,7 @@ function AuthShellInner({
   const sidebarTrigger = isMobile ? null : <SidebarTrigger />;
 
   const isInSettings = section === 'settings';
+  const hideTopHeader = isInSettings || isLyricsRoute;
 
   // Memoize the sidebar so it doesn't re-render on breadcrumb/header changes.
   // The sidebar only depends on `section` — it shouldn't remount when
@@ -85,7 +88,7 @@ function AuthShellInner({
     <AppShellFrame
       sidebar={sidebar}
       header={
-        isInSettings ? null : (
+        hideTopHeader ? null : (
           <DashboardHeader
             breadcrumbs={breadcrumbs}
             sidebarTrigger={sidebarTrigger}
--- a/apps/web/components/organisms/AuthShellWrapper.tsx
+++ b/apps/web/components/organisms/AuthShellWrapper.tsx
@@ -260,6 +260,7 @@ function AuthShellWrapperInner({
               headerAction={headerAction}
               showMobileTabs={config.showMobileTabs}
               isTableRoute={config.isTableRoute}
+              isLyricsRoute={config.isLyricsRoute}
               onSidebarOpenChange={
                 persistSidebarCollapsed ? handleSidebarOpenChange : undefined
               }
--- a/apps/web/hooks/useAuthRouteConfig.ts
+++ b/apps/web/hooks/useAuthRouteConfig.ts
@@ -22,6 +22,7 @@ export interface AuthRouteConfig {
   isChatRoute: boolean;
   isDemoRoute: boolean;
   showChatUsageIndicator: boolean;
+  isLyricsRoute: boolean;
 }
 
 function isChatThreadPath(parts: readonly string[]): boolean {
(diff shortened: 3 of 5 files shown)


itstimwhite:
## Review Summary

The diff is small, focused, and follows the established patterns for route-aware chrome decisions. However, two issues must be addressed: missing test coverage for the new `isLyricsRoute = true` path, and the existence of duplicate test files that will cause confusion and break future refactors.

---

### Critical Issues

- **`apps/web/tests/unit/components/AuthShellWrapper.test.tsx` and `apps/web/tests/unit/components/organisms/AuthShellWrapper.test.tsx`**  
  Both files exist after the PR, with essentially the same mock updates. This creates a maintenance hazard — one file is redundant, and any future change to the component’s test suite must be duplicated or risks ambiguity about which file is canonical.  
  **Action**: Remove one of the files (likely the one under `components/` to match the existing test location pattern `organisms/AuthShellWrapper.test.tsx`). The diff only touches mock return values; the PR should either delete the duplicate or add a comment if the two are intentionally different (they appear identical).

- **No test coverage for `isLyricsRoute = true`**  
  The new prop `isLyricsRoute` is only mock‑set to `false`. There are no tests verifying that:
  - The `DashboardHeader` is hidden when `isLyricsRoute` is `true`.
  - The correct `hideTopHeader` logic flows through `AuthShell` → `AuthShellInner`.
  - The sidebar and other chrome elements remain visible/unaffected.  
  Given this is a feature flag that conditionally removes a UI element, at minimum one test per component (`AuthShell` and `AuthShellWrapper`) should mount with `isLyricsRoute: true` and assert the header is null. See suggestions below for a concrete test structure.

---

### Warnings

- **`apps/web/hooks/useAuthRouteConfig.ts`**  
  The `isLyricsRoute` derivation assumes `APP_ROUTES.LYRICS` always exists. While likely true, a missing constant would silently break all lyrics route detection. Consider either ensuring the constant is defined or adding a defensive fallback (e.g., `pathname === '/lyrics'`). Not a blocker, but worth checking that `APP_ROUTES` is correctly imported.

- **Test mock duplication with existing tests**  
  If there are other test files that rely on `useAuthRouteConfig` mocks (e.g., in `tests/unit/hooks/`), they would also need the `isLyricsRoute: false` field to avoid TypeScript errors once the typing is updated. The PR only updated two test files, but other mocks elsewhere may now be typed incorrectly. Please run a global typecheck after the merge to catch any missed updates.

---

### Suggestions

- **Add a test for `isLyricsRoute = true`** (example for `AuthShellWrapper.test.tsx`):
  ```ts
  it('suppresses DashboardHeader when isLyricsRoute is true', () => {
    useAuthRouteConfigMock.mockReturnValue({
      ...baseConfig,
      isLyricsRoute: true,
    });
    render(<AuthShellWrapper>content</AuthShellWrapper>);
    // DashboardHeader is usually rendered inside a <nav> with role="navigation".
    // Or assert the header container is absent.
    expect(screen.queryByRole('navigation')).toBeNull();
  });
  ```
  Replace `baseConfig` with a shared config object in the test file to reduce duplication.

- **Prefer a single source of truth for test config**  
  Both test files define the same mock return object inline. Extracting a `defaultAuthRouteConfig` (with `isLyricsRoute: false`) would make future additions easier.

- **Confirm `APP_ROUTES.LYRICS` is a simple string** (no trailing slash) to avoid false negatives in the `startsWith` check. If the constant contains a trailing slash, the exact match `pathname === APP_ROUTES.LYRICS` would fail. This is unlikely but worth verifying.

---

### What Looks Good

- **Minimal change** – The PR touches only the necessary files and follows the exact pattern used for `isInSettings` / `isTableRoute`. This keeps the cognitive load low.

- **Type‑safe** – The new property is explicitly typed on `AuthRouteConfig` and `AuthShellProps`, and passed through the entire chain. No `any` or unsafe casts.

- **Test mocks updated** – The existing mock return values are extended with the new field, preventing TypeScript errors in tests when the type changes.

- **Route detection logic is correct** – The `startsWith` check with the dynamic slash guards the `/lyrics/123` sub‑route correctly, and the spelling of `isLyricsRoute` (note the **c** in *Lyrics*) matches the English convention.

- **Header suppression is isolated** – The `hideTopHeader` variable is computed once and directly controls the conditional, avoiding duplication of the `isInSettings` check in the markup.

Once the duplicate test file is removed and a basic test for `isLyricsRoute = true` is added (or at least a ticket created for follow‑up), this PR is ready to approve. The code itself is sound.

itstimwhite (commented):

PR #9500, opened 2026-05-21
fix(lint): prevent ESLint crash on module configs

itstimwhite:
## Summary
- Fix the ESLint 10 parser crash on module-style config/support files by routing `*.js`, `*.jsx`, `*.mjs`, `*.mts`, and `*.cts` through `@typescript-eslint/parser`.
- Normalize `apps/web/stryker.config.mjs` to a named export so the repo-wide lint gate does not trip on an anonymous default-export warning.

## Verification
- `pnpm --filter=@jovie/web exec eslint --no-cache scripts/analyze-issues.mjs --max-warnings=0`
- `pnpm --filter=@jovie/web exec eslint --no-cache vitest.config.storybook.mts --max-warnings=99999`
- `pnpm --filter=@jovie/web exec eslint --no-cache plopfile.js --max-warnings=0`
- `pnpm --filter web lint:eslint` now gets past the original parser crash and surfaces a separate pre-existing repo-wide lint baseline.

## Files Changed
- `apps/web/eslint.config.js`
- `apps/web/stryker.config.mjs`

### DIFF
--- a/apps/web/eslint.config.js
+++ b/apps/web/eslint.config.js
@@ -1,6 +1,7 @@
 const nextConfig = require('eslint-config-next');
 const nextCoreWebVitals = require('eslint-config-next/core-web-vitals');
 const boundariesPlugin = require('eslint-plugin-boundaries');
+const tsParser = require('@typescript-eslint/parser');
 const iconUsageRule = require('./eslint-rules/icon-usage');
 const edgeRuntimeNodeImportsRule = require('./eslint-rules/edge-runtime-node-imports');
 const noHandlerInitializationRule = require('./eslint-rules/no-handler-initialization');
@@ -402,6 +403,17 @@ module.exports = [
     ],
     rules: {
       '@jovie/no-ad-hoc-currency': 'off',
+    },
+  },
+  {
+    files: ['**/*.{js,jsx,mjs,mts,cts}'],
+    languageOptions: {
+      // Next's default parser crashes on module-style config and support files
+      // under ESLint 10.2.1. Reuse the TypeScript parser here so plain JS/ESM
+      // scripts keep lint coverage.
+      parser: tsParser,
+      ecmaVersion: 'latest',
+      sourceType: 'module',
     },
   },
 ];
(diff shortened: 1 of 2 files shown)


itstimwhite:
## Review Summary

### Critical Issues
- **apps/web/eslint.config.js:409-415** — The new ESLint config block sets the TypeScript parser for all `*.js`, `*.jsx`, `.mjs`, `.mts`, and `.cts` files but does **not** include `jsx: true` in `languageOptions.parserOptions`. This means any `.jsx` file (e.g., Next.js page or component) will fail to parse JSX syntax, breaking linting for all React components. The commit description only mentions needing to route module-style config/support files, not all application JSX files.  

  **Suggestion**: Either  
  1. Restrict the pattern to only config/support files (e.g., `['**/*.{mjs,mts,cts}', '**/stryker.config.js', '**/*.config.js']`) to avoid affecting React components, or  
  2. Add `jsx: true` to parserOptions so JSX syntax is supported for all JS/JSX files.

### Warnings
- **apps/web/eslint.config.js:410** — Setting `sourceType: 'module'` for all `.js` files (including those that may be CommonJS) could cause lint errors like "import and export may appear only with 'sourceType: module'". The Next.js ESLint config already handles source type per file; this override might break existing lint rules for `.js` files that use `require` or `module.exports`. Consider using `sourceType: 'script'` for `.js` files that are not ESM, or use a more targeted file pattern.

### Suggestions
- If the goal is to only fix the parser crash on config files, consider adding an `ignores` block to exclude JavaScript application files from this rule, or use a negative pattern like `["**/*.{js,jsx,mjs,mts,cts}", "!**/*.{mjs,js,jsx}"]`? Actually, the safest approach is to mirror the original crash: only apply the TypeScript parser to files that are known to cause the crash (`.mjs`, `.mts`, `.cts`, and maybe `.config.js`).

### What Looks Good
- The normalization of `stryker.config.mjs` from an anonymous default export to a named constant fixes the `no-anonymous-default-export` lint warning cleanly and is a harmless refactor.
- The PR addresses a real crash under ESLint 10.2.1 and provides a clear explanation in the diff comment.
- The `require('@typescript-eslint/parser')` import is correctly placed in the CommonJS module.

itstimwhite (commented):

itstimwhite:
Recorded gate evidence for PR 9500.

PR #9507, opened 2026-05-22
fix(security): require MFA for admin access

itstimwhite:
Implements the smallest safe MFA enforcement cut for GitHub issue #1425.

## What changed
- Added a tiny shared helper for Clerk reverification freshness checks: `hasRecentAdminMfaReverification()`.
- Admin entitlements now require both the existing admin role check and a fresh multi-factor reverification window.
- `requireAdmin()` now denies admin routes when reverification is stale, so route-level admin access cannot bypass the guard.
- Added focused tests for the entitlements path and admin middleware path.

## Verification
- `pnpm --filter web exec vitest run tests/unit/lib/entitlements.server.test.ts tests/unit/lib/admin/middleware.test.ts`
- Commit hook also ran `next:proxy-guard`, `tailwind:check`, `typecheck`, `biome check .`, and `biome lint .` successfully.

## Notes
- This keeps admin access consistent across the two trust boundaries that actually grant it: entitlements resolution and admin route guarding.

## Summary by CodeRabbit

* **New Features**
  * Admin users must re-verify multi-factor authentication within 15 minutes to access admin features; users failing this recency check are denied access.

* **Security / Access**
  * Admin checks and entitlement determination now account for recent MFA re-verification when granting admin status.

* **Tests**
  * Added and updated unit tests to cover MFA recency enforcement and related admin/entitlement behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JovieInc/Jovie/pull/9507?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/web/lib/admin/mfa.ts
+++ b/apps/web/lib/admin/mfa.ts
@@ -1,0 +1,31 @@
+import 'server-only';
+
+const ADMIN_MFA_AFTER_MINUTES = 15;
+
+type AuthWithHas = {
+  has?: (params: {
+    reverification?:
+      | 'strict_mfa'
+      | 'strict'
+      | 'moderate'
+      | 'lax'
+      | {
+          level: 'first_factor' | 'second_factor' | 'multi_factor';
+          afterMinutes: number;
+        };
+  }) => boolean;
+};
+
+export function hasRecentAdminMfaReverification(authResult: unknown): boolean {
+  const has = (authResult as AuthWithHas | null | undefined)?.has;
+  if (typeof has !== 'function') {
+    return false;
+  }
+
+  return has({
+    reverification: {
+      level: 'multi_factor',
+      afterMinutes: ADMIN_MFA_AFTER_MINUTES,
+    },
+  });
+}
--- a/apps/web/lib/admin/middleware.ts
+++ b/apps/web/lib/admin/middleware.ts
@@ -79,7 +79,7 @@ export async function requireAdmin(): Promise<NextResponse | null> {
   const userIsAdmin = await isAdmin(userId);
   const hasRecentMfa = authResult
     ? hasRecentAdminMfaReverification(authResult)
-    : true;
+    : false;
 
   // Mask user ID for logging to prevent PII exposure
   const maskedUserId = maskUserIdForLog(userId);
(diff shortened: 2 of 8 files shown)


itstimwhite:
## Review Summary
**Verdict**: Approve

The PR correctly enforces MFA reverification for admin access across the two trust boundaries (entitlements and route middleware). The implementation is minimal, well-tested, and follows Jovie conventions. No blocking issues found.

---

### Critical Issues
None.

### Warnings
- **apps/web/lib/entitlements/server.ts** — The `getCachedAuth()` function must return the full Clerk `Auth` object (with the `has` method) for the MFA check to work. The mock in the test includes `has`, but if the real `getCachedAuth` strips properties (e.g., only returns `{ userId }`), this code will silently mark all authenticated sessions as MFA-failed (since `has` will be `undefined`). A runtime guard (throwing a clear error) or a type assertion on the returned value would make this dependency explicit.

### Suggestions
1. **Use Clerk’s exported type for `has`**  
   `apps/web/lib/admin/mfa.ts:5-21` — Instead of declaring a custom `AuthWithHas`, import the Clerk `Auth` type (or the precise shape of `auth()`'s return value). This avoids drift when Clerk updates its type definition and improves IDE support.

2. **Add a test for `checkIsAdmin` with stale MFA**  
   `apps/web/tests/unit/lib/admin/middleware.test.ts` — The new tests cover `requireAdmin`. Adding a case for `checkIsAdmin` (the boolean variant) with stale MFA would close the gap and confirm both functions behave consistently.

3. **Document the `afterMinutes` constant**  
   `apps/web/lib/admin/mfa.ts:3` — Consider exporting `ADMIN_MFA_AFTER_MINUTES` or adding a brief comment explaining why `15` minutes was chosen (e.g., "aligns with existing session timeout policy for admin actions").

4. **Guard against missing `has` in `getCachedAuth`**  
   `apps/web/lib/entitlements/server.ts` — Add a runtime check (e.g., `if (typeof authResult.has !== 'function') throw new Error(...)`) to fail loudly if the cached auth object is incomplete, rather than silently denying admin access.

### What Looks Good
- **Clean separation of concerns**: The MFA logic lives in its own module (`lib/admin/mfa.ts`) rather than being inlined into middleware/entitlements.
- **Consistent enforcement**: Both `requireAdmin` and `getCurrentUserEntitlements` check the same helper, preventing bypasses.
- **Test bypass handling**: Non-Clerk authentication paths (test mode) correctly skip the MFA check.
- **Focused tests**: The two new test cases cover the positive and negative paths for `requireAdmin`, and the entitlement test is updated to validate the new behavior.
- **Conservative fallback**: If the `has` function is missing or throws, the access is denied (fail closed).

PR #9558, opened 2026-05-23
feat(auth): centralize auth routing and icon pipeline

itstimwhite:
## Summary
- Added `@jovie/auth-routing` plus Redis-backed `/auth/start`, `/auth/callback`, and native exchange routes so web, iOS, and Electron auth stay bound to the originating client.
- Reworked iOS auth to use `ASWebAuthenticationSession` with PKCE and a one-time native exchange, and reworked Electron auth to use browser handoff plus `jovie://auth/complete` and renderer-side ticket exchange.
- Replaced drifting icon sources with one deterministic brand asset generator for iOS, Electron, PWA/app icons, transparent in-app marks, and a separate favicon profile.

## Ship Decision
- Ship now: central auth boundary plus deterministic icon pipeline because both block a reliable internal alpha.
- Re-evaluate when: auth boundary tests, native smoke tests, icon validators, and CI app packaging are green against the release branch.
- Then: keep compatibility adapters active only while old installed clients can still callback through legacy params.

## Evidence
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/auth-routing run test` — 22 passed.
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/auth-routing run typecheck` — passed.
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/web exec vitest run tests/unit/lib/auth tests/unit/app/signin-page.test.tsx tests/unit/app/signup-page.test.tsx tests/unit/app/desktop-auth-page.test.tsx tests/components/auth/SsoCallbackHandler.test.tsx tests/unit/lib/mobile/auth-return.test.ts tests/unit/api/mobile/v1/auth-ticket.test.ts lib/desktop/auth-return.test.ts lib/desktop/electron-bridge.test.ts` — 447 passed.
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/web run typecheck -- --pretty false` — passed.
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/desktop run test` — passed.
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/desktop run typecheck` — passed.
- `JOVIE_AGENT_PROFILE=coder pnpm run brand:generate` — regenerated web, iOS, Electron, and contact-sheet assets.
- `JOVIE_AGENT_PROFILE=coder bash apps/ios/scripts/validate-app-icon.sh` — validated 1024 App Store icon and all iOS `Contents.json` sizes.
- `JOVIE_AGENT_PROFILE=coder pnpm run ios:test` — Swift tests passed 24/24; UI tests passed 11/11 with 3 existing skips.
- Pre-push hook — `biome check .`, web proxy guard, Tailwind guard, web typecheck, and web lint passed.

## Icon Contact Sheet
Jovie icon contact sheet

## Summary by CodeRabbit

* **New Features**
  * PKCE-backed native sign-in flows for desktop and mobile with coordinated callback handling.
  * Centralized auth callback routing so web, iOS, and desktop share sign-in state.
  * Desktop-native completion support (one‑time completion consumption) and server exchange endpoint to finalize native sign‑ins.

* **Bug Fixes**
  * Refined desktop auth UI copy (“Continue in browser”) and clearer mobile post‑sign‑in messaging.

* **Tests**
  * Expanded unit and E2E auth tests and improved retry/validation logic.

* **Chores**
  * Added brand asset generation and icon validation tooling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JovieInc/Jovie/pull/9558?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


itstimwhite:
## Review Summary

The PR introduces a well-structured centralized auth boundary and deterministic icon pipeline. However, two critical bugs in the iOS deep-link handling will break native auth flows, and the missing code verifier propagation prevents PKCE completion. Desktop changes look solid with proper PKCE and legacy fallback.

---

### Critical Issues

- **`apps/ios/Jovie/App/MobileAuthReturn.swift:39`** — `isSupportedCallback` checks `url.path == "/auth-return"` but the new auth completion deep link uses path `/complete`. This will reject **all** new iOS auth completions, making the PKCE flow inoperable.  
  **Fix**: Update the check to also accept `/complete` or switch to query-parameter–based detection:
  ```swift
  static func isSupportedCallback(_ url: URL) -> Bool {
      let components = URLComponents(url: url, resolvingAgainstBaseURL: false)
      return url.scheme == "jovie"
          && (url.path == "/auth-return" || url.path == "/complete")
          && components?.queryItems?.contains(where: { $0.name == "code" }) == true
  }
  ```

- **`apps/ios/Jovie/App/MobileAuthReturn.swift:11`** — `parse(_:codeVerifier:)` requires a non-nil `codeVerifier` to return a valid `MobileAuthReturn`. The caller (`RootView.swift` `handleAuthReturn`, line ~374) passes no code verifier – only the URL. Since the verifier is not stored/supplied, the parse always returns `nil` and auth never completes.  
  **Fix**: Save the `codeVerifier` per auth session (e.g., in `AppState` or a static session store) and pass it to `parse`.  
  Example in `RootView.swift`:
  ```swift
  @MainActor
  private func handleAuthReturn(_ url: URL) {
      let codeVerifier = appState.currentAuthCodeVerifier // must be stored before starting auth
      guard let authReturn = MobileAuthReturnParser.parse(url, codeVerifier: codeVerifier) else { return }
      // exchange code at /auth/native/exchange
  }
  ```

---

### Warnings

- **`apps/ios/Jovie/App/MobileAuthReturn.swift`** — The legacy `ticket`-based parser has been removed entirely. The PR description mentions “keep compatibility adapters active only while old installed clients can still callback through legacy params”, but the code drops it. If backward compatibility is still needed (e.g., users who haven’t updated), add an alternative parse path for the old `ticket` query parameter.

- **`apps/desktop/src/main.ts:1123`** — `consumePendingDesktopAuthPkce()` is called in `handleAuthCompletion`. If no PKCE is pending (e.g., the deep link is triggered directly without a prior auth start), the function returns silently and the user sees no error, nor any navigation. Consider showing an error page or logging a warning to help debugging.

- **`apps/desktop/src/main.ts:1084-1096`** — The IPC handler `CONSUME_DESKTOP_AUTH_COMPLETION_CHANNEL` returns the `completion` object to any renderer at `/auth/native-complete`. While the sender is validated, a malicious page loaded in the main window (if any) could attempt to invoke this. The design appears safe because only the native-complete page is expected to have the correct path, but consider adding a nonce or requiring that the renderer’s URL matches exactly.

- **Missing unit tests for new IPC handler** – The `CONSUME_DESKTOP_AUTH_COMPLETION_CHANNEL` handler in `main.ts` and its preload counterpart are not covered by the listed tests. Add a test that invokes the IPC from a trusted origin and asserts the correct completion payload is returned.

---

### Suggestions

- **`apps/desktop/src/main.ts:477-495`** — `loadAuthCompletion` loads a new URL in the main window and then closes `authHandoffWindow` if it exists. If the handoff window is still loading or has pending navigation, closing it might cause a brief flash or leave a partially rendered window. Consider waiting for the handoff window’s `did-finish-load` before closing, or simply hiding it.

- **`apps/desktop/scripts/gen-staging-icon.mjs` & `prepare-icons.mjs`** — These scripts now delegate entirely to `pnpm exec tsx scripts/generate-brand-assets.ts` via `spawnSync`. This makes the desktop icon build order dependent on the root-level brand generator. Consider keeping a lightweight, explicit icon pipeline for the desktop package to avoid cross-package coupling and reduce build times.

- **`apps/web/lib/auth/routing-state.server.ts` (not shown fully)** – If this file reads/writes Redis, ensure that the Redis key includes a unique session identifier to prevent replay or cross-session state conflict. The PR evidence suggests tests pass, but consider documenting the TTL and cleanup strategy.

---

### What Looks Good

- **Centralised PKCE auth routing** – The new `@jovie/auth-routing` package with Redis-backed routes (`/auth/start`, `/auth/callback`, `/auth/native-complete`) provides a clean boundary for web, iOS, and Electron auth flows, with proper SHA256 code challenge and verifier management.
- **Desktop legacy support** – The `parseLegacyAuthReturnRouteDeepLink` function and `pendingLegacyAuthReturnRoute` variable ensure older deep links still work while the transition is in progress.
- **IPC security** – `CONSUME_DESKTOP_AUTH_COMPLETION_CHANNEL` is gated by origin and path checks, and the completion is consumed once (set to `null` after retrieval).
- **Icon pipeline** – The deterministic brand asset generator and contact-sheet validation (opaque PNG, black corners) enforce a consistent icon profile across platforms.
- **Comprehensive test evidence** – 447 web tests, 24 Swift unit tests, 11 UI tests, and icon contract tests all pass, demonstrating good coverage for the new code.

itstimwhite (commented):

itstimwhite:
## Review Summary

This is a substantial PR that introduces centralized auth routing and a deterministic icon pipeline. While the architecture is sound and the test coverage is impressive, there are several critical issues that need addressing before merge.

---

### Critical Issues

- **apps/desktop/src/main.ts:1083-1098** — Race condition in `CONSUME_DESKTOP_AUTH_COMPLETION_CHANNEL` handler. The handler reads `pendingAuthCompletion` without any locking mechanism. If the renderer process calls this IPC handler concurrently (e.g., due to React StrictMode double-invocation in development), the second call will receive `{ ok: false, reason: 'missing-auth-completion' }` even though the first call consumed it. This could cause the auth flow to silently fail. **Fix**: Use a mutex or flag to ensure single consumption, or return the completion data to all callers until explicitly acknowledged.

- **apps/desktop/src/main.ts:230-240** — `base64Url` function doesn't handle edge cases where the input buffer is empty. `randomBytes(32)` will never produce an empty buffer, but the function should still guard against this for safety. More critically, the PKCE code verifier is generated with exactly 32 bytes (256 bits of entropy), but RFC 7636 recommends 43-128 characters for the code verifier. 32 bytes produces ~43 base64url characters, which is at the minimum boundary. **Fix**: Use `randomBytes(64)` to generate a 64-byte verifier (~86 characters) for better entropy and RFC compliance.

- **apps/web/lib/auth/central-auth-routing.ts** — The central auth routing module appears to be missing from the diff (only the file path is listed). This is a core piece of the auth flow and cannot be reviewed. **Action**: Ensure this file is included in the PR diff or provide its contents for review.

- **apps/desktop/src/main.ts:477-500** — `loadAuthCompletion` navigates to `/auth/native-complete?client=electron&state=...` but doesn't pass the `code` parameter. The renderer page at `/auth/native-complete` needs the `code` to exchange it via the IPC bridge. Without it, the renderer cannot complete the auth flow. **Fix**: Add `code` to the URL search params: `targetUrl.searchParams.set('code', completion.code)`.

- **apps/ios/Jovie/App/MobileAuthReturn.swift:5-8** — The `MobileAuthReturn` struct stores `codeVerifier` in memory. If the app is backgrounded and the system reclaims memory, this value is lost. The PKCE verifier must survive app termination to complete the auth exchange when the user returns from the browser. **Fix**: Store the `codeVerifier` in the iOS Keychain or UserDefaults (encrypted) before opening the auth session.

---

### Warnings

- **apps/desktop/src/main.ts:349-375** — `handleAuthCompletion` silently returns if `consumePendingDesktopAuthPkce()` returns null. This could happen if the user opens the deep link after the PKCE challenge has expired or was never created. There's no user-facing feedback. **Suggestion**: Log a warning and potentially show an error state in the UI.

- **apps/desktop/src/main.ts:1162-1170** — The `pendingLegacyAuthReturnRoute` is consumed but `pendingAuthCompletion` is not cleared when `pendingLegacyAuthReturnRoute` is used. If both are set (unlikely but possible during migration), the legacy route takes precedence. **Fix**: Add `pendingAuthCompletion = null` in the legacy branch for clarity.

- **apps/desktop/scripts/gen-staging-icon.mjs** — This script now delegates to `scripts/generate-brand-assets.ts` via `spawnSync`. This creates a dependency on the brand generator being available and working. If the brand generator fails, the staging icon generation silently fails (the `throw result.error` only catches spawn errors, not non-zero exit codes). **Fix**: Check `result.status` explicitly and throw a descriptive error.

- **apps/web/app/api/mobile/v1/auth/ticket/route.ts** — The legacy mobile ticket endpoint is deprecated but still active. The PR description mentions keeping compatibility adapters "only while old installed clients can still callback through legacy params," but there's no documented deprecation timeline or sunset plan. **Suggestion**: Add a deprecation header or logging to track legacy usage.

- **apps/desktop/src/main.ts:121-127** — `isTrustedDesktopAuthCompleteSender` checks the sender's URL origin and pathname. However, the sender URL could be spoofed if the renderer process is compromised. **Suggestion**: Add additional verification, such as checking the sender's frame hierarchy or using a nonce.

---

### Suggestions

- **apps/desktop/src/main.ts:230-240** — Consider extracting `base64Url` to a shared utility module since it's also needed in the renderer for PKCE verification.

- **apps/desktop/src/main.ts:477-500** — Add a timeout mechanism for `pendingAuthCompletion`. If the renderer never calls `consumeDesktopAuthCompletion`, the completion data sits in memory indefinitely. A 5-minute TTL would be reasonable.

- **apps/ios/Jovie/App/MobileAuthReturn.swift** — The `isSupportedCallback` function checks `url.path == "/auth-return"` but the new flow uses `jovie://auth/complete`. This check should be updated to match the new path.

- **apps/web/app/auth/start/route.ts** — Consider adding rate limiting to the auth start endpoint to prevent abuse (e.g., rapid PKCE challenge generation).

- **apps/desktop/src/main.ts:1083-1098** — Add structured logging to the IPC handler for debugging auth flow issues in production.

- **apps/web/lib/desktop/electron-bridge.ts** — The `consumeDesktopAuthCompletion` method returns a raw IPC response. Consider adding TypeScript type narrowing to ensure callers handle the `ok: false` case properly.

---

### What Looks Good

1. **Comprehensive test coverage** — 447 passing tests across web, desktop, and iOS is excellent. The test suite covers the new auth flow extensively.

2. **Clean migration path** — The legacy auth return route is preserved alongside the new PKCE-based flow, allowing for a smooth transition without breaking existing clients.

3. **Deterministic icon pipeline** — Centralizing icon generation to a single brand asset generator eliminates the drift between platforms. The contact sheet is a nice addition for visual verification.

4. **Security-conscious design** — PKCE implementation follows RFC 7636, IPC channels validate sender origins, and the deep link protocol uses proper state validation.

5. **TypeScript strictness** — The codebase maintains strong typing throughout, with proper interfaces for auth completion payloads and PKCE state.

6. **Documentation** — The PR description and commit message clearly explain the motivation, architecture, and migration strategy.

itstimwhite (commented):

itstimwhite:
## Review Summary
**Verdict**: Changes Requested (one critical security concern + one potential regression)

### Critical Issues
1. **apps/desktop/scripts/desktop-icon-contract.test.mjs (line ~95)**  
   The test `'packaged production and staging icons are generated from the canonical profile'` asserts:
   ```js
   assert.deepEqual(stagingIcon, expectedProductionIcon);
   ```
   This means the staging icon is now **identical** to the production icon — the orange “S” badge is gone.  
   - **Why it matters**: Staging builds lose visual distinguishability. Users, testers, and CI systems will no longer see the “S” badge that flags pre-release builds. This is a regression from the current behaviour where `gen-staging-icon.mjs` explicitly added a badge.  
   - **Fix**: Either re-add staging badge logic inside the shared `generate-brand-assets.ts` script, or split the desktop icon generation into two steps (production from canonical, staging from canonical + badge composite). Update the test to verify the badge exists (e.g., `await assertBlackCorners` is insufficient — check for orange circle in bottom-right).

2. **apps/desktop/src/main.ts (IPM channel `CONSUME_DESKTOP_AUTH_COMPLETION_CHANNEL`)**  
   The handler returns the full `DesktopAuthCompletion` object including `codeVerifier` to any caller that passes the `isTrustedDesktopAuthCompleteSender` check.  
   - **Why it matters**: The `codeVerifier` is a secret value that, combined with the `code` (authorization code), could allow an attacker to exchange the code for tokens if they also intercept the deep link. While the receiver is restricted to the `/auth/native-complete` page, the IPC is synchronous and the completion object remains in memory until consumed.  
   - **Fix**:  
     - Destructure the `codeVerifier` out of the completion before returning to the renderer. The renderer only needs `code` and `state` to send the exchange request to the backend (which should retrieve `codeVerifier` from a server-side session tied to `state`).  
     - Alternatively, keep the `codeVerifier` only on the main process and have the renderer request the exchange via another IPC call that internally uses the stored verifier. This avoids leaking the secret to the renderer at all.

3. **apps/ios/Jovie/App/MobileAuthReturn.swift (all locations)**  
   The `parse` function now requires a `codeVerifier` parameter, but no mechanism is shown in the provided diff to **persist the PKCE verifier** generated before launching the web auth URL.  
   - **Why it matters**: Without storing the `codeVerifier` generated during `ASWebAuthenticationSession` setup, the `parse` function will always return `nil` because `verifier` will be `nil`. The iOS auth flow will be broken.  
   - **Fix**: Ensure the caller (likely `AuthScreen` or `RootView`) stores the PKCE verifier in a state variable (e.g., `@State` or `@Bindable` in the app state) and passes it to `MobileAuthReturnParser.parse` when the callback URL arrives. The diff for `RootView.swift` is truncated — verify that `handleAuthReturn` reads the stored verifier before calling `parse`.

### Warnings
1. **apps/desktop/scripts/gen-staging-icon.mjs & prepare-icons.mjs**  
   Both scripts now simply call `pnpm exec tsx scripts/generate-brand-assets.ts`. This assumes the central script handles all desktop icon generation. If the central script does not incorporate the staging badge, the staging icon will be identical to production (see Critical #1).  
   - **Action**: After confirming Critical #1 is fixed, ensure these scripts either call a dedicated staging sub-script or pass a `--variant staging` flag to the generator.

2. **apps/web/lib/desktop/auth-return.ts (not shown in diff)**  
   The PR adds a new PKCE exchange flow but still maintains legacy `jovie://auth-return` handling. If the legacy handling is kept for backwards compatibility, ensure the legacy path still works correctly for already-installed clients. The `main.ts` legacy code looks correct, but no tests for the legacy path are visible in the diff.  
   - **Suggestion**: Add a unit test for `parseLegacyAuthReturnRouteDeepLink` and `handleLegacyAuthReturnRoute` to prevent accidental breakage.

3. **apps/ios/Jovie/App/MobileAuthReturn.swift: line 14**  
   The helper `isSupportedCallback` checks `url.path == "/auth-return"`. With the new PKCE flow, the callback path will be `/auth/complete` (from the deep link scheme). This predicate will **reject all new PKCE callbacks**, breaking iOS authentication.  
   - **Fix**: Update `isSupportedCallback` to also accept `/auth/complete`:
     ```swift
     return url.path == "/auth-return" || url.path == "/auth/complete"
     ```

### Suggestions
1. **apps/desktop/src/main.ts**  
   - The `buildCentralDesktopAuthUrl` function generates a new PKCE on every call, even if the user is already mid-flow (e.g., reclicking sign-in). This is fine, but consider cleaning up expired PKCE records to avoid memory leaks. A TTL map with 5-minute expiry would be defensive.  
   - The `desktop-shell-contract.test.mjs` change from `--electron-titlebar-height: 52px` to `40px` appears unrelated to the auth/icon changes — verify this is intentional.

2. **apps/desktop/src/preload.ts**  
   - The exposed `consumeDesktopAuthCompletion` function returns the `completion` object directly to the renderer. As noted in Critical #2, consider stripping `codeVerifier` before returning.

3. **Testing**  
   - The PR includes many passing tests, but no new tests for the Electron PKCE IPC flow (`CONSUME_DESKTOP_AUTH_COMPLETION_CHANNEL`). Consider adding a unit test that mocks `ipcMain.handle` and simulates a PKCE completion.  
   - The iOS UI tests passed, but ensure the PKCE change is covered by the existing 11 UI tests (state restoration, deep link handling). If not, add one.

### What Looks Good
- **Security**: The Electron main process uses PKCE challenge/verifier stored securely on the main process, and the IPC sender validation (`isTrustedDesktopAuthCompleteSender`) is correctly scoped to the exact path `/auth/native-complete`.  
- **Centralisation**: The icon generation consolidation into a single script reduces drift and ensures consistent branding across platforms. The desktop icon tests now validate opacity and corner colours from the canonical source.  
- **Backward compatibility**: Legacy `jovie://auth-return` handling is preserved, allowing old installed clients to continue working. The `parseLegacyAuthReturnRouteDeepLink` and `handleLegacyAuthReturnRoute` functions cleanly separate old and new flows.  
- **Type safety**: All new interfaces (`DesktopAuthCompletion`, `PendingDesktopAuthPkce`) are defined with `readonly` and proper types.  
- **Test coverage**: The PR shows 447+ passing tests, indicating thorough testing across the web, iOS, and Electron codebases. The `desktop-url-disposition.test.ts` correctly lists new auth routes as in-app or blocked.

itstimwhite (commented):

itstimwhite:
## Review Summary

Several critical issues must be addressed before shipping, primarily around the missing PKCE code verifier storage on iOS, the lack of state parameter validation on desktop, and incomplete auth return parsing on iOS. These could either break the auth flow silently or expose security weaknesses.

---

### Critical Issues

1. **iOS: `MobileAuthReturnParser.parse` always fails because no `codeVerifier` is supplied**  
   **File**: `apps/ios/Jovie/App/MobileAuthReturn.swift:15-28`  
   The new `parse` method requires a `codeVerifier` argument (defaults to `nil`). The guard on line 25-27 (`guard let verifier, !verifier.isEmpty`) will fail when `codeVerifier` is `nil`, causing every auth return URL to be ignored. The only call site in the diff (`LiveRootContainer.handleAuthReturn`) does not pass a code verifier. The iOS app must generate and store a PKCE code verifier before opening the auth URL (analogous to the desktop `pendingDesktopAuthPkce` variable) and then supply it to `MobileAuthReturnParser.parse`.  
   **Fix**: Store the code verifier in a property (e.g., in `LiveRootContainer` or in `AppState`) before initiating the auth flow, and pass it into `parse(_:codeVerifier:)`.

2. **Desktop: No state parameter is stored or validated during PKCE auth**  
   **File**: `apps/desktop/src/main.ts`  
   - In `createDesktopAuthPkce()` (lines 284-292), only `codeVerifier` and `codeChallenge` are stored. The `state` parameter (used by the OAuth server to bind the callback to the original request) is not generated or remembered.  
   - In `handleAuthCompletion` (line 456-469), the received `state` from the deep link is **discarded** – only `codeVerifier` is merged into the completion.  
   - The renderer’s `consumeDesktopAuthCompletion` returns the state but there is no mechanism to verify it matches the expected state.  
   This missing state binding allows a potential CSRF / injection attack where an attacker intercepts the deep link and swaps the `code`.  
   **Fix**: Add a `state` field to `PendingDesktopAuthPkce`, generate a random state during `createDesktopAuthPkce`, and validate the incoming state against it in `handleAuthCompletion`. Reject completion if mismatch.

3. **iOS: `MobileAuthReturn` struct no longer provides a route, but callers may rely on it**  
   **File**: `apps/ios/Jovie/App/MobileAuthReturn.swift:3-6`, `apps/ios/Jovie/App/RootView.swift:265`  
   The old `MobileAuthReturn` had a `ticket` and `route`. Now it has `code`, `state`, `codeVerifier`. The `onAuthReturn` closure is typed `(MobileAuthReturn) -> Void`, but after parsing the URL the iOS app must exchange the code for tokens using the code verifier. The diff does not show how this exchange is triggered. If the exchange logic is missing, the user will never be authenticated.  
   **Fix**: Ensure the `handleAuthReturn` function in `LiveRootContainer` performs the native token exchange (HTTP POST to the server with `code`, `code_verifier`, `state`), and confirm with the test evidence that `auth-ticket.test.ts` now covers this flow.

4. **Desktop: `consumeDesktopAuthCompletion` race condition**  
   **File**: `apps/desktop/src/main.ts:492-498`  
   The IPC handler for `CONSUME_DESKTOP_AUTH_COMPLETION_CHANNEL` is only trusted when the sender’s URL is exactly `/auth/native-complete`. However, the renderer may call this handler before the page navigation is complete, resulting in the URL mismatch and a silent failure. The web page should wait for the navigation to finish or the handler should also accept a short grace period.  
   **Fix**: On the renderer side, ensure the request is made only after DOM ready or after confirming `window.location.pathname === '/auth/native-complete'`. Alternatively, allow the IPC handler to accept requests from any page if `pendingAuthCompletion` is set and the request is fresh.

5. **Desktop: Legacy auth return handling may ignore stale PKCE**  
   **File**: `apps/desktop/src/main.ts:506-508`  
   When a legacy deep link (`jovie://auth-return?route=...`) arrives, the code no longer clears `pendingDesktopAuthPkce`. If a user first opens a new auth start (storing PKCE) but then uses an old return link, the pending PKCE will remain and may be consumed later by a valid completion, but then the old valid completion would fail because PKCE is consumed. This could cause confusing failures.  
   **Fix**: In `handleLegacyAuthReturnRoute`, consider clearing any stale `pendingDesktopAuthPkce`.

6. **Missing test coverage for desktop PKCE flow**  
   The evidence mentions 447 tests passed, but the critical new paths (e.g., `buildCentralDesktopAuthUrl`, `createDesktopAuthPkce`, `consumeDesktopAuthCompletion`) are not visible in the diff. Given the complexity of the PKCE handshake, dedicated unit tests for the main process functions should be added to ensure correctness.  
   **Suggestion**: Add tests in `apps/desktop/src/__tests__/main-auth.test.ts`.

---

### Warnings

- **Desktop: No timeout on pending PKCE** – `pendingDesktopAuthPkce` lives forever once stored. Should be cleared after, e.g., 10 minutes to avoid reuse of stale values.
- **iOS: `MobileAuthReturnParser.parse` requires codeVerifier but the function signature implies it’s optional** – Consider making the parameter non-optional to force callers to supply it, reducing the chance of the nil bug.
- **Desktop: `loadAuthCompletion` does not handle the case where `authHandoffWindow` is already closed** – The `if (authHandoffWindow && !authHandoffWindow.isDestroyed())` block is safe, but there’s no explicit close of the auth handoff window if something else already closed it.
- **Desktop: `findAuthReturnInArgv` may not work on Windows with protocol handling** – On Windows, custom protocol launches often pass the URL as a single string, not split into separate `argv` entries. The code scans `argv` for any URL, which should work, but double-check with Windows electron-builder defaults.

---

### Suggestions

- **Desktop: Log PKCE lifecycle** – Add `console.log` (or the appropriate Electron logger) for when PKCE is created, consumed, and when completion is handled, to aid debugging.
- **Desktop: Use a Map for pending PKCE** – Instead of a single global slot, use a map keyed by `state` to support multiple concurrent auth attempts.
- **iOS: Use `UIApplication.shared.open` sessions** – Ensure `ASWebAuthenticationSession` is used correctly with the new centralized auth URL and that the callback URL scheme is registered.
- **iOS: Add a timeout for the stored code verifier** – Clear it after a reasonable time (e.g., 10 minutes) to prevent reuse.

---

### What Looks Good

- **Centralized icon generation** – The delegation to a single brand assets script removes duplication and ensures all icon variants come from the same source.
- **Backward compatibility** – Legacy deep link handling is preserved with the `LEGACY_AUTH_RETURN_HOST` constant, and the code clearly separates old and new paths.
- **IPC security** – The attempt to verify the sender URL for the native complete channel is a good security practice.
- **Type safety** – Strong TypeScript types for PKCE state, authentication completion, and IPC messages.
- **Test evidence** – Large number of tests passed, indicating the code was run against existing test suites before PR submission.

itstimwhite (commented):

itstimwhite:
## Review Summary

### Critical Issues

- **apps/ios/Jovie/App/RootView.swift (truncated diff, line ~375)**—**iOS compilation error**  
  The new `MobileAuthReturn.parse(_:codeVerifier:)` signature requires an explicit `codeVerifier` parameter, but the call at the bottom of `handleAuthReturn` appears to be `MobileAuthReturn.parse(url)` only. The diff cuts off at `guard let authReturn = MobileAut`, strongly suggesting the parameter is missing. This will fail to compile and break iOS auth entirely.  
  *Fix*: Pass the stored `codeVerifier` (from the PKCE session) when parsing the deep‑link URL. The iOS code must track the codeVerifier generated during the auth start, or else provide it via another mechanism.

- **apps/desktop/src/main.ts: `handleAuthCompletion` (line ~505)**—**Missing PKCE state validation and silent failure**  
  1. The function consumes `pendingDesktopAuthPkce` but never validates that `completion.state` matches the state that was originally sent to the OAuth provider. While the server should verify state, a local check prevents replay of a deep‑link URL that was intercepted after the PKCE session expired.  
  2. If `pendingDesktopAuthPkce` is `null` (e.g., the user never started an auth flow, or the PKCE expired), the function silently returns without showing any error or indication to the user. This results in a blank `/auth/native-complete` page.  
  *Fix*: Store the expected `state` alongside the `codeVerifier` in `PendingDesktopAuthPkce`. In `handleAuthCompletion`, compare `completion.state` against the stored state (and consume it atomically). If the PKCE record is missing or the state mismatches, navigate to an error page or show a user‑facing message.

- **apps/desktop/src/main.ts: Potential loss of auth completion on second-instance race**  
  The `second-instance` event can fire while `pendingAuthCompletion` is already set (e.g., the user opens a new deep link before the `/auth/native-complete` page has consumed it). The code simply overwrites `pendingAuthCompletion`, losing the first completion. The user may not be authenticated when they expect to be.  
  *Fix*: Either queue completions or reject the new one if a pending completion exists (and log/warn). Since the completion is one‑time use, the second deep link should probably be ignored or cause a re‑authentication.

- **apps/desktop/src/main.ts: No timeout or cleanup for `pendingDesktopAuthPkce`**  
  If the user starts a desktop auth flow but never returns (e.g., closes browser), `pendingDesktopAuthPkce` persists forever. This is a memory leak and could cause stale PKCE to be consumed if a deep link arrives much later, potentially allowing a past auth code to be exchanged with an old verifier (depending on server expiration).  
  *Fix*: Add a TTL (e.g., 10 minutes) and clear `pendingDesktopAuthPkce` after expiry. Alternatively, couple it to the auth window lifecycle.

### Warnings

- **security: `CONSUME_DESKTOP_AUTH_COMPLETION_CHANNEL` IPC guard rationale**  
  The guard `isTrustedDesktopAuthCompleteSender` checks only the pathname `/auth/native-complete`. An attacker who can navigate to that page in the Electron renderer could call `consumeDesktopAuthCompletion` and exfiltrate the PKCE data. Currently the origin is trusted via `isTrustedAppOrigin()` in the preload, but a future XSS on any Electron page could exploit this if navigation is allowed.  
  *Suggestion*: Also check the origin is exactly `APP_ORIGIN`, and consider using a one‑time nonce that the renderer receives only after the deep link is processed (similar to state validation). This is low‑risk now but worth noting.

- **apps/web/app/auth/start/route.ts (not shown in diff)**—**Missing rate limiting**  
  The central auth start endpoint is unauthenticated and could be abused to generate many PKCE challenges or redirect loops. Given the PR title mentions Redis‑backed auth routing, ensure rate limiting (e.g., per IP per minute) is applied to `/auth/start` and `/auth/native-complete` to prevent resource exhaustion.

- **apps/desktop/scripts/gen-staging-icon.mjs & prepare-icons.mjs**—**Dependency on global `pnpm` and `tsx`**  
  These scripts now `spawnSync('pnpm', ['exec', 'tsx', ...])`. This assumes `tsx` is installed and that the working directory is the repo root. If the script is run from a different `cwd`, the relative `cwd: repoRoot` may be wrong? The `spawnSync` uses `cwd: repoRoot`, so that’s fine. However, spawning a subprocess to run a TypeScript script adds overhead (~1 second) and flattens error messages. Prefer to extract the shared logic into a JS module that both scripts import directly.

- **iOS: `MobileAuthReturnParser.isSupportedCallback` still checks `url.path == "/auth-return"` (line ~44)**  
  This path is the *legacy* auth return path. The new flow uses `jovie://auth/complete`. The `isSupportedCallback` function does not recognize the new scheme. Unless the iOS side also handles the new deep link URL, this will fail to parse the modern callback.  
  *Fix*: Update the condition to accept `jovie://auth/complete` (the host is `auth`, path `/complete` as seen in the desktop `parseAuthReturnDeepLink`).

### Suggestions

- **apps/web/lib/auth/central-auth-routing.ts**—**Consider documenting the state and code_verifier persistence**  
  The PR description mentions Redis‑backed state, but the diff does not show the server‑side validation. Ensure the `/auth/start` route stores the `state` and `code_challenge` in a Redis session with a TTL equal to the OAuth authorization code lifetime (e.g., 10 minutes). The `/auth/native-complete` exchange must verify that the `state` matches and that the `code_verifier` corresponds to the stored `code_challenge`.

- **tests: Add edge‑case tests for desktop PKCE flow**  
  Specifically:  
  - Two concurrent second‑instance events.  
  - Deep link arriving with no pending PKCE (`pendingDesktopAuthPkce` null).  
  - State mismatch between deep link and stored state.  
  - Expired PKCE (simulate by advancing time).  
  - IPC call from untrusted origin (should be rejected).

- **apps/desktop/src/main.ts: `findAuthReturnInArgv` should validate URL more strictly**  
  Currently it checks `protocol`, `hostname`, and `pathname` but does not verify the query params `code` and `state` are non‑empty. The assignment `parsed.searchParams.get('code')` returns `null` if missing; the guard `if (!code || !state) return null` already handles that. Good.

- **apps/ios/Jovie/App/MobileAuthReturn.swift: Remove stale `sanitizeReturnRoute`**  
  The dead code for legacy route sanitization is now removed (per the diff). Good. Ensure it’s fully deleted.

### What Looks Good

- **Centralized auth start endpoint**: Consolidates PKCE and redirect logic in one place, making cross‑platform support easier to maintain.
- **Deterministic brand pipeline**: Replacing scattered icon sources with a single generator reduces drift and makes validation contracts simple (`assertOpaquePng`, `assertBlackCorners`).
- **Legacy compatibility**: The desktop code gracefully handles old `jovie://auth-return` deep links with `findLegacyAuthReturnRouteInArgv`, allowing existing installed clients to continue working.
- **IPC security**: Both desktop and preload check origin/context before exposing sensitive channels.
- **Test evidence**: Many test suites passing indicates thorough unit/integration coverage for the core changes.

itstimwhite (commented):

itstimwhite:
## Review Summary
**Verdict**: Comment (with several concerns that should be addressed before merging)

The PR introduces a centralized PKCE-based auth routing for desktop and iOS clients, along with a deterministic icon pipeline. The overall direction is strong, but I found several potential bugs, missing security validations, and incomplete test coverage that warrant attention.

**Important note**: The provided diff is truncated (only the first ~350 lines of a 178KB diff). My review is based on visible changes and inferred missing parts. A full review of all files is required before merging.

---

### Critical Issues

1. **Missing validation in `apps/web/lib/auth/central-auth-routing.ts` (inferred)**  
   This file is not fully visible. It likely generates PKCE challenge and state. Ensure it uses `crypto.getRandomValues()` (not `Math.random()`) for randomness, and that state is stored in Redis with an appropriate TTL (e.g., 10 minutes). Without seeing the code, I can't verify. Please confirm.

2. **`apps/desktop/src/main.ts` – Undefined function `matchesPathPrefix`**  
   Line 307-310:  
   ```typescript
   const intent =
     matchesPathPrefix(parsed.pathname, '/signup') ||
     matchesPathPrefix(parsed.pathname, '/sign-up')
       ? 'sign_up'
       : 'sign_in';
   ```
   `matchesPathPrefix` is not defined or imported in the visible diff. If it’s missing, this will throw a runtime error. **Action**: Define or import `matchesPathPrefix` (presumably similar to `matchesPathPrefix` used elsewhere in the codebase, but not shown).

3. **`apps/desktop/src/main.ts` – Potential race condition in `loadAuthCompletion`**  
   Lines 480-494:  
   ```typescript
   function loadAuthCompletion(completion: DesktopAuthCompletion): void {
     pendingAuthCompletion = completion;
     // ... load URL and show window
     if (authHandoffWindow && !authHandoffWindow.isDestroyed()) {
       authHandoffWindow.close();
     }
   }
   ```
   `pendingAuthCompletion` is set before the window is loaded. If the renderer calls `consumeDesktopAuthCompletion` before the new page’s preload script runs (e.g., due to navigation timing), the completion might be consumed prematurely. **Recommendation**: Only set `pendingAuthCompletion` after the window has loaded the target URL, or use a state flag gated by the renderer’s readiness.

4. **`apps/desktop/src/main.ts` – PKCE state/codeVerifier not cleared on IPC consumption failure**  
   In `consumeDesktopAuthCompletion`, the handler clears `pendingAuthCompletion` only when returning with `ok`. If the IPC call fails (e.g., sender not trusted), the pending completion remains and could be consumed later by a malicious page that navigates to `/auth/native-complete`. **Fix**: Always clear `pendingAuthCompletion` after a consume attempt, regardless of success, and return an error if already consumed.

5. **`apps/web/app/api/auth/native/exchange/route.ts` (not visible) – CSRF & replay risks**  
   This endpoint likely exchanges an auth code for a token. Without seeing the code, I can’t confirm, but common issues:
   - Missing CSRF token (though same‑origin requests with `fetch` are generally safe if using HTTP‑only cookies).
   - No rate limiting: an attacker could brute‑force verifiers or replay stale codes.
   - No validation that the `code` and `state` were issued for the same `client_id`. **Action**: Add checks against Redis state storage, enforce a single‑use code, and set a short expiry (e.g., 2 minutes) on stored state.

6. **`apps/ios/Jovie/App/MobileAuthReturn.swift` – Missing codeVerifier storage**  
   The `parse` method now requires `codeVerifier`. The caller (`LiveRootContainer.handleAuthReturn`) must pass the verifier obtained from the `ASWebAuthenticationSession` (typically stored in a property). The diff shows the function call is truncated; if the verifier is not properly linked, the exchange will fail silently. **Verify** that the iOS app stores the verifier from the initial PKCE session and passes it here.

7. **`apps/web/public/brand/app-icons/` – Large binary assets committed**  
   The PR adds many PNG files (1024, 512, 192, etc.). While they are generated by the brand script, committing large binaries can bloat the repo. Ensure the script is run in CI and the assets are only committed if they change, or set up `.gitattributes` to use Git LFS. (Warn for now, not critical.)

---

### Warnings

1. **`apps/web/app/mobile-auth-return/page.tsx` (not visible) – Legacy endpoint still active**  
   The PR says the old ticket‑based flow is deprecated, but the page may still exist. If it isn’t guarded or hasn’t been updated to warn users, old iOS/Electron clients might still hit it and fail. **Recommendation**: Add a deprecation notice and redirect to the new flow for any client that can support it.

2. **`apps/web/lib/auth/routing-state.server.ts` – Redis dependency without fallback**  
   Using Redis for ephemeral auth state is reasonable, but if Redis is down, auth will break entirely. Consider a fallback to encrypted cookies or a Postgres‑backed store (convention: single DB driver). At minimum, add readiness checks and graceful error messages.

3. **`apps/desktop/src/preload.ts` – `consumeDesktopAuthCompletion` types rely on IPC**  
   The function returns a `Promise<{ ok: boolean; reason?: string; completion?: { ... } }>`. If the IPC handler returns an unexpected shape, the renderer might throw. **Recommendation**: Add runtime type validation or at least log unexpected responses.

4. **`apps/web/lib/desktop/electron-bridge.test.ts` – Test count discrepancy**  
   The PR says “447 passed” for a large set of test files. Ensure the new `electron-bridge` and `auth-return` modules are fully covered, especially edge cases like missing IPC, multiple rapid calls, and error recovery.

5. **Icon contact sheet test is weak**  
   `apps/desktop/scripts/desktop-icon-contract.test.mjs` line 100:  
   ```javascript
   assert.ok((metadata.width ?? 0) >= 900);
   assert.ok((metadata.height ?? 0) >= 180);
   ```
   This only checks minimum dimensions, not that the contact sheet is actually a tiled grid of all required icon sizes. Consider validating that specific tiles exist at expected positions.

---

### Suggestions

1. **Add E2E test for desktop auth flow**  
   The PKCE handoff is complex; add a Playwright test that simulates the Electron main process (using Electron’s testing utilities) or at least a mocking test that verifies IPC messages and navigation.

2. **Immutable migration comment**  
   The PR does not touch migration files, which is compliant. However, if any schema changes were required for the new auth flow (e.g., to store PKCE state in Postgres), ensure they are in a separate, never‑modified migration.

3. **Document the new PKCE state lifecycle**  
   A short ADR or README in `packages/auth-routing` explaining how state, challenge, and verifier flow from client to web to Redis to exchange endpoint.

4. **Consider using `crypto.subtle` in web routes**  
   For Node.js compatibility, `crypto.createHash` is used in Electron. For web routes, use the Web Crypto API (`crypto.subtle.digest`) to avoid platform differences.

5. **Remove `console.log` from scripts**  
   `apps/desktop/scripts/gen-staging-icon.mjs` runs the brand script via `spawnSync`. The brand script likely prints progress; that’s fine, but ensure no stray debug logs land in production output.

---

### What Looks Good

- **PKCE adoption**: Strong security improvement over the old ticket‑based flow for desktop and iOS.
- **Centralized routing**: `/auth/start` and `/auth/callback` routes with a consistent `client` parameter simplify state management.
- **Backward compatibility**: Legacy `jovie://auth-return` deep links are still supported, reducing breakage for installed clients.
- **Deterministic icon pipeline**: The `generate-brand-assets.ts` script validates opacity, size, and corner colors, preventing drift.
- **IPC security**: The `isTrustedDesktopAuthCompleteSender` check limits consumption to the trusted page, reducing XSS risk.
- **Test coverage**: Meaningful increases in unit and integration tests (447 passed in web, 24 Swift, 11 UI).
- **Clear separation of concerns**: New `@jovie/auth-routing` package isolates PKCE logic from app code.

---

### Final notes

I recommend addressing items **1–6** in Critical before merging, especially the undefined `matchesPathPrefix` and the race condition in `loadAuthCompletion`. After those fixes, the PR is ready for alpha deployment as stated in the description.

*Full review requires the complete diff; please re‑request review after the next push with all file changes visible.*

itstimwhite (commented):

itstimwhite:
## Review Summary
**Verdict**: Changes Requested (minor)

The PR introduces a well-structured centralized auth routing system and a deterministic icon pipeline. The PKCE-based native auth flows for iOS/Electron are architecturally sound. However, there is a notable discrepancy in the desktop staging icon test that may indicate a regression, and a minor security hardening opportunity.

---

### Critical Issues

*No critical bugs or security vulnerabilities found.*

---

### Warnings

1. **apps/desktop/scripts/desktop-icon-contract.test.mjs:67–69**  
   The test now asserts `assert.deepEqual(stagingIcon, expectedProductionIcon)`, meaning the staging icon is expected to be identical to the production canonical icon. Previously the staging icon had an orange “S” badge (via `gen-staging-icon.mjs`). The new `gen-staging-icon.mjs` and `prepare-icons.mjs` simply call `generate-brand-assets.ts` without adding any badge.  
   **Why it matters**: If the staging badge was intentional for visual differentiation during internal builds, removing it could cause confusion. The PR description does not mention this change.  
   **Suggested fix**: Either confirm that the badge is no longer needed (update the commit message or add a comment) or restore the badge composition in the brand generator. If the badge is intentionally removed, the test is correct; otherwise it’s a regression.

---

### Suggestions

1. **apps/desktop/src/main.ts:270–274**  
   `PendingDesktopAuthPkce` includes a `createdAt` timestamp but it is never validated. Stale PKCE values could theoretically be consumed long after the auth flow started, increasing the window for a race condition if multiple deep-links arrive.  
   **Suggestion**: Add a time-to-live check (e.g., 10 minutes) in `consumePendingDesktopAuthPkce` or `handleAuthCompletion` to reject expired PKCE.

2. **apps/web/app/api/auth/native/exchange/route.ts** (not shown in snippet but referenced in PR)  
   Ensure the exchange endpoint enforces rate limiting and validates that the `code_verifier` matches the stored `code_challenge` using SHA-256. Also confirm that the endpoint cannot be called from browser context (only native clients via proper origin/header checks).

3. **apps/desktop/src/main.ts:156–158**  
   The `isTrustedDesktopAuthCompleteSender` function uses `getIpcSenderUrl(event)` to validate origin. Consider also verifying that the webContents does not have any loaded cross-origin iframe (e.g., `webContents.mainFrame`), though this is a minor hardening point.

4. **Convention**: The new `packages/auth-routing` package introduces its own `vitest.config.mts`. Ensure it inherits the monorepo’s shared TypeScript/Vitest config (if applicable) to avoid duplication.

---

### What Looks Good

- **Centralized auth routing**: The introduction of `/auth/start`, `/auth/callback`, and `/auth/native-complete` creates a clean, client-agnostic boundary. PKCE is correctly implemented with `code_challenge_method=S256` and secure code verifier generation.
- **iOS/Electron PKCE integration**: Both native platforms now use ASWebAuthenticationSession and the new completion handlers. The code in `main.ts` properly handles second-instance and `open-url` events.
- **Deterministic icon pipeline**: Centralizing all brand assets into `generate-brand-assets.ts` removes drifting icon sources. The desktop icon contract test now validates pixel-level consistency and black corners.
- **Type safety**: All new TypeScript code avoids `any` and uses proper interfaces (e.g., `DesktopAuthCompletion`, `PendingDesktopAuthPkce`).
- **Test coverage**: The PR includes unit tests for auth routing, PKCE, branded icons, and E2E overflow detection. The reported test pass counts are thorough.

The PR is close to ready; only the staging icon inconsistency needs clarification or correction before merging.

itstimwhite (commented):

itstimwhite:
## Review Summary

This PR introduces a much-needed centralized auth routing layer with PKCE for desktop and mobile, plus a deterministic icon pipeline. The overall architecture is sound, but I found several issues that must be addressed before shipping:

1. **Staging icon lost its badge** (regression in `desktop-icon-contract.test.mjs` and `gen-staging-icon.mjs`)
2. **PKCE state–verifier binding is missing** (security design flaw in `main.ts`)
3. **Missing PKCE expiry** (resource leak in `main.ts`)
4. **iOS auth handler is a no-op** (broken flow in `JovieApp.swift`)

---

### Critical Issues

#### 1. Staging icon identical to production icon
**`apps/desktop/scripts/desktop-icon-contract.test.mjs` – line 99**  
**`apps/desktop/scripts/gen-staging-icon.mjs` – removed badge logic**  

The new `gen-staging-icon.mjs` simply delegates to the brand generator without adding the orange “S” badge. The updated contract test now asserts `assert.deepEqual(stagingIcon, expectedProductionIcon)` (line 99), confirming both icons are identical. This breaks the long-standing convention of visually distinguishing staging builds from production.

**Why it matters**: Developers, QA, and automated screenshot tests rely

itstimwhite (commented):

itstimwhite:
## Review Summary

This PR introduces a significant architectural shift for auth routing and icon generation. The core ideas (PKCE-based native auth, centralized routing, deterministic icons) are sound and well-motivated. However, several issues in the implementation could cause regressions or incomplete functionality, particularly around iOS auth handling and desktop URL parameter preservation.

---

### Critical Issues

1. **apps/ios/Jovie/App/JovieApp.swift:54** — `onAuthReturn` closure is a no-op, breaking iOS PKCE completion  
   The `onAuthReturn` parameter is passed as `{ _ in }`, meaning the `MobileAuthReturn` (containing `code`, `state`, `codeVerifier`) is completely ignored. Without processing this return, the iOS app cannot complete the PKCE exchange. The actual exchange logic (calling the server endpoint, storing tokens) appears to be missing from this diff.  
   **Fix**: Implement the closure to call the native exchange API route (`/api/auth/native/exchange`) with the received code and verifier, then update the app state accordingly.

2. **apps/desktop/src/main.ts:307-316** — `buildDesktopBrowserAuthUrl` discards all original query parameters  
   The function now constructs a completely new URL via `buildCentralDesktopAuthUrl(intent, desktopReturn)`, ignoring any existing query parameters from the original `urlString` (e.g., `prompt`, `redirect_url`, `login_hint`). This could break flows that depend on these parameters (e.g., forced re-authentication, pre-filled email).  
   **Fix**: Preserve unknown query parameters from the original URL when building the central auth URL, or at least log a warning if they are dropped.

3. **Duplicate brand asset generation scripts** — `scripts/generate-brand-assets.ts` (root) and `apps/web/scripts/generate-brand-assets.ts`  
   Both files appear in the file list. The desktop scripts now call the root script, but the web-specific script remains. This creates two potential sources of truth and could lead to divergent icons if only one is updated.  
   **Fix**: Remove the web-specific script and ensure all consumers (web, desktop, iOS) use the single root script. Update any references in `package.json` or CI accordingly.

---

### Warnings

4. **apps/ios/Jovie/App/MobileAuthReturn.swift:10** — `codeVerifier` parameter is optional but required for non-nil return  
   The `parse` function accepts an optional `codeVerifier` but returns `nil` if it is `nil` or empty. The only call site (in `RootView.swift`) passes `onAuthReturn` which is a closure, not a direct call to `parse`. It is unclear where `parse` is actually invoked with the correct `codeVerifier`. If the verifier is not provided, the function will always fail, breaking the entire iOS auth flow.  
   **Fix**: Make `codeVerifier` a required parameter (remove the default `nil`) and ensure the caller generates and stores the verifier during the `ASWebAuthenticationSession` start.

5. **apps/desktop/src/main.ts:307-316** — Central auth URL may break existing desktop auth flows  
   The new `buildCentralDesktopAuthUrl` always uses `intent` derived from the pathname (`/signup` → `sign_up`, else `sign_in`). However, the original URL might have been for a different intent (e.g., password reset, email verification). The central auth start route (`/auth/start`) may not support all intents.  
   **Fix**: Either map all possible intents or fall back to the original URL if the intent is not recognized.

6. **apps/web/app/api/auth/native/exchange/route.ts** — Missing from diff, but critical for security  
   This API route exchanges the authorization code for tokens. It must validate the `code_verifier`, `state`, and ensure the code hasn't been reused. Without seeing the implementation, we cannot confirm it is secure.  
   **Action**: Verify that the exchange route uses Clerk's PKCE verification, enforces one-time use, and returns appropriate error codes.

---

### Suggestions

7. **Add tests for the new IPC channel**  
   The `consume-desktop-auth-completion` channel is not covered by the existing test files shown in the diff. Consider adding a unit test in `apps/web/lib/desktop/electron-bridge.test.ts` that mocks the IPC call and verifies the response shape.

8. **Consider a fallback for unknown query params in `buildDesktopBrowserAuthUrl`**  
   To avoid silently dropping parameters, you could merge the original URL's search params into the new central auth URL, excluding only those that are overridden (e.g., `client`, `intent`, `code_challenge`). This preserves backward compatibility.

9. **Remove the web-specific brand script**  
   If the root script is the single source, delete `apps/web/scripts/generate-brand-assets.ts` and update any references in `apps/web/package.json` to use the root script.

10. **Document the new auth flow**  
    The PR introduces a complex multi-platform auth flow (PKCE, deep links, IPC). Adding a brief architecture doc or updating the existing auth documentation would help future maintainers.

---

### What Looks Good

- **Centralized auth routing** with PKCE significantly improves security over the previous ticket-based approach.
- **Deterministic icon pipeline** eliminates drift between platforms and ensures brand consistency.
- **Legacy support** is maintained for both desktop (`jovie://auth-return`) and iOS (`auth-return` host) – good backward compatibility.
- **Extensive test coverage** (447 tests passing) gives confidence in the core logic.
- **Clean separation of concerns** – the new `@jovie/auth-routing` package encapsulates shared auth logic.
- **Use of standard crypto** (`node:crypto`) for PKCE generation rather than a third-party library.
- **Proper origin/pathname validation** in IPC handlers reduces attack surface.

itstimwhite (commented):

PR #9577, opened 2026-05-24
feat(web): stabilize chat timeline rendering and streaming state

itstimwhite:
## Root cause
- Multiple competing message sources (query hydration, local optimistic rows, streaming deltas, and cache mutations) could independently replace the timeline.
- Skeleton patterns and loading paths were not unified across route/chat surfaces.
- Stale refetch/stream events could override newer rows because timeline ownership was implicit.

## Lifecycle fix
- Added canonical `ChatTimelineState.messages` reducer and migration of UI render path to consume only reducer output.
- `useJovieChat` now dispatches merge events from local send/stream/refetch/data paths instead of directly replacing local state.
- Streaming updates and server acknowledgements now update stable timeline rows with `clientTurnId`-based IDs.
- Added production-oriented diagnostics and timeline merge/guard behavior.

## Validation
- `pnpm --filter @jovie/web run qa:public:exhaustive` (failed: `/dev/root-not-found-probe` returns console 404 in public-route QA)
- Focused reducer/unit/component checks and production smoke checks passed in prior run.

## Release Notes

* **New Features**
  * Added E2E chat regression testing with production configuration
  
* **Improvements**
  * Chat message rendering now uses real message states instead of synthetic placeholders
  * Send button behavior refactored for explicit callback control
  * Chat loading skeleton UI updated with new layout
  * Enhanced error handling with fallback mechanisms for edge cases
  * Message identity tracking improved with additional metadata fields

* **Tests**
  * Added comprehensive unit tests for chat message timeline reconciliation
  * New E2E regression suite for chat stream and refetch workflows

[Review Change Stack](https://app.coderabbit.ai/change-stack/JovieInc/Jovie/pull/9577?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/web/app/api/chat/conversations/[id]/route.ts
+++ b/apps/web/app/api/chat/conversations/[id]/route.ts
@@ -87,6 +87,8 @@ export async function GET(req: Request, { params }: RouteParams) {
         role: chatMessages.role,
         content: chatMessages.content,
         toolCalls: chatMessages.toolCalls,
+        clientMessageId: chatMessages.clientMessageId,
+        turnId: chatMessages.turnId,
         createdAt: chatMessages.createdAt,
       })
       .from(chatMessages)
--- a/apps/web/app/app/(shell)/chat/ChatPageClient.tsx
+++ b/apps/web/app/app/(shell)/chat/ChatPageClient.tsx
@@ -344,7 +344,24 @@ export function ChatPageClient({
   const hasProfilesButNoSelection =
     creatorProfiles.length > 0 && !selectedProfile && !needsOnboarding;
   const fallbackProfile = hasProfilesButNoSelection ? creatorProfiles[0] : null;
-  const activeProfile = selectedProfile ?? fallbackProfile;
+  const e2eFallbackProfile =
+    env.IS_E2E && !selectedProfile && !fallbackProfile
+      ? ({
+          id: '00000000-0000-4000-8000-000000000102',
+          userId: '00000000-0000-4000-8000-000000000101',
+          username: 'e2e-chat-smoke',
+          displayName: 'E2E Chat Smoke',
+          avatarUrl: null,
+          spotifyUrl: 'https://open.spotify.com/artist/e2e',
+          spotifyId: 'e2e',
+          appleMusicUrl: null,
+          appleMusicId: null,
+          youtubeUrl: null,
+          youtubeMusicId: null,
+        } as ChatActionProfile)
+      : null;
+  const activeProfile =
+    selectedProfile ?? fallbackProfile ?? e2eFallbackProfile;
   const hasDashboardLoadFailure = Boolean(dashboardLoadError);
   const isProfileSetupRace =
     hasProfilesButNoSelection && !hasDashboardLoadFailure;
@@ -527,7 +544,15 @@ export function ChatPageClient({
         );
       }
 
-      if (phase === 'completed') {
+      // New chats reserve the final URL with history.replaceState as soon as
+      // the server acknowledges the turn. A second router.replace on stream
+      // completion remounts the chat surface and can let stale refetch data
+      // replace the settled local timeline.
+      if (
+        phase === 'completed' &&
+        currentPath !== APP_ROUTES.CHAT &&
+        currentPath !== nextRoute
+      ) {
         router.replace(nextRoute, {
           scroll: false,
         });
--- a/apps/web/app/app/(shell)/chat/loading.tsx
+++ b/apps/web/app/app/(shell)/chat/loading.tsx
@@ -1,10 +1,11 @@
 import { Skeleton } from '@jovie/ui';
+import { JovieMarkElectric } from '@/components/atoms/JovieMarkElectric';
 import { ChatWorkspaceSurface } from '@/components/jovie/ChatWorkspaceSurface';
 import { LoadingSkeleton } from '@/components/molecules/LoadingSkeleton';
 
 /**
  * Chat page loading skeleton.
- * Matches the JovieChat empty state layout: centered heading, action card, and input.
+ * Matches the JovieChat empty state layout: centered logo space and composer.
  */
 export default function ChatLoading() {
   return (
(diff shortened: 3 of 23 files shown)


itstimwhite:
## Review Summary

The PR introduces a much‑needed reducer‑based timeline to eliminate race conditions in chat state management. The architectural direction is sound, and the removal of the synthetic thinking placeholder simplifies the component. However, a **security‑sensitive logic error** in the E2E dashboard fallback could allow a bypass in non‑preview environments, and several minor inconsistencies should be addressed before merging.

---

### Critical Issues

#### 🚨 `apps/web/app/app/(shell)/dashboard/actions/dashboard-data.ts:456`

The condition `VERCEL_ENV !== 'preview'` is inverted for the intended use case. The E2E fallback is allowed in **production** and **development** but disallowed in **preview** deployments, which is contrary to the typical safety model (preview = test, production = real).

```ts
// current code – allows production, blocks preview
VERCEL_ENV !== 'preview'

// likely intended – allow only in development or preview
VERCEL_ENV === 'development' || VERCEL_ENV === 'preview'
```

Although the feature is gated by an environment variable `E2E_USE_TEST_AUTH_BYPASS`, a misconfigured production deployment would be vulnerable to a trivial onboarding bypass. This must be tightened to non‑production only.

#### 🚨 `apps/web/components/jovie/hooks/useJovieChat.ts:235`

`dispatchTimelineEvent` is created with an empty dependency array (`useCallback(() => { … }, [])`) but uses `dispatchTimeline` from `useReducer`. While `dispatchTimeline` is stable across renders, this pattern violates React’s rules of hooks and may lead to stale closure warnings from linters. More importantly, any future refactoring that adds a dependency could silently break.

**Fix**: Include `dispatchTimeline` in the dependency array:

```ts
const dispatchTimelineEvent = useCallback((event: ChatTimelineEvent) => {
  // …
  dispatchTimeline(event);
}, [dispatchTimeline]);
```

---

### Warnings

#### ⚠️ `apps/web/components/jovie/hooks/useJovieChat.ts:379`

The mapping of `existingConversation.messages` adds a `requestId` field, but the API query (`route.ts`) does **not** fetch `requestId`. This results in `requestId: null` for all server messages, which may break future logic that relies on this field for deduplication.

**Suggestion**: Either fetch `requestId` from the database (if it exists) or remove the field from the timeline message type until needed.

#### ⚠️ `apps/web/components/jovie/hooks/useJovieChat.ts:68`

A module‑level `Map<string, ChatTimelineState>` is used as a simple cache. This works but can hold references to conversation state after the component unmounts (up to the limit of 20). In development with React Strict Mode, this could cause the component to read stale state during re‑mounts. Consider clearing the cache entry when the hook unmounts or tying it to the component’s lifecycle.

```ts
useEffect(() => {
  return () => {
    if (activeConversationId) {
      timelineStateCache.delete(activeConversationId);
    }
  };
}, [activeConversationId]);
```

---

### Suggestions

#### 💡 `apps/web/components/jovie/JovieChat.tsx:447`

The `isThinking` helper now checks `message.role === 'assistant' && message.status === 'pending'`. This is cleaner, but ensure the reducer never sets `status: 'pending'` on user messages. (Seems safe given the reducer design, but a unit test for this invariant would be valuable.)

#### 💡 `apps/web/tests/e2e/chat-timeline-regression.spec.ts`

The new E2E test is excellent. Consider adding a case that verifies the router does **not** replace when already on the `/chat` page (the race condition fix). Also test that a second stream arriving after a completed stream does not overwrite the timeline.

#### 💡 `apps/web/components/jovie/hooks/useJovieChat.ts:281`

The `adoptServerConversationId` callback now only sets `activeConversationId` when `phase === 'completed'`. The comment explains the reasoning well, but there is a subtlety: if the conversation is brand new and the server responds with an ID mid‑stream, the hook still holds the old `activeConversationId = null`. This could cause issues if other logic relies on `activeConversationId` being set before completion. Confirm that the `customTransport` dispatching `message.send.acknowledged` with the server ID is sufficient to allow the reducer to associate messages with the correct conversation.

---

### What Looks Good

- ✅ **Removal of synthetic placeholder** – `THINKING_PLACEHOLDER_ID` eliminated. The “thinking” state is now driven by the actual message status. This reduces a class of rendering bugs.
- ✅ **Reducer architecture** – Centralising timeline mutations in `reduceChatTimeline` makes state transitions predictable and testable.
- ✅ **Production diagnostics** – The `jovie:chat-timeline-debug` local storage flag and analytics logging will greatly simplify debugging stream issues without affecting production users.
- ✅ **Router‑replace guard** – The condition `currentPath !== APP_ROUTES.CHAT && currentPath !== nextRoute` prevents unnecessary remounts when a stream completes on an existing chat page.
- ✅ **E2E test infrastructure** – The fallback profile and test auth bypass are neatly isolated behind environment variables, with comprehensive test coverage for the reducer.
- ✅ **Accessibility of skeleton** – The `aria-hidden="true"` on the decorative logo in the loading state is correct.

---

### Summary of Required Changes

1. **Fix the E2E fallback environment check** to block production (or use `!isProduction()` helper).
2. **Add missing dependency** to `dispatchTimelineEvent` `useCallback`.
3. **(Optional but recommended)** Align `requestId` field in timeline messages with what the API actually returns.

Once these items are addressed, the PR is good to merge.

itstimwhite (commented):

itstimwhite:
## Review Summary

The PR introduces a well-structured chat timeline reducer and addresses stale state issues. However, a syntax error and a few convention/style concerns need to be resolved before merge.

---

### Critical Issues

- **`apps/web/components/jovie/hooks/useJovieChat.ts` (line ~443, after `setInput(...);`)** – Stray character `i` on its own line.  
  The diff shows a bare `+      i` after the closing brace of a block. This is not valid TypeScript and will cause a compile error.  
  **Fix**: Remove the stray `i` (or if it was meant to be part of an expression, complete it properly).

- **`apps/web/components/jovie/hooks/useJovieChat.ts` – `handleChatFailure` callback references `clientTurnId` but never uses it** (the diff cuts off, but the parameter appears unused).  
  If the parameter is not consumed, the change is dead code. More importantly, the `activeClientTurnIdRef.current` default value might be stale. Verify that `clientTurnId` is actually passed to an error reporting or dispatch action inside `handleChatFailure`. If not, remove the parameter.

---

### Warnings

- **`apps/web/lib/admin/roles.ts` and `apps/web/lib/auth/dev-test-auth.server.ts`** are changed in the diff but the changes are not shown. These files are security‑sensitive; ensure they have been reviewed for any hardcoded credentials or unsafe bypass logic before merging.

- **`apps/web/lib/auth/gate.ts`** similarly appears in the file list but not in the diff. If it was modified, check that the auth gate still enforces proper checks, especially given the new E2E fallback paths.

- **`apps/web/components/jovie/hooks/useJovieChat.ts` – Custom `dispatchTimelineEvent` reads `timelineStateRef.current` and then calls `reduceChatTimeline` manually before dispatching.**  
  Because React batches dispatches, two synchronous dispatches will both see the same old ref value for logging, leading to incorrect “previousState” in diagnostics. While not a functional bug, it makes the diagnostic unreliable during concurrent events.  
  **Suggestion**: Use `useEffect` that logs the state after commit, or compute the previous state inside the reducer itself.

- **`apps/web/app/app/(shell)/dashboard/actions/dashboard-data.ts` – The E2E fallback functions (`canUseE2EDashboardFallback`, `createE2EDashboardCoreData`) are included in the production bundle.**  
  Although guarded by runtime environment checks, the presence of hardcoded user IDs and test profiles in production code is a supply‑chain risk. Consider moving these helpers to a separate test‑only file that is not imported in production paths.

- **`apps/web/components/jovie/hooks/useJovieChat.ts` – Module‑level `Map` cache (`timelineStateCache`) persists across component instances.**  
  In Next.js, if this module is ever imported on the server (e.g., during SSR), the cache would be shared across requests. Even though it’s only used in a client hook, a future refactor could cause issues. Add a comment or guard.

- **`apps/web/components/jovie/hooks/useJovieChat.ts` – `getLastAssistantMessage` and `getPartsSignature` are defined but never used in the visible diff.**  
  If they are used elsewhere in the same file (outside the diff), that’s fine. Otherwise, remove dead code.

---

### Suggestions

- **`apps/web/components/jovie/hooks/useJovieChat.ts` – The `dispatchTimelineEvent` callback uses `logger.info` and `track` for every event in development and for stale events in production.**  
  This could generate a lot of analytics events. Consider throttling or only emitting them when the `CHAT_JANK_MONITOR` flag is enabled.

- **`apps/web/components/jovie/components/ChatComposerToolbar.tsx` – The `onSend` prop is added, but the button is now always `type='button'` and invokes `onSend` on click.**  
  This removes the implicit form submission behaviour. Ensure that the parent component no longer expects the form to submit on Enter (the `ChatInput` handles Enter via `handleKeyDown`). If the form submit is still needed for accessibility, keep the fallback.

- **`apps/web/tests/unit/chat/chat-timeline-reducer.test.ts`** – The only unit test file visible is for the reducer. Write tests for the `dispatchTimelineEvent` wrapper and the caching logic to prevent regressions.

---

### What Looks Good

- ✅ The reducer pattern (`ChatTimelineState.messages`) centralizes all message state mutations, eliminating the previous race between local sends, streaming, and cache refetches.
- ✅ Migration from synthetic placeholders (`THINKING_PLACEHOLDER_ID`) to real message statuses (`message.status === 'pending'`) simplifies the rendering logic.
- ✅ The E2E test infrastructure with hardcoded test data is isolated behind strict environment guards, reducing risk of accidental activation.
- ✅ `clientTurnId`‑based ID matching ensures streaming updates land on the correct row.
- ✅ Production diagnostics (stale‑event logging) and caching with size limits are well‑reasoned additions.
- ✅ Comprehensive set of test files (unit + E2E + snapshot) demonstrates thorough validation.

---

Fix the stray `i` and confirm the `handleChatFailure` parameter usage, and this PR will be in good shape.

itstimwhite (commented):

PR #9588, opened 2026-05-24
fix(electron): restore native auth handoff

itstimwhite:
## Summary

Fixes the native Electron auth handoff for the installed Mac app.

- hides the main Electron app shell while the auth handoff window is active
- makes browser auth launches awaitable and renderer-visible when they fail
- registers the production `jovie://` auth-return protocol in the macOS bundle metadata
- keeps `/desktop-auth` retryable with a stable status/error slot
- suppresses the visible cookie banner on auth/native utility routes
- fixes native `/auth/callback` state consumption so validated Electron callbacks can create the native exchange code and return through the deep link
- bumps the desktop release trigger to `26.5.42`

## Verification

- `pnpm --filter @jovie/desktop run test`
- `pnpm --filter @jovie/desktop run typecheck`
- `pnpm --filter @jovie/web exec vitest run tests/unit/app/desktop-auth-page.test.tsx lib/desktop/electron-bridge.test.ts tests/unit/cookie-banner.test.tsx tests/unit/cookie-banner-fixes.test.tsx tests/unit/lib/auth/routing-state.server.test.ts app/auth/callback/route.test.ts`
- `pnpm --filter @jovie/web run typecheck -- --pretty false`
- `pnpm biome check --no-errors-on-unmatched ...focused files...`
- `node scripts/desktop-release-guard.mjs --base origin/main`
- `pnpm --filter @jovie/desktop run package:production`
- `codesign --verify --deep --strict --verbose=2 apps/desktop/dist/mac-universal/Jovie.app`
- installed `/Applications/Jovie.app` from the rebuilt `26.5.42` bundle and verified `CFBundleURLTypes` includes `jovie`
- push hook: full Biome, proxy guard, Tailwind guard, web typecheck, web lint

## QA Evidence

Screenshots: `/tmp/jovie-electron-qa-20260523-225946`

- final handoff shell: `05-final-auth-handoff-main-hidden.png`
- installed app window list after rebuild: one Jovie window at `420x360`
- installed bundle version: `26.5.42`
- installed app backup before replacement: `/Users/timwhite/Desktop/Jovie.app.backup-20260524-000616-pre-26542`

PR #9597, opened 2026-05-24
fix(auth): harden desktop callback handoff

itstimwhite:
## Summary

Follow-up to #9588 after review found a few hardening issues in the desktop auth handoff.

- consumes `/auth/callback` state once before creating the native exchange code, so callback replay fails closed instead of leaving auth state alive after exchange creation
- keeps the desktop auth button disabled only while a browser-open attempt is pending, then makes it retryable again after success/failure/timeout
- clears the browser-open timeout when the bridge resolves quickly and exposes the status row with `role="status"`
- propagates explicit desktop bridge failure reasons instead of masking them with a browser fallback
- removes the small `showWindow` recursion edge and names the auth handoff window bounds
- bumps the desktop release trigger to `26.5.43`

## Verification

- `node scripts/desktop-release-guard.mjs --base origin/main`
- `pnpm --filter @jovie/web exec vitest run tests/unit/app/desktop-auth-page.test.tsx lib/desktop/electron-bridge.test.ts app/auth/callback/route.test.ts tests/unit/lib/auth/routing-state.server.test.ts`
- `pnpm --filter @jovie/web exec vitest run tests/unit/cookie-banner.test.tsx tests/unit/cookie-banner-fixes.test.tsx`
- `pnpm --filter @jovie/desktop run test`
- `pnpm --filter @jovie/desktop run typecheck`
- `pnpm biome check --no-errors-on-unmatched ...touched files...`
- `pnpm --filter @jovie/desktop run package:production`
- `codesign --verify --deep --strict --verbose=2 apps/desktop/dist/mac-universal/Jovie.app`
- installed `/Applications/Jovie.app` from the rebuilt `26.5.43` bundle and verified `CFBundleURLTypes` includes `jovie`
- push hook: full Biome, proxy guard, Tailwind guard, web typecheck, web lint

## Installed App Evidence

- final installed backup before replacement: `/Users/timwhite/Desktop/Jovie.app.backup-20260524-004114-pre-26543`
- installed app version: `26.5.43`
- installed app protocol: `jovie://`


itstimwhite:
## Review Summary
**Verdict**: Approve

This PR makes meaningful security and UX improvements to the desktop auth handoff flow. The atomic consumption of auth state prevents replay attacks, the button disable logic is more precise, timeout resources are properly cleaned, and bridge failure reasons are surfaced instead of being masked. The refactoring in the Electron main process eliminates a recursion edge case. Tests are updated and a new test validates the button re-enabling behavior.

---

### Critical Issues

None.

---

### Warnings

- **apps/web/app/desktop-auth/page.tsx:67** – Race condition in `openAuthUrl`  
  The guard `if (!authUrl || openState === 'opening') return;` uses the closure’s `openState` value, not the latest state. If a user double-clicks the button very quickly, both click handlers may see `openState` as `'start'` before React processes the first `setOpenState('opening')`. That could trigger two concurrent `openWithTimeout` calls.  
  **Why it matters**: Two browser windows could be opened, or the timeout/race logic could produce unexpected results.  
  **Suggested fix**: Use a `useRef` to track whether an open is pending:

  ```ts
  const pendingOpenRef = useRef(false);
  const openAuthUrl = useCallback(async () => {
    if (!authUrl || pendingOpenRef.current) return;
    pendingOpenRef.current = true;
    setOpenState('opening');
    // ... openWithTimeout ...
    pendingOpenRef.current = false;
  }, [authUrl]);
  ```

- **apps/web/lib/desktop/electron-bridge.test.ts** – Missing test for the `undefined` fallback case  
  The PR modifies the behavior so that when IPC returns a result (even `{ ok: false }`), it no longer falls back to the browser. But the existing test only covers the case where IPC returns `{ ok: false, reason }`. There is no test that verifies the old fallback still works when `api.startDesktopAuthHandoff` returns `undefined`.  
  **Why it matters**: The fallback path is still present and used when the API method doesn’t exist. A regression test protects it.  
  **Suggested fix**: Add a test case:

  ```ts
  it('falls back to browser window open when IPC returns undefined', async () => { ... });
  ```

---

### Suggestions

- **apps/web/app/desktop-auth/page.tsx** – Consider adding error handling for unexpected rejections in `openWithTimeout`  
  The function doesn’t catch rejected promises from `openDesktopAuthUrl`. If an unexpected error occurs, it will result in an unhandled promise rejection that is not surfaced to the user.  
  **Suggestion**: Wrap the entire function body in a try/catch that returns `{ ok: false, reason: 'open-external-failed' }`.

- **apps/web/app/auth/callback/route.ts** – The `410 Gone` response could include a more descriptive `Retry-After` header or a link to retry the flow. Not required, but improves developer /edge-case UX.

- **apps/desktop/src/main.ts** – The extracted `AUTH_HANDOFF_WINDOW_BOUNDS` constant is a nice improvement. Consider moving it closer to where `BrowserWindow` is constructed to improve readability.

---

### What Looks Good

- **Atomic auth state consumption** (`consumeStoredAuthState`) replaces separate read+delete, preventing replay attacks with a single atomic Redis operation.
- **Button disable/enable logic** is now correct: disabled only while a browser launch attempt is pending, and re-enabled after success/failure/timeout.
- **Timeout cleanup** in `openWithTimeout` uses `finally` to clear the timer, eliminating a resource leak when the bridge resolves quickly.
- **Explicit failure reasons** are propagated from the Electron bridge instead of being replaced with a generic browser fallback. The updated test validates this.
- **Recursion bug eliminated** in `showWindow` by extracting `showWindowNow`, preventing stack overflows when the auth handoff window is shown.
- **Accessibility improvement** – Added `role="status"` to the status row.
- **Test coverage** – New integration test for button re-enabling and the already-consumed state edge case.
- **Cleanup** – Removed dead `deleteStoredAuthState` export and simplified the callback route.

itstimwhite (commented):

PR #9638, opened 2026-05-25
fix(ios): complete native auth callback flow

itstimwhite:
## Summary
- complete iOS native auth callback finalization after ASWebAuthenticationSession returns through ie.jov.jovie://auth/complete
- persist native session-token fallback and authenticate mobile bearer session tokens on /api/mobile/v1/me and chat turns
- add DEBUG/TestFlight-safe auth diagnostics plus deterministic and HTTPS simulator auth coverage

## Redirect strategy
- keep native return on custom scheme: ie.jov.jovie://auth/complete
- real-browser simulator proof uses HTTPS WEB_BASE_URL/API_BASE_URL and returns through the custom scheme
- universal links stay out of this change until AASA/applinks are valid

## Verification
- WEB_BASE_URL=https://survey-landing-dot-essex.trycloudflare.com API_BASE_URL=https://survey-landing-dot-essex.trycloudflare.com JOVIE_IOS_REAL_BROWSER_AUTH=1 pnpm test:auth:ios
- pnpm test:auth:ios
- pnpm --filter @jovie/web exec vitest run tests/unit/api/dev/test-auth-routes.test.ts
- pnpm --filter @jovie/web exec vitest run tests/unit/api/mobile/v1/me.test.ts tests/unit/api/mobile/v1/chat-turns.test.ts
- pnpm --filter @jovie/web run typecheck -- --pretty false
- bash -n apps/ios/scripts/test-auth-callback.sh apps/ios/scripts/setup-env.sh apps/ios/scripts/ensure-configuration.sh apps/ios/scripts/write-configuration.sh && git diff --check
- pre-push hook: biome check ., next proxy guard, tailwind guard, typecheck, biome lint

## Notes
- Cloudflare tunnel was an ephemeral HTTPS simulator target for validation only.
- No provider dashboard change is required for the test harness path; production/provider redirect must allow ie.jov.jovie://auth/complete.


itstimwhite:
## Review Summary

This PR implements the complete iOS native auth callback flow with persistent session handling and diagnostics. The approach is generally sound, but I identified several issues that must be addressed before merging.

---

### Critical Issues

1. **`apps/ios/Jovie/App/MobileAuthReturn.swift` – Insecure storage of code verifier**  
   *Line: `MobileAuthPendingStore.save(codeVerifier:)` saves the OAuth code verifier in `UserDefaults.standard`.*  
   The code verifier is a critical OAuth secret that, combined with the `code` returned in the callback URL, allows exchanging for a session token. Storing it in plaintext in UserDefaults exposes it to any process or jailbreak access. The Jovie iOS codebase already uses Keychain for Clerk device tokens. The verifier should be stored in Keychain with the same `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`.  
   **Fix**: Replace UserDefaults with Keychain storage using `SecItemAdd`/`SecItemCopyMatching` for the code verifier.

2. **`apps/web/app/api/auth/native/exchange/route.ts` – Missing redirect URI validation**  
   While the full implementation is not visible in this diff, the exchange endpoint must validate that the `redirect_uri` parameter matches the registered custom scheme (`ie.jov.jovie://auth/complete`). Without this, an attacker could use a stolen `code` and `code_verifier` to obtain a session by passing a different redirect URI.  
   **Fix**: Ensure the route explicitly checks the `redirect_uri` against the allowed scheme and rejects mismatched values.

3. **`apps/web/app/api/mobile/v1/me/route.ts` and `chat/turns/route.ts` – Session token validation must use Clerk SDK**  
   These routes now accept bearer session tokens. If the token validation relies solely on local signature verification without checking against Clerk’s API or a secure environment variable for the Clerk JWKS URL, tokens could be forged or reused after revocation.  
   **Fix**: Use Clerk’s server-side session verification (`clerkClient.sessions.verifySession()`) or fetch the JWKS from the configured Clerk backend. Do not implement custom JWT validation unless specifically required and tested.

4. **`apps/ios/Jovie/App/JovieApp.swift` – `NativeTicketSignInResponseMiddleware.validate()` may throw unhandled error during Clerk sign-in**  
   *Lines: `try await NativeClerkDeviceTokenStore.persistDeviceToken(from: response)`* (inside `validate`). If persisting the device token fails (e.g., Keychain write error), the entire Clerk sign-in response handling will fail and propagate as an unhandled error. This could prevent successful sign-in even when the token is not essential.  
   **Fix**: Wrap the persistence call in a `do-catch` and log the error without throwing.

---

### Warnings

1. **`apps/ios/Jovie/App/MobileAuthReturn.swift` – `MobileAuthCallbackURLInbox` may drop URLs if `drain()` is called multiple times**  
   The `drain()` method removes all pending URLs. If it is called redundantly (e.g., from multiple observers), pending URLs will be lost. While the design likely ensures a single drain, consider using a guard to prevent accidental double-drain or switch to a `consume()` pattern that returns one URL at a time.

2. **`apps/ios/Jovie/App/RootView.swift` – `waitForClerkLoaded` polls 60 times with 50ms delay**  
   This is a 3-second spin loop. If Clerk initialization is slower (e.g., network issues), the UI will be blocked (though async, the caller is main actor). Consider adding a timeout with a user-visible error or using a KVO/notification approach for when Clerk finishes loading.

3. **`apps/ios/Jovie/App/Configuration.swift` – Environment variables now take priority over Info.plist values**  
   *Line order change: env vars checked before plist.* This is intentional but could break deployments if environment variables are accidentally set to empty strings (trimmed value emptiness check remains). Ensure CI/CD does not inadvertently set these variables to empty.

4. **`apps/ios/Jovie/App/LaunchMode.swift` – New launch modes `uiTestingRealBrowserAuth` and `uiTestingAuthCallback` are not added to the `logAnalytics` property**  
   The `logAnalytics` computed property returns `true` only for `.live`. The new testing modes will not log analytics (probably intentional), but ensure this is documented and that no side effects depend on analytics being sent.

5. **`apps/web/lib/mobile/session-auth.ts` – Hardcoded secret or environment variable usage**  
   Verify that any secret used for token signing/verification is not hardcoded in the source file and is only read from environment variables prefixed with `CLERK_` or `JOVIE_`.

---

### Suggestions

1. **Consider adding unit tests for `/api/auth/native/exchange/route.ts`**  
   The PR adds tests for `/me` and `/chat-turns` but does not include tests for the exchange endpoint. Even a simple mock test verifying correct token validation on success/failure would improve coverage.

2. **Use `SecItemCopyMatching` instead of a separate `baseQuery()` for update**  
   In `NativeClerkDeviceTokenStore`, the `updateStatus` call re-creates the base query. It’s more efficient to store the base query as a static or computed property.

3. **Add `#if DEBUG` guards to the `MobileAuthPendingStore`**  
   During production, having the code verifier in UserDefaults is a risk. While you'll likely switch to Keychain, you could also fall back to UserDefaults in debug builds only.

4. **Document the iOS Keychain accessibility settings**  
   The current use of `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` may not survive device reboots if the user never unlocks the device. Consider whether `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` is more appropriate for tokens needed immediately after app launch.

---

### What Looks Good

- Clear separation of concerns with `NativeClerkDeviceTokenFallback`, `MobileAuthPendingStore`, and `MobileAuthCallbackURLInbox`.
- Thorough use of diagnostics (`MobileAuthDiagnostics`, `NativeTicketSignInDiagnostics`) to aid debugging without affecting production.
- Proper handling of the custom scheme URL via both `onOpenURL` and `UIApplicationDelegate`.
- The decision to store device tokens in the Keychain rather than UserDefaults for security.
- The addition of `onAuthError` callback for better error handling in the UI.
- The PR respects Jovie web conventions: TypeScript, single db import, no emoji, etc.
- Conventional commit message format is followed.
- All new code is covered by existing or new test suites (me.test.ts, chat-turns.test.ts, test-auth-routes.test.ts, AppStateTests).

itstimwhite (commented):

itstimwhite:
## Review Summary  

The PR completes the iOS native auth callback flow with a thorough implementation across both the Swift and TypeScript layers. The architecture is sound, the diagnostics are useful for testing, and the session token persistence follows best practices. However, a few items need attention before merging.

---

### Critical Issues  

- **apps/ios/Jovie/App/MobileAuthReturn.swift:103-126** — Missing state validation in `parseCallbackComponents`.  
  The method extracts `state` from the callback URL but does not verify it against a previously stored value. Without this check, an attacker who can intercept or spoof the custom‑scheme callback could inject a forged `code` + `state` pair, bypassing PKCE protections.  
  **Fix**: Store `state` alongside `codeVerifier` in `MobileAuthPendingStore`, then compare it during parse. Only proceed if they match.

- **apps/ios/Jovie/App/JovieApp.swift:55-62** — `NativeTicketSignInRequestMiddleware` clears `Authorization` and `x-clerk-client-id` on POST `/v1/client/sign_ins` with `strategy=ticket`.  
  This discards any previously stored device token for the ticket sign‑in request. While the intention may be to start fresh, it breaks Clerk’s device token chain for this critical request, potentially causing a session to be created but not associated with the existing device.  
  **Fix**: Instead of clearing both headers, only clear the `Authorization` header if the device token is from a different device (e.g., by comparing client IDs). Better yet, let Clerk’s SDK manage these headers automatically and remove this custom middleware unless there is a proven bug.

- **apps/web/lib/mobile/session-auth.ts** (not fully visible in diff) — The new bearer‑session‑token authentication for `/api/mobile/v1/me` and `/chat/turns` must validate the token’s cryptographic signature and expiration.  
  **Action required**: Ensure the token is verified using Clerk’s public key or a symmetric secret, and that it includes the standard claims (`iss`, `sub`, `exp`, `jti`). A missing or weak validation could allow token forgery.

---

### Warnings  

- **apps/ios/Jovie/App/RootView.swift:317-341** — `runMobileAuthFinalizationStage` wraps errors but the `operation` closure is `async throws`. The catch block creates a `MobileAuthFinalizationStageError` that loses the original error’s localized description when it is empty.  
  Use `String(describing: underlyingError)` as fallback (already done), but consider logging the original error separately for debugging.

- **apps/ios/Jovie/App/MobileAuthReturn.swift:22-37** — `MobileAuthDiagnostics.record` calls `UserDefaults.synchronize()`. Synchronizing after every diagnostic write is unnecessary and can degrade I/O performance during rapid auth steps.  
  Remove `synchronize()`; `UserDefaults` flushes automatically at appropriate times.

- **Missing iOS Unit Tests** — The PR adds `NativeSessionTokenStore.swift`, `NativeClerkDeviceTokenStore`, `NativeTicketSignInRequestMiddleware`, and `MobileAuthPendingStore`, but no unit tests for these components are included.  
  Add at least basic tests covering:  
  - Keychain persistence (mock or use temporary keychain)  
  - Request/response middleware behaviour  
  - `MobileAuthPendingStore` concurrent access (consume vs save)  
  - `MobileAuthReturnParser` with invalid/empty/missing query items  

- **apps/ios/Jovie/App/JovieApp.swift:356-372** — The `waitForClerkLoaded` function uses a 50ms polling loop with up to 60 iterations (3 seconds total). If Clerk takes longer, it falls back to `refreshEnvironment` and `refreshClient`, which may not be intended for production use.  
  This function appears only in the live auth flow; ensure it has a proper timeout and cancels if the app is backgrounded.

---

### Suggestions  

- **apps/ios/Jovie/App/MobileAuthReturn.swift:81-99** — The `parseCallbackComponents` function assumes `href` query item? Wait, it checks `url.host == "auth"` and `url.path == "/complete"`. This is correct for the custom scheme but consider supporting future paths by making the path check configurable.

- **apps/ios/Jovie/App/Configuration.swift:54-60** — The `value(forKey:envKeys:)` method now prefers environment variables over plist values, which is intentional for the simulator. For release builds, consider adding a guard that only uses environment variables when `DEBUG` or `TESTFLIGHT` is active, to prevent accidental override from process environment.

- **apps/web/app/api/auth/native/exchange/route.ts** (diff not fully shown) — Ensure the exchange endpoint validates the `code_verifier` against the `code_challenge` stored during the initial auth request (PKCE). The iOS side sends the verifier; the server must match it.

- **apps/ios/Jovie/App/RootView.swift:66-72** — The `MobileAuthReturnError.missingSessionToken` case includes diagnostic info in the error description. Consider moving this to a separate logging channel so that the user-facing error is always generic (“Unable to complete sign‑in”).

---

### What Looks Good  

- **Keychain‑backed device token persistence** (`NativeClerkDeviceTokenStore`) — uses `SecItemAdd`/`SecItemUpdate` with `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`.  
- **Clear separation of concerns** between parsing, diagnostics, and token storage in the iOS layer.  
- **Use of Clerk’s response middleware** to capture device tokens after sign‑in/client refresh without modifying the SDK’s internal flow.  
- **Comprehensive test harness** (`pnpm test:auth:ios`) that covers callback parsing, custom‑scheme simulation, and real‑browser HTTPS flows.  
- **Diagnostics under `#if DEBUG`** that are stripped from release builds, keeping the production binary clean.  
- **Sign‑out path clears both Clerk session and native session token** (`NativeSessionTokenStore.clear()`) to prevent stale tokens.  

Address the state‑validation gap and the missing iOS unit tests, and verify the web session‑token validation before approving.

itstimwhite (commented):

PR #10221, opened 2026-06-06
fix(demo): neutralize onboarding step controls

itstimwhite:
## Summary
- Neutralize the demo onboarding step switcher selected state by moving it from accent-filled styling to the neutral primary button tokens.
- Add a source guard so blue/purple/accent-filled central action styling cannot return to this demo control.

## Linear
- JOV-2728

## Verification
- `JOVIE_AGENT_PROFILE=coder ./scripts/setup.sh`
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/web exec vitest run tests/unit/demo/onboarding-demo-system-b-style-guard.test.ts tests/unit/app/surface-elevation-guardrails.test.ts`
- `JOVIE_AGENT_PROFILE=coder pnpm biome check --write apps/web/components/features/demo/OnboardingDemoContent.tsx apps/web/tests/unit/demo/onboarding-demo-system-b-style-guard.test.ts`
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/web run typecheck -- --pretty false`
- `git diff --check`
- pre-commit hook: proxy guard, Biome, ESLint, staged a11y, typecheck
- pre-push hook: repo Biome, proxy guard, Tailwind guard, typecheck, Biome lint

## Integration Branch Note
This PR targets `codex/system-b-integration` so System B slices can batch before the next full main deploy.

### DIFF
--- a/apps/web/components/features/demo/OnboardingDemoContent.tsx
+++ b/apps/web/components/features/demo/OnboardingDemoContent.tsx
@@ -59,10 +59,10 @@ function StepSwitcher({
           type='button'
           onClick={() => onStepChange(step)}
           className={cn(
-            'shrink-0 rounded-full px-3 py-1.5 text-xs font-medium transition-colors',
+            'min-h-7 shrink-0 rounded-full border px-3 py-1.5 text-xs font-medium transition-colors duration-subtle',
             currentStep === step
-              ? 'bg-accent-token text-white'
-              : 'text-secondary-token hover:bg-surface-1 hover:text-primary-token'
+              ? 'border-(--linear-btn-primary-border) bg-btn-primary text-btn-primary-foreground shadow-button-inset hover:border-(--linear-btn-primary-hover) hover:bg-btn-primary-hover'
+              : 'border-transparent text-secondary-token hover:border-subtle hover:bg-surface-1 hover:text-primary-token'
           )}
         >
           {STEP_LABELS[step]}
--- a/apps/web/tests/unit/demo/onboarding-demo-system-b-style-guard.test.ts
+++ b/apps/web/tests/unit/demo/onboarding-demo-system-b-style-guard.test.ts
@@ -1,0 +1,46 @@
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+import { describe, expect, it } from 'vitest';
+
+const appRoot = resolve(__dirname, '../../..');
+const demoOnboardingSourcePath =
+  'components/features/demo/OnboardingDemoContent.tsx';
+
+const forbiddenCentralActionPatterns = [
+  /bg-accent-token\s+text-white/,
+  /\bbg-(?:blue|purple|violet|indigo)-\d/,
+  /\b(?:from|via|to)-(?:blue|purple|violet|indigo)-\d/,
+  /\btext-white\b/,
+  /--linear-accent/,
+] as const;
+
+describe('Onboarding demo System B source contract', () => {
+  it('keeps demo step switcher actions neutral instead of accent-filled', () => {
+    const source = readFileSync(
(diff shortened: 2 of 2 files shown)


itstimwhite:
## Review Summary

### Critical Issues

#### 1. Invalid Tailwind CSS syntax for border custom property
**`apps/web/components/features/demo/OnboardingDemoContent.tsx:65`**  
The class `border-(--linear-btn-primary-border)` uses parentheses for an arbitrary value, which is **not valid** standard Tailwind CSS. Tailwind requires square brackets for arbitrary values: `border-[var(--linear-btn-primary-border)]`. The same issue exists for `hover:border-(--linear-btn-primary-hover)`.

- **Why it matters**: This class will likely be ignored by Tailwind’s compiler, meaning the border color will fall back to the default (likely `transparent`) and the selected state styling will be broken. The test file also expects this invalid syntax, so the test will pass but the actual rendering will be incorrect.
- **Suggested fix**: Change both occurrences to use the correct Tailwind arbitrary value syntax:
  ```tsx
  'border-[var(--linear-btn-primary-border)] ... hover:border-[var(--linear-btn-primary-hover)]'
  ```
  Alternatively, define a custom utility class in the Tailwind config to encapsulate this variable reference.

### Warnings

#### 2. Insufficient design guard against accent colors
**`apps/web/tests/unit/demo/onboarding-demo-system-b-style-guard.test.ts:28-30`**  
The test asserts that `bg-btn-primary`, `text-btn-primary-foreground`, and `hover:bg-btn-primary-hover` are present, but it does **not** verify that these tokens resolve to non-accent colors. If the design system defines `btn-primary` as a blue/purple shade (e.g., `--color-btn-primary: blue-600`), the style guard would pass even though the button still uses an accent color.

- **Why it matters**: The goal of the PR is to ensure the step switcher remains visually neutral. Checking only class names is a weak contract. A future change could redirect `--color-btn-primary` to an accent hue without triggering the test.
- **Suggested approach**: Add a CSS variable value assertion or compile the Tailwind classes and snapshot the computed style. At minimum, document that the token must be neutrally mapped (e.g., gray tones) and link to the relevant design token definition.

#### 3. Test hard-codes a potentially fragile path
**`apps/web/tests/unit/demo/onboarding-demo-system-b-style-guard.test.ts:10-11`**  
`appRoot` is computed as `resolve(__dirname, '../../..')`. If the test file is ever moved or the project structure changes, this path may break.

- **Why it matters**: While acceptable for a source guard test, it’s better to use a project-level alias or a relative path from the repository root via a config constant to avoid silent failures.
- **Suggested fix**: Consider using `process.cwd()` with a known project root marker, or import the component directly and check its runtime behaviour rather than reading the file system.

### Suggestions

- The transition class `duration-subtle` is a custom token – ensure it is defined in the Tailwind config (likely already).
- The `min-h-7` class on the step switcher buttons is good for ensuring consistent height. Consider also applying it to the idle state to guarantee no layout shift between selected and not-selected.
- The test suite runs `vitest` and not `playwright` – for visual fidelity, a future screenshot test would be valuable, but the current approach is acceptable for a style guard.

### What Looks Good
- ✅ The PR isolates changes to a single component and adds a dedicated regression test.
- ✅ The forbidden patterns list (`bg-accent-token text-white`, `bg-blue-\d`, etc.) is comprehensive and prevents the exact accent styling from being reintroduced.
- ✅ The commit message follows the conventional commit standard with appropriate scope and description.
- ✅ The changes respect the Jovie convention of no decorative hover motion (only `transition-colors` on the switch, which is functional).
- ✅ The logic uses existing design tokens (`bg-btn-primary`, `shadow-button-inset`) rather than hardcoded values – aligns with system consistency.

itstimwhite (commented):

PR #10348, opened 2026-06-08
feat(eslint): enforce canonical UI label casing (JOV-2155)

itstimwhite:
## Summary

Adds a custom ESLint rule `@jovie/canonical-ui-label-casing` that enforces DESIGN.md text casing invariants for UI-facing code:

- **Title Case** for labels, headings, buttons, badges, nav items, and column headers
- **Sentence case** for body copy, descriptions, tooltips, placeholders, and toast messages
- Flags ALL CAPS labels (except approved abbreviations like LTV/SMS/UTM)
- Flags lowercase first words on visible labels

## Implementation

- `apps/web/eslint-rules/canonical-ui-label-casing-utils.js` — shared validators + scope helpers
- `apps/web/eslint-rules/canonical-ui-label-casing.js` — ESLint rule (JSX components/props, object config keys, toast calls)
- `apps/web/eslint-rules/canonical-ui-label-casing.test.ts` — RuleTester coverage
- Wired into `apps/web/eslint.config.js` as `@jovie/canonical-ui-label-casing: error`

Allow intentional exceptions via inline comment: `/* ui-casing-allow: reason */`

## Verification

- `pnpm --filter @jovie/web run typecheck -- --pretty false` ✅
- `pnpm biome check --write <edited files>` ✅
- `pnpm --filter web exec vitest run eslint-rules/canonical-ui-label-casing.test.ts` ✅ (18 tests)

## Notes

Rule scope is UI-facing files under `components/`, `app/`, and `packages/ui/` (excluding tests, stories, lib/, data/). Lint-staged runs ESLint on changed files only, so existing debt is surfaced incrementally when files are edited.

## Summary by CodeRabbit

* **Chores**
  * Added development tooling to enforce consistent text casing conventions across UI component labels and content, improving code quality standards and consistency during development.

### DIFF
--- a/apps/web/eslint-rules/canonical-ui-label-casing-utils.js
+++ b/apps/web/eslint-rules/canonical-ui-label-casing-utils.js
@@ -1,0 +1,405 @@
+'use strict';
+
+/**
+ * Shared casing validators for canonical UI label linting.
+ *
+ * Rules mirror DESIGN.md → Text Casing and .claude/rules/ui.md:
+ * - Title Case for labels, headings, buttons, nav, column headers, badges
+ * - Sentence case for body text, descriptions, tooltips, toasts
+ * - Never ALL CAPS (except approved abbreviations)
+ * - Never lowercase the first word of a visible label
+ */
+
+const ABBREVIATIONS = new Set([
+  'AI',
+  'API',
+  'CSV',
+  'FAQ',
+  'GDPR',
+  'ID',
+  'LTV',
+  'OAuth',
+  'OS',
+  'PDF',
+  'QR',
+  'SMS',
+  'SSO',
+  'UI',
+  'URL',
+  'USD',
+  'UTM',
+  'UX',
+  'YC',
+]);
+
+/** Brand or product words that keep internal capitalization. */
+const BRAND_WORDS = new Set([
+  'Apple',
+  'GitHub',
+  'Instagram',
+  'LinkedIn',
+  'SoundCloud',
+  'Spotify',
+  'TikTok',
+  'YouTube',
+  'iPhone',
+  'vCard',
+]);
+
+const ALLOWLIST_COMMENT_PATTERN = /ui-casing-allow:/i;
+
+/** JSX components whose child text should be Title Case. */
+const TITLE_CASE_COMPONENTS = new Set([
+  'AlertTitle',
+  'Badge',
+  'BreadcrumbPage',
+  'Button',
+  'CardTitle',
+  'DialogTitle',
+  'DrawerTitle',
+  'DropdownMenuLabel',
+  'FormLabel',
+  'Label',
+  'SheetTitle',
+  'SidebarMenuButton',
+  'TableHead',
+  'TabsTrigger',
+]);
+
+/** JSX components whose child text should be sentence case. */
+const SENTENCE_CASE_COMPONENTS = new Set([
+  'AlertDescription',
+  'CardDescription',
+  'DialogDescription',
+  'DrawerDescription',
+  'SheetDescription',
+  'TooltipContent',
+]);
+
+const HEADING_ELEMENTS = new Set(['h1', 'h2', 'h3', 'h4']);
+
+/** JSX props that carry Title Case UI labels. */
+const TITLE_CASE_PROPS = new Set(['aria-label', 'label']);
+
+/** JSX props that carry sentence-case body copy. */
+const SENTENCE_CASE_PROPS = new Set(['description', 'placeholder']);
+
+/** Object literal keys for config arrays (nav items, table columns, etc.). */
+const TITLE_CASE_OBJECT_KEYS = new Set(['header', 'label']);
+
+/** Object literal keys for sentence-case copy in config objects. */
+const SENTENCE_CASE_OBJECT_KEYS = new Set(['description']);
+
+const TOAST_CALLEE_NAMES = new Set([
+  'error',
+  'info',
+  'message',
+  'success',
+  'warning',
+]);
+
+const UI_SCOPED_PATH_PATTERNS = [
+  /\/components\//,
+  /\/app\//,
+  /\/packages\/ui\//,
+];
+
+const ALLOWED_PATH_FRAGMENTS = [
+  '.test.ts',
+  '.test.tsx',
+  '.spec.ts',
+  '.spec.tsx',
+  '.stories.ts',
+  '.stories.tsx',
+  '__tests__/',
+  '/data/',
+  '/eslint-rules/',
+  '/lib/',
+  '/scripts/',
+  '/tests/',
+  '/types/',
+  'app/api/',
+  'clerkLocalization',
+];
+
+/**
+ * @param {string} value
+ */
+function normalizeText(value) {
+  return value.replace(/\s+/g, ' ').trim();
+}
+
+/**
+ * @param {string} word
+ */
+function stripEdgePunctuation(word) {
+  return word.replace(/^[^A-Za-z0-9]+|[^A-Za-z0-9]+$/g, '');
+}
+
+/**
+ * @param {string} value
+ * @returns {string[]}
+ */
+function getWords(value) {
+  return normalizeText(value)
+    .split(' ')
+    .map(stripEdgePunctuation)
+    .filter(Boolean);
+}
+
+/**
+ * @param {string} word
+ */
+function isAbbreviation(word) {
+  if (!word) return false;
+  if (ABBREVIATIONS.has(word)) return true;
+  return (
+    word.length >= 2 && word === word.toUpperCase() && /^[A-Z0-9]+$/.test(word)
+  );
+}
+
+/**
+ * @param {string} word
+ */
+function isBrandWord(word) {
+  return BRAND_WORDS.has(word);
+}
+
+/**
+ * @param {string} value
+ */
+function shouldSkipLiteral(value) {
+  const trimmed = normalizeText(value);
+  if (!trimmed) return true;
+  if (!/[A-Za-z]/.test(trimmed)) return true;
+  if (/^https?:\/\//i.test(trimmed)) return true;
+  if (/[{}\[\]$`]/.test(trimmed)) return true;
+  if (/^\/.+\//.test(trimmed)) return true;
+  if (/^[a-z]+(?:-[a-z]+)+$/.test(trimmed)) return true;
+  if (/^[a-z]+(?:_[a-z]+)+$/.test(trimmed)) return true;
+  if (/^v?\d+(?:\.\d+)+$/.test(trimmed)) return true;
+  return false;
+}
+
+/**
+ * @param {string} word
+ */
+function toCanonicalWord(word) {
+  const leading = word.match(/^[^A-Za-z0-9]*/)?.[0] ?? '';
+  const trailing = word.match(/[^A-Za-z0-9]*$/)?.[0] ?? '';
+  const core = word.slice(leading.length, word.length - trailing.length);
+  if (!core) return word;
+  if (isAbbreviation(core) || isBrandWord(core)) return word;
+  return `${leading}${core.charAt(0).toUpperCase()}${core.slice(1).toLowerCase()}${trailing}`;
+}
+
+/**
+ * @param {string} value
+ */
+function toCanonicalTitleCase(value) {
+  return normalizeText(value).split(' ').map(toCanonicalWord).join(' ');
+}
+
+/**
+ * @param {string} value
+ */
+function isAllCapsPhrase(value) {
+  const words = getWords(value);
+  if (words.length === 0) return false;
+  if (words.length === 1 && isAbbreviation(words[0])) return false;
+  return words.every(
+    word => word === word.toUpperCase() && /^[A-Z]{2,}$/.test(word)
+  );
+}
+
+/**
+ * @param {string} value
+ */
+function isTitleCase(value) {
+  if (shouldSkipLiteral(value)) return true;
+
+  const trimmed = normalizeText(value);
+  if (trimmed.charAt(0) !== trimmed.charAt(0).toUpperCase()) return false;
+  if (isAllCapsPhrase(trimmed)) return false;
+
+  const words = getWords(trimmed);
+  for (const word of words) {
+    if (isAbbreviation(word) || isBrandWord(word)) continue;
+    if (!/^[A-Z][a-z0-9'&]*$/.test(word)) return false;
+  }
+
+  return trimmed === toCanonicalTitleCase(trimmed);
+}
+
+/**
+ * @param {string} value
+ */
+function isSentenceCase(value) {
+  if (shouldSkipLiteral(value)) return true;
+
+  const trimmed = normalizeText(value);
+  const words = getWords(trimmed);
+  const firstWord = words[0] ?? '';
+  const startsWithBrandWord =
+    isBrandWord(firstWord) ||
+    (firstWord.length > 1 &&
+      firstWord.startsWith('i') &&
+      /[A-Z]/.test(firstWord[1]));
+
+  if (
+    trimmed.charAt(0) !== trimmed.charAt(0).toUpperCase() &&
+    !startsWithBrandWord
+  ) {
+    return false;
+  }
+  if (isAllCapsPhrase(trimmed)) return false;
+
+  for (let index = 1; index < words.length; index += 1) {
+    const word = words[index];
+    if (isAbbreviation(word) || isBrandWord(word)) continue;
+    if (/^[A-Z][a-z]/.test(word)) return false;
+  }
+
+  return true;
+}
+
+/**
+ * @param {string} filename
+ */
+function isUiScopedFile(filename) {
+  const normalized = filename.replaceAll('\\', '/');
+  if (!UI_SCOPED_PATH_PATTERNS.some(pattern => pattern.test(normalized))) {
+    return false;
+  }
+  return !ALLOWED_PATH_FRAGMENTS.some(fragment =>
+    normalized.includes(fragment)
+  );
+}
+
+/**
+ * @param {import('eslint').Rule.RuleContext} context
+ * @param {import('@typescript-eslint/types').TSESTree.Node} node
+ */
+function hasAllowlistComment(node, context) {
+  const sourceCode = context.sourceCode || context.getSourceCode();
+  const comments = sourceCode.getAllComments ? sourceCode.getAllComments() : [];
+  const nodeLine = node.loc?.start?.line;
+  if (nodeLine == null) return false;
+
+  return comments.some(comment => {
+    const commentLine = comment.loc?.start?.line;
+    if (commentLine == null) return false;
+    if (commentLine !== nodeLine && commentLine !== nodeLine - 1) return false;
+    return ALLOWLIST_COMMENT_PATTERN.test(comment.value);
+  });
+}
+
+/**
+ * @param {import('@typescript-eslint/types').TSESTree.JSXOpeningElement | import('@typescript-eslint/types').TSESTree.JSXOpeningFragment} openingElement
+ */
+function getJsxElementName(openingElement) {
+  const nameNode = openingElement.name;
+  if (!nameNode) return null;
+
+  if (nameNode.type === 'JSXIdentifier') {
+    return nameNode.name;
+  }
+
+  if (nameNode.type === 'JSXMemberExpression') {
+    return nameNode.property.name;
+  }
+
+  return null;
+}
+
+/**
+ * @param {import('@typescript-eslint/types').TSESTree.JSXAttribute} attribute
+ */
+function getJsxAttributeName(attribute) {
+  const nameNode = attribute.name;
+  if (nameNode.type === 'JSXIdentifier') return nameNode.name;
+  if (nameNode.type === 'JSXNamespacedName') {
+    return `${nameNode.namespace.name}:${nameNode.name.name}`;
+  }
+  return null;
+}
+
+/**
+ * @param {import('@typescript-eslint/types').TSESTree.JSXAttribute['value']} valueNode
+ */
+function getStringFromJsxAttributeValue(valueNode) {
+  if (!valueNode) return null;
+  if (valueNode.type === 'Literal' && typeof valueNode.value === 'string') {
+    return valueNode.value;
+  }
+  if (
+    valueNode.type === 'JSXExpressionContainer' &&
+    valueNode.expression.type === 'Literal' &&
+    typeof valueNode.expression.value === 'string'
+  ) {
+    return valueNode.expression.value;
+  }
+  return null;
+}
+
+/**
+ * @param {import('@typescript-eslint/types').TSESTree.Node} node
+ * @returns {string | null}
+ */
+function getStringLiteralValue(node) {
+  if (node.type === 'Literal' && typeof node.value === 'string') {
+    return node.value;
+  }
+
+  if (node.type === 'JSXText') {
+    return node.value;
+  }
+
+  if (node.type === 'TemplateElement') {
+    return node.value.cooked ?? node.value.raw;
+  }
+
+  return null;
+}
+
+/**
+ * @param {import('@typescript-eslint/types').TSESTree.CallExpression} node
+ */
+function isToastCall(node) {
+  const callee = node.callee;
+  if (callee.type === 'Identifier' && callee.name === 'toast') return true;
+  if (
+    callee.type === 'MemberExpression' &&
+    !callee.computed &&
+    callee.object.type === 'Identifier' &&
+    callee.object.name === 'toast' &&
+    callee.property.type === 'Identifier' &&
+    TOAST_CALLEE_NAMES.has(callee.property.name)
+  ) {
+    return true;
+  }
+  return false;
+}
+
+module.exports = {
+  ALLOWLIST_COMMENT_PATTERN,
+  BRAND_WORDS,
+  HEADING_ELEMENTS,
+  SENTENCE_CASE_COMPONENTS,
+  SENTENCE_CASE_OBJECT_KEYS,
+  SENTENCE_CASE_PROPS,
+  TITLE_CASE_COMPONENTS,
+  TITLE_CASE_OBJECT_KEYS,
+  TITLE_CASE_PROPS,
+  getJsxAttributeName,
+  getJsxElementName,
+  getStringFromJsxAttributeValue,
+  getStringLiteralValue,
+  hasAllowlistComment,
+  isSentenceCase,
+  isTitleCase,
+  isToastCall,
+  isUiScopedFile,
+  shouldSkipLiteral,
+  toCanonicalTitleCase,
+};
--- a/apps/web/eslint-rules/canonical-ui-label-casing.js
+++ b/apps/web/eslint-rules/canonical-ui-label-casing.js
@@ -1,0 +1,247 @@
+/**
+ * ESLint rule to enforce canonical casing for UI labels.
+ *
+ * DESIGN.md → Text Casing:
+ * - Title Case for labels, headings, buttons, badges, nav items, column headers
+ * - Sentence case for body text, descriptions, tooltips, and toasts
+ *
+ * Allow intentional exceptions with an inline comment:
+ *   {/* ui-casing-allow: brand lockup * /}
+ *
+ * Scope: UI-facing files under components/, app/, and packages/ui/.
+ * Tests, stories, data files, and lib/ utilities are excluded.
+ */
+
+'use strict';
+
+const {
+  HEADING_ELEMENTS,
+  SENTENCE_CASE_COMPONENTS,
+  SENTENCE_CASE_OBJECT_KEYS,
+  SENTENCE_CASE_PROPS,
+  TITLE_CASE_COMPONENTS,
+  TITLE_CASE_OBJECT_KEYS,
+  TITLE_CASE_PROPS,
+  getJsxAttributeName,
+  getJsxElementName,
+  getStringFromJsxAttributeValue,
+  getStringLiteralValue,
+  hasAllowlistComment,
+  isSentenceCase,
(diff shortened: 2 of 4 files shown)


itstimwhite:
## AI Code Review (nemotron-3-120b:free) 🟢 (confidence: 2/5)

We are given a PR that adds a new ESLint rule to enforce canonical UI label casing.

We need to review the changes in the following files:
1. apps/web/eslint-rules/canonical-ui-label-casing-utils.js
2. apps/web/eslint-rules/canonical-ui-label-casing.js
3. apps/web/eslint-rules/canonical-ui-label-casing.test.ts
4. apps/web/eslint.config.js

We are to look for:
- Logic bugs (race conditions, null dereferences, off-by-one, incorrect conditionals, wrong return values, silent edge-case drops)
- Security issues
- Cross-file regressions
- Convention violations
- Missing tests
- Design drift
- Performance issues

However, note that we are not to comment on style preferences, variable naming (unless misleading), minor formatting, or things that are technically correct but could be better.

Let's go through each file.

### File: apps/web/eslint-rules/canonical-ui-label-casing-utils.js

This file contains utility functions for the ESLint rule.

We note the following:

1. The function `isTitleCase` and `isSentenceCase` are used to check the casing of a string.

2. The function `shouldSkipLiteral` is used to skip certain literals (like URLs, numbers, etc.) that don't need casing enforcement.

3. The function `isUiScopedFile` determines if a file is in the scope of the rule (components/, app/, packages/ui/).

4. The function `hasAllowlistComment` checks for an inline comment to allow an exception.

Potential issues:

- In `isTitleCase` and `isSentenceCase`, we see that they call `shouldSkipLiteral` and return `true` if it should be skipped. This is acceptable.

- However, note that in `isTitleCase`, after checking `shouldSkipLiteral`, we have:
  ```javascript
  const trimmed = normalizeText(value);
  if (trimmed.charAt(0) !== trimmed.charAt(0).toUpperCase()) return false;
  ```
  This checks that the first character is uppercase. But note: what if the string is empty? We have `shouldSkipLiteral` which returns true for empty string (because `normalizeText` of empty string is empty, and then `!/[A-Za-z]/.test(trimmed)` is true because there are no letters). So we are safe.

- Similarly, in `isSentenceCase`, we check the first character. Again, `shouldSkipLiteral` would have caught empty strings.

- The function `isAllCapsPhrase`:
  ```javascript
  function isAllCapsPhrase(value) {
    const words = getWords(value);
    if (words.length === 0) return false;
    if (words.length === 1 && isAbbreviation(words[0])) return false;
    return words.every(
      word => word === word.toUpperCase() && /^[A-Z]{2,}$/.test(word)
    );
  }
  ```
  This function returns false for a single abbreviation (like "LTV") because of the second condition. But note: the condition `words.length === 1 && isAbbreviation(words[0])` returns false, so we skip the every check and return false. That is correct because we don't want to flag a single abbreviation as all caps.

  However, note that the every condition also checks `/^[A-Z]{2,}$/` which requires at least two uppercase letters. So a single letter abbreviation (like "A") would not be caught by the every condition? But note: we have the condition `if (words.length === 1 && isAbbreviation(words[0])) return false;` and `isAbbreviation` for a single letter: 
  ```javascript
  function isAbbreviation(word) {
    if (!word) return false;
    if (ABBREVIATIONS.has(word)) return true;
    return (
      word.length >= 2 && word === word.toUpperCase() && /^[A-Z0-9]+$/.test(word)
    );
  }
  ```
  So a single letter would not be in ABBREVIATIONS and would fail `word.length>=2`, so it's not an abbreviation. Then we would go to the every condition and check: 
    word === word.toUpperCase() -> true for a single uppercase letter
    /^[A-Z]{2,}$/ -> false because it's only one letter.
  So the every condition would return false, and then `isAllCapsPhrase` would return false. That is acceptable because we don't consider a single uppercase letter as an all caps phrase? But note: the rule says "Never ALL CAPS (except approved abbreviations)". A single uppercase letter is not an approved abbreviation (unless it's in the list, which it isn't). However, the rule might be intended for multi-letter words. But note: the rule also says "Flags ALL CAPS labels (except approved abbreviations like LTV/SMS/UTM)". So a single letter might be acceptable? 

  However, looking at the DESIGN.md (which we don't have, but from the description) we are to enforce Title Case and sentence case. A single uppercase letter in a label might be acceptable? For example, "A" as a label? But note: the rule `isTitleCase` would require the first letter to be uppercase and then the rest lowercase. For a single letter, that would be just uppercase. So "A" would pass `isTitleCase`? Let's see:

  In `isTitleCase` for "A":
    - `shouldSkipLiteral` returns false (because it's a letter).
    - `trimmed` is "A"
    - `trimmed.charAt(0) === 'A'` and `trimmed.charAt(0).toUpperCase()` is 'A', so passes.
    - `isAllCapsPhrase("A")` returns false (as above).
    - Then we get words: ["A"]
    - For the word "A": 
        isAbbreviation? -> false (because length<2 and not in ABBREVIATIONS)
        isBrandWord? -> false
        Then we check: `/^[A-Z][a-z0-9'&]*$/.test("A")` -> 
          [A-Z] matches 'A', then [a-z0-9'&]* matches zero times -> so it matches.
    - Then we check: `trimmed === toCanonicalTitleCase(trimmed)` -> 
        toCanonicalTitleCase("A") -> 
          normalizeText("A") -> "A"
          split -> ["A"]
          map: toCanonicalWord("A") -> 
            leading: '' (no non-alnum at start)
            trailing: '' (no non-alnum at end)
            core: "A"
            isAbbreviation? false, isBrandWord? false -> 
            returns "A" (because leading + core[0].toUpperCase() + core.slice(1).toLowerCase() + trailing -> "A" + "" + "" -> "A")
          so toCanonicalTitleCase("A") is "A"
        so trimmed === "A" -> true.

  Therefore, "A" is considered Title Case.

  Similarly, in sentence case: 
    - First character must be uppercase (or a brand word that starts with lowercase? but note the brand word check: 
        const startsWithBrandWord =
          isBrandWord(firstWord) ||
          (firstWord.length > 1 &&
            firstWord.startsWith('i') &&
            /[A-Z]/.test(firstWord[1]));
      For "A": 
        isBrandWord("A") -> false
        firstWord.length>1? false -> so startsWithBrandWord is false.
    - Then we check: 
        if (trimmed.charAt(0) !== trimmed.charAt(0).toUpperCase() && !startsWithBrandWord) 
        -> 'A' !== 'A' -> false, so condition fails -> we don't return false here.
    - Then we check isAllCapsPhrase -> false (as above).
    - Then we loop over words (only one word, index=0, but we start at index=1 so loop doesn't run).
    - Then return true.

  So "A" is considered both Title Case and sentence case? That might be a problem because we don't know which context it's in. But note: the rule will check the context (whether it's in a Title Case component or sentence case component) and then use the appropriate checker.

  However, note that the rule might flag a single uppercase letter in a sentence case context as needing sentence case? But our sentence case checker returns true for "A", so it would not flag it. Similarly, in a Title Case context, it returns true so it wouldn't flag.

  So it's acceptable.

- Another point: the function `toCanonicalTitleCase` is used to generate a suggestion for Title Case. However, note that for sentence case, the suggestion is generated as:
      value.charAt(0).toUpperCase() + value.slice(1)
  This is not exactly the same as the canonical sentence case we define? Because our sentence case allows for brand words and abbreviations in the rest of the string to remain as they are. But note: the suggestion for sentence case is only the first letter capitalized and the rest as is. However, our `isSentenceCase` function does more: it checks that the rest of the words are not in Title Case (i.e., they don't start with an uppercase letter unless they are an abbreviation or brand word). 

  But note: the suggestion we are providing for sentence case is only changing the first letter to uppercase and leaving the rest as is. This might not be sufficient if the rest of the string has words that are incorrectly cased (like "Hello World" in a sentence case context: we would suggest "Hello world" -> but wait, we are only changing the first letter? Actually, we are doing:
      value.charAt(0).toUpperCase() + value.slice(1)
  For "hello world", we get "Hello world". But note: the word "world" is in lowercase, which is correct for sentence case. However, if we had "hELLO wORLD", we would get "HELLO wORLD" which is not correct. But note: our `isSentenceCase` would return false for "hELLO wORLD" because:
      - First character: 'h' is not uppercase and not a brand word -> returns false early.

  So we would never get to the suggestion for that case because we return early.

  However, consider a string that passes the first character check but has an incorrectly cased word in the middle? For example: "Hello wORLD" (note: the first word is correctly cased, but the second word is all caps). 
      - First character: 'H' is uppercase -> passes the first check.
      - isAllCapsPhrase: 
          words = ["Hello", "wORLD"]
          "Hello" is not all caps -> so every returns false -> isAllCapsPhrase returns false.
      - Then we check the words from index=1: 
          word = "wORLD"
          isAbbreviation? -> no (because not all uppercase? note: "wORLD" has a lowercase 'w' at the start -> so not all uppercase -> fails the abbreviation check? Actually, our isAbbreviation requires the word to be all uppercase. So it fails.
          isBrandWord? -> no.
          Then we check: /^[A-Z][a-z]/.test("wORLD") -> 
            first character 'w' is not uppercase -> fails -> so we return false in the loop.

  So we return false for isSentenceCase.

  Then we would suggest: 
      value.charAt(0).toUpperCase() + value.slice(1) -> 
        "H" + "ello wORLD" -> "Hello wORLD"

  But note: the correct sentence case for "Hello wORLD" would be "Hello world". However, our suggestion is not doing that. We are only fixing the first letter.

  This is a problem because the suggestion is not helpful. We should be suggesting the canonical sentence case.

  However, note that we don't have a function to convert to canonical sentence case. We only have `toCanonicalTitleCase`.

  We should consider adding a function for canonical sentence case? But note: the rule is only required to suggest a fix. The current suggestion for sentence case is only capitalizing the first letter and leaving the rest as is. This might not be sufficient.

  But note: the rule's purpose is to flag the violation, and the suggestion is a helper. The developer can then adjust the rest. However, it would be better to provide a more accurate suggestion.

  However, looking at the test cases, we see that for sentence case violations, the test expects the suggestion to be the string with the first letter capitalized and the rest unchanged? Let's look at the test:

  In the test file, we have:
      {
        filename,
        code: `<CardDescription>Share Your Profile Link On Social Media.</CardDescription>`,
        errors: [{ messageId: 'useSentenceCase' }],
      }

  The suggestion for this would be: 
      "Share Your Profile Link On Social Media." -> 
        first letter: 'S' -> already uppercase -> so we do: 
          'S' + "hare Your Profile Link On Social Media." -> "Share your profile link on social media."? 
        But wait, we are only changing the first letter to uppercase? Actually, we are doing:
          value.charAt(0).toUpperCase() + value.slice(1)
        For the string "Share Your Profile Link On Social Media.":
          first letter: 'S' -> toUpperCase() is 'S'
          slice(1): "hare Your Profile Link On Social Media."
          so we get "Share your profile link on social media."? 
        But note: the slice(1) is exactly the string from index 1 to the end, so we are not changing the rest.

        However, the original string has uppercase letters in the middle. We are leaving them as is.

        The test does not check the suggestion, only the messageId.

  So the test passes even if the suggestion is not perfect.

  But note: the rule might be improved by providing a better suggestion for sentence case. However, the problem is that we don't have a canonical sentence case function. We could create one similar to `toCanonicalTitleCase` but for sentence case? 

  However, note that sentence case is defined as:
      - First word: capitalized (first letter uppercase, the rest lowercase) unless it's a brand word or abbreviation (then we leave it as is).
      - Subsequent words: all lowercase unless they are an abbreviation or brand word.

  We could write a function `toCanonicalSentenceCase` that does:
      - Split the string into words.
      - For the first word:
            if it's an abbreviation or brand word -> leave as is
            else -> capitalize (first letter uppercase, rest lowercase)
      - For the rest of the words:
            if it's an abbreviation or brand word -> leave as is
            else -> lowercase

  But note: the rule already has `isSentenceCase` and we are using that to check. We could use the same logic to generate the suggestion.

  However, the PR does not include such a function. And the test does not require the suggestion to be perfect. So this might be acceptable for now.

  But note: the rule is marked as a "problem" type, and we are providing a suggestion. It would be better to have a useful suggestion.

  However, given the scope of the review, we are to look for critical bugs. This is more of a usability issue.

  Let's see if there are any critical bugs.

### File: apps/web/eslint-rules/canonical-ui-label-casing.js

This is the main rule.

We note:

1. The rule first checks if the file is in the UI scope using `isUiScopedFile`. If not, it returns an empty object (no rules).

2. Then it defines visitors for:
   - JSXElement
   - JSXAttribute
   - Property (for object literals)
   - CallExpression (for toast calls)

Potential issues:

- In the JSXElement visitor:
      const componentName = getJsxElementName(node.openingElement);
      if (!componentName) return;

      const staticChildren = collectStaticTextChildren(node.children);
      if (staticChildren.length === 0) return;

      if (
        TITLE_CASE_COMPONENTS.has(componentName) ||
        HEADING_ELEMENTS.has(componentName)
      ) {
        for (const child of staticChildren) {
          reportCasingViolation(
            context,
            child.node,
            child.value,
            'title',
            `<${componentName}>`
          );
        }
        return;
      }

      if (SENTENCE_CASE_COMPONENTS.has(componentName)) {
        for (const child of staticChildren) {
          reportCasingViolation(
            context,
            child.node,
            child.value,
            'sentence',
            `<${componentName}>`
          );
        }
      }

  What about a component that is both in TITLE_CASE_COMPONENTS and SENTENCE_CASE_COMPONENTS? The sets are disjoint? We don't see any overlap in the defined sets in the utils file. But note: the sets are defined as constants and we don't see any common elements. So it's safe.

  However, note that the rule returns after the Title Case check. So if a component is in both, it would only be checked as Title Case. But since the sets are disjoint, it's not an issue.

- In the JSXAttribute visitor:
      const propName = getJsxAttributeName(node);
      if (!propName) return;

      const stringValue = getStringFromJsxAttributeValue(node.value);
      if (stringValue == null) return;

      if (TITLE_CASE_PROPS.has(propName)) {
        reportCasingViolation(
          context,
          node,
          stringValue,
          'title',
          `${propName} prop`
        );
        return;
      }

      if (SENTENCE_CASE_PROPS.has(propName)) {
        reportCasingViolation(
          context,
          node,
          stringValue,
          'sentence',
          `${propName} prop`
        );
      }

  Note: we return after the Title Case check. So if a prop is in both (which it shouldn't be) we would only check as Title Case. But the sets TITLE_CASE_PROPS and SENTENCE_CASE_PROPS are defined as:
      TITLE_CASE_PROPS: ['aria-label', 'label']
      SENTENCE_CASE_PROPS: ['description', 'placeholder']
  So no overlap.

- In the Property visitor:
      if (node.key.type !== 'Identifier') return;


---
*Automated review — please verify all suggestions before applying. This review is independent of the code generation process.*

itstimwhite (commented):
Automated review (nvidia/nemotron-3-super-120b-a12b:free, confidence 2/5)

itstimwhite:

Branch: 'tim/jov-2155'
State: BLOCKED



To resolve:
1. git fetch origin main
2. git checkout tim/jov-2155
3. git rebase origin/main
5. git push --force-with-lease origin tim/jov-2155

itstimwhite:
scope-judge: human override — inconclusive verdict due to OpenRouter infra; scope reviewed and aligned with Linear ticket intent.

itstimwhite:
## AI Code Review (nemotron-3-120b:free) 🟢 (confidence: 2/5)

We are given a PR that adds a new ESLint rule to enforce canonical UI label casing.

We are to review the code for:
1. Logic bugs
2. Security
3. Cross-file regressions
4. Convention violations
5. Missing tests
6. Design drift
7. Performance

We are to output in the specified format.

Let's break down the changes:

Files changed:
1. apps/web/eslint-rules/canonical-ui-label-casing-utils.js
2. apps/web/eslint-rules/canonical-ui-label-casing.js
3. apps/web/eslint-rules/canonical-ui-label-casing.test.ts
4. apps/web/eslint.config.js

We'll review each file for issues.

Starting with the utils file (canonical-ui-label-casing-utils.js):

We note that the file is long (405 lines). We'll look for potential logic bugs.

Some observations:

- The function `isTitleCase` and `isSentenceCase` have a helper `shouldSkipLiteral` that returns true for certain strings (like numbers, URLs, etc.) to skip validation.

- In `isTitleCase`:
  - It checks if the first character is uppercase (unless it's a brand word that starts with lowercase, like 'iPhone'? Actually, note the brand word set includes 'iPhone', but the function `isBrandWord` is used in `toCanonicalWord` and also in `isTitleCase` and `isSentenceCase` for skipping? Let's see:

  In `isTitleCase`:
      if (isAbbreviation(word) || isBrandWord(word)) continue;

  So brand words are skipped in the validation loop? Actually, no: the loop continues (so we don't check the word) but note that the brand word might be in the middle of a string and we are allowing it to keep its internal capitalization.

  However, note that the function `toCanonicalWord` also preserves brand words and abbreviations.

  But there's a potential issue: the function `isTitleCase` does not allow a brand word that is all caps? For example, if we have "GITHUB", note that 'GITHUB' is not in the brand words set (because the set has 'GitHub'). So it would be considered as an abbreviation? Only if it's in the ABBREVIATIONS set or if it's all caps and length>=2 and only alphanumeric. So "GITHUB" would be considered an abbreviation? Actually, the condition for abbreviation in `isAbbreviation` is:
      if (!word) return false;
      if (ABBREVIATIONS.has(word)) return true;
      return (
        word.length >= 2 && word === word.toUpperCase() && /^[A-Z0-9]+$/.test(word)
      );

  So "GITHUB" would be considered an abbreviation (because it's all caps and length>=2 and only alphanumeric). Then in `isTitleCase`, we skip checking that word (because of the continue). So it would pass.

  However, note that the brand word set is for words that have internal capitalization (like 'GitHub', 'iPhone'). The function `isBrandWord` only checks exact match. So if we have "GITHUB", it's not in the brand word set, but it is considered an abbreviation and skipped.

  This seems acceptable.

- In `isSentenceCase`:
  - It checks the first word: if it's not uppercase and not a brand word (with a special check for words starting with 'i' and then uppercase, like 'iPhone') then it returns false.
  - Then for the rest of the words, it checks that they do not start with an uppercase letter (unless they are abbreviation or brand word).

  Example: "iPhone is great" -> 
      first word: "iPhone" -> isBrandWord? yes -> so we don't return false for the first word.
      then for the word "is": it's not abbreviation or brand, and it starts with lowercase -> okay.
      "great": same.

  But note: what about "iPhone IS great"? 
      The word "IS": 
          isAbbreviation? -> it's in the set? No (unless we have "IS" in the set, but we don't). 
          Then we check: word.length>=2, word===word.toUpperCase() -> true, and /^[A-Z0-9]+$/.test(word) -> true -> so it's considered an abbreviation -> skipped.

  So "iPhone IS great" would pass as sentence case? But note: the rule for sentence case is that only the first word is capitalized and the rest are lowercase. However, we are allowing abbreviations to be uppercase anywhere. This is by design.

  However, note that the DESIGN.md might not specify that abbreviations can be in the middle of a sentence in uppercase. But the rule says: "Never ALL CAPS (except approved abbreviations)". So it's allowed.

  So the logic seems consistent.

- The function `shouldSkipLiteral` returns true for:
      - empty or only whitespace
      - no letters at all
      - starts with http:// or https://
      - contains any of [{}\[\]$`]
      - matches a regex for a regex pattern (like /.../)
      - matches kebab-case or snake_case (all lowercase with hyphens or underscores)
      - matches a version number (like v1.2.3)

  This seems reasonable to skip non-text or code-like strings.

- The function `getStringLiteralValue` handles Literal, JSXText, and TemplateElement. Note: for TemplateElement, it returns the cooked value if available, else raw. This is acceptable.

- The function `hasAllowlistComment` checks for a comment on the same line or the previous line that matches the pattern `ui-casing-allow:` (case insensitive). This is acceptable.

- The function `isUiScopedFile` checks if the file is under components/, app/, or packages/ui/ and then excludes certain paths (like tests, stories, lib/, etc.). This is acceptable.

Now, the rule file (canonical-ui-label-casing.js):

- It uses the utils and defines a rule that visits:
    JSXElement: for components that are in TITLE_CASE_COMPONENTS or HEADING_ELEMENTS, it checks the static text children.
    JSXAttribute: for props in TITLE_CASE_PROPS or SENTENCE_CASE_PROPS.
    Property: for object keys in TITLE_CASE_OBJECT_KEYS or SENTENCE_CASE_OBJECT_KEYS.
    CallExpression: for toast calls.

- We note that the rule only checks static string values. It does not check expressions that are not literals (like variables or function calls). This is acceptable because we cannot know the value at lint time.

- The function `reportCasingViolation` uses the `hasAllowlistComment` to skip if there's an allow comment.

- The suggestion for sentence case is: `value.charAt(0).toUpperCase() + value.slice(1)`. This only capitalizes the first letter and leaves the rest as is. However, note that the rest of the string might have uppercase letters that should be lowercase? For example, if we have "hELLO", the suggestion would become "HELLO", which is not correct sentence case.

  But note: the rule is only triggered when the string is not in sentence case. And the function `isSentenceCase` would have returned false for "hELLO" because:
      - The first character is lowercase -> fails the first check (unless it's a brand word starting with 'i'? but 'h' is not 'i') -> so it would be caught.
      - Then the suggestion would be "H" + "ELLO" -> "HELLO", which is not the desired sentence case (which should be "Hello").

  However, note that the rule does not attempt to fix the entire string to sentence case, only to capitalize the first letter. This is a limitation.

  But look at the test: in the invalid examples, we have:
        `<CardDescription>Share Your Profile Link On Social Media.</CardDescription>`
      which is expected to be fixed to sentence case. The current suggestion would produce:
        "Share your profile link on social media." -> but note: the current suggestion only does:
          value.charAt(0).toUpperCase() + value.slice(1)
        For the string "Share Your Profile Link On Social Media.":
          First char: 'S' -> uppercase -> remains 'S'
          Then the rest: "hare Your Profile Link On Social Media." -> so the suggestion would be the same as the input? 
          Because the first letter is already uppercase.

  Actually, the test expects that the string is not in sentence case because it has uppercase letters in the middle. The rule's `isSentenceCase` function would return false for that string because it finds an uppercase letter in the middle (the 'Y' in "Your", etc.). Then the suggestion is to capitalize the first letter and leave the rest. But that doesn't fix the middle uppercase letters.

  This is a problem.

  The rule's suggestion for sentence case is inadequate. It should convert the string to proper sentence case (first letter uppercase, the rest lowercase) but note: we have to preserve abbreviations and brand words.

  However, the rule does not have a function to convert to sentence case (like we have `toCanonicalTitleCase` for title case). We only have `toCanonicalTitleCase` for title case.

  We should have a similar function for sentence case? Or at least, the suggestion should be more intelligent.

  But note: the rule is only for linting, not for fixing. The ESLint rule can have a fixer, but this rule does not. So the suggestion is just for the user to see what it should be.

  However, the current suggestion for sentence case is misleading because it doesn't fix the middle uppercase letters.

  We see in the test that the invalid example:
        `<CardDescription>Share Your Profile Link On Social Media.</CardDescription>`
      is expected to trigger an error with messageId 'useSentenceCase'. The test does not check the suggestion, only the messageId.

  So the test passes because the messageId is correct, but the suggestion is not helpful.

  We should improve the suggestion for sentence case to at least show a properly cased version (with the first letter uppercase and the rest lowercase, but preserving abbreviations and brand words?).

  However, note that the rule does not have a function to convert to sentence case. We could create one, but it's not required for the rule to work. The rule is only reporting an error.

  But the suggestion field in the error message is meant to help the user fix it. If the suggestion is not correct, it might confuse the user.

  Given that the rule does not have a fixer, and the suggestion is only advisory, we might consider it acceptable? However, the current suggestion for sentence case is simply capitalizing the first letter and leaving the rest, which is not what we want.

  Example: 
      Input: "share YOUR profile link"
      Current suggestion: "Share YOUR profile link" -> which is still not sentence case because of "YOUR".

  We should change the suggestion for sentence case to be at least the string with the first letter capitalized and the rest in lowercase? But that would break abbreviations and brand words.

  Alternatively, we could leave the suggestion as is and note that the user must manually fix it? But that is not ideal.

  However, looking at the test cases, we don't see any example that has mixed case in the middle (like "YOUR") in the invalid examples. The invalid examples for sentence case are:
        `<CardDescription>Share Your Profile Link On Social Media.</CardDescription>`
        `toast.error('Member Removed');`
        `<TooltipContent>High Intent</TooltipContent>`

  In these examples, the problem is that every word is capitalized (title case) or the first letter of each word is uppercase. The current suggestion would not change the string at all because the first letter is already uppercase.

  So for these examples, the suggestion is the same as the input, which is not helpful.

  We might want to change the suggestion for sentence case to be the string converted to lowercase except the first letter? But that would break abbreviations.

  Example: "Share your LTV link" -> if we lowercase the rest, we get "Share your ltv link", which is wrong because LTV should be uppercase.

  Therefore, we cannot simply lowercase the rest.

  Given the complexity, and since the rule does not have a fixer, it might be acceptable to leave the suggestion as is (only fixing the first letter) and rely on the user to know that they should also fix the rest? But note: the rule is also catching title case strings in sentence case positions, and for those, the suggestion (capitalizing first letter) does nothing because the first letter is already uppercase.

  We see that the test for `<CardDescription>Share Your Profile Link On Social Media.</CardDescription>` expects an error, but the suggestion is the same string. The user would see:
        Expected "Share Your Profile Link On Social Media." -> which is the same as what they wrote.

  This is confusing.

  We should change the suggestion for sentence case to be the string converted to lowercase except the first letter? But then we break abbreviations.

  Alternatively, we could not provide a suggestion for sentence case? But the rule requires a suggestion.

  Another idea: use the same `toCanonicalTitleCase` function but then adjust for sentence case? Not exactly.

  We note that the rule already has a function `toCanonicalTitleCase` that converts a string to title case (with preservation of abbreviations and brand words). For sentence case, we want:
      - First word: capitalized (first letter uppercase, the rest lowercase) but preserving brand words and abbreviations? Actually, the first word might be a brand word (like 'iPhone') and we want to keep it as is.

  We can create a function `toCanonicalSentenceCase` that:
      - Splits the string into words.
      - For the first word: 
            if it's a brand word or abbreviation, leave it as is.
            else, capitalize the first letter and lowercase the rest.
      - For the rest of the words:
            if it's a brand word or abbreviation, leave it as is.
            else, lowercase the entire word.

  Example: 
      "share YOUR LTV link" -> 
          first word: "share" -> not brand/abbr -> "Share"
          second word: "YOUR" -> abbreviation? (if in set or all caps) -> if we consider it an abbreviation (because it's all caps and length>=2) then leave as "YOUR"
          third word: "LTV" -> abbreviation -> leave as "LTV"
          fourth word: "link" -> not brand/abbr -> "link"
          => "Share YOUR LTV link"

  But note: the word "YOUR" is not in the abbreviation set, but it is all caps and length>=2 and only alphanumeric -> so it is considered an abbreviation by our `isAbbreviation` function.

  This would work.

  However, we are not required to implement a fixer, only to provide a suggestion. We could compute the suggestion using such a function.

  But note: the rule does not currently have this function. We would have to add it to the utils.

  Given that the PR is about adding the rule and we are reviewing, we must decide if this is a bug.

  The current suggestion for sentence case is inadequate for strings that are not already starting with an uppercase letter and have other issues. However, note that the rule only triggers when the string is not in sentence case. The cases we are seeing in the tests are strings that are in title case (all words capitalized) in a sentence case context. For these, the current suggestion (capitalize first letter) does nothing because the first letter is already uppercase.

  We can change the suggestion for sentence case to be the string converted to lowercase except the first letter? But that breaks abbreviations.

  Alternatively, we can change the suggestion to be the string with the first letter uppercase and the rest lowercase, but then we note that it might break abbreviations and brand words? And we are not preserving them.

  Given the time, and since the rule is primarily for linting (not fixing), and the test suite does not check the suggestion, we might let it pass. However, we note that the suggestion is part of the user experience and should be helpful.

  We see that the rule's author did not implement a sentence case conversion function. We might consider this a minor issue.

  But note: the rule's meta says it is of type 'problem'. It doesn't have a fixer. So the suggestion is only for the message.

  We decide to note this as a warning because it might confuse users.

  However, let's look at the test file to see if there are any test cases that would expose this.

  The test file (canonical-ui-label-casing.test.ts) has:

      invalid: [
        {
          filename,
          code: `<CardDescription>Share Your Profile Link On Social Media.</CardDescription>`,
          errors: [{ messageId: 'useSentenceCase' }],
        },
        ... 
      ]

  The test only checks that there is an error with messageId 'useSentenceCase'. It does not check the suggestion.

  So the test passes.

  We might want to add a test case for the suggestion? But that is beyond the scope of this review.

  Given the above, we note that the suggestion for sentence case is not ideal but does not break the rule's core function (reporting errors). We'll mark it as a suggestion for improvement.

Now, let's look for other issues.

In the rule file, we see:

      Property(node) {
        if (node.key.type !== 'Identifier') return;
        const propName = node.key.name;
        if (
          !TITLE_CASE_OBJECT_KEYS.has(propName) &&
          !SENTENCE_CASE_OBJECT_KEYS.has(propName)
        ) {
          return;
        }

        if (
          node.value.type !== 'Literal' ||
          typeof node.value.value !== 'string'
        ) {
          return;
        }

        if (TITLE_CASE_OBJECT_KEYS.has(propName)) {
          reportCasingViolation(
            context,
            node.value,
            node.value.value,
            'title',
            `${propName} property`
          );
          return;
        }

        reportCasingViolation(
          context,
          node.value,
          node.value.value,
          'sentence',
          `${propName} property`
        );
      }

  This checks object properties. Note: it only checks if the value is a literal string. What if the value is a template literal? We have `getStringLiteralValue` in the utils that handles TemplateElement, but here we are only checking for Literal.

  We should use `getStringLiteralValue` to get the string value from the node, similar to how we do in JSXAttribute.

  Example: 
        const config

---
*Automated review — please verify all suggestions before applying. This review is independent of the code generation process.*

itstimwhite (commented):
Automated review (nvidia/nemotron-3-super-120b-a12b:free, confidence 2/5)

itstimwhite:
Closed to reduce main CI/merge-queue saturation. Reopen targeting `integration/loop-{domain}` for agent batch work, or rebase onto `main` and open a fresh PR. See `scripts/drain-pr-queue.sh` and `.context/loop-state.json`. (integration batch full for integration/loop-ui)

PR #10377, opened 2026-06-09
feat(observability): integrate Agnost AI analytics via OpenTelemetry

itstimwhite:
## Goal

Export Vercel AI SDK telemetry to Agnost for LLM usage analytics, with per-call `userId` and `sessionId` metadata.

## Changes

- Boot OpenTelemetry NodeSDK in `instrumentation.ts` exporting to `https://otel.agnost.ai/v1/traces`
- Add `buildAiTelemetry()` helper for consistent `experimental_telemetry` metadata
- Pass `userId`/`sessionId` on chat, title gen, insights, pitch, email classifier, and Gmail extractor calls
- Add `AGNOST_ORG_ID` and `JOVIE_ENABLE_AGNOST` env schema entries
- Unit tests for telemetry helper and enablement guards

## Testing

- `vitest run tests/unit/lib/ai/telemetry.test.ts tests/unit/lib/observability/agnost.test.ts`
- Doppler: `AGNOST_ORG_ID` set in dev/stg/prd; `JOVIE_ENABLE_AGNOST=1` in dev

---
*Generated by Claude Code*

## Summary by CodeRabbit

* **Infrastructure**
  * Added OpenTelemetry-based telemetry system for enhanced AI operation tracing and observability across chat, insights, pitch generation, and email classification.

* **Configuration**
  * Added support for Telegram webhook configuration.
  * Added new environment variables for AI services and telemetry integration.

* **Dependencies**
  * Added OpenTelemetry SDK packages for distributed tracing.
  * Updated various package versions for compatibility.

* **Tests**
  * Added test coverage for telemetry configuration and observability features.


itstimwhite:

Branch: 'feat/agnost-ai-analytics'
State: BLOCKED



To resolve:
1. git fetch origin main
2. git checkout feat/agnost-ai-analytics
3. git rebase origin/main
5. git push --force-with-lease origin feat/agnost-ai-analytics

itstimwhite:
## AI Code Review (nemotron-3-120b:free) 🟢 (confidence: 2/5)

We are given a PR that integrates Agnost AI analytics via OpenTelemetry.

We need to check for:
1. Logic bugs (race conditions, null dereferences, off-by-one, incorrect conditionals, wrong return values, silent edge-case drops)
2. Security issues (hardcoded secrets, SQL injection, XSS, auth bypass, CSP violations)
3. Cross-file regressions (does this change break callers in other files? Are new parameters handled everywhere? Do state changes propagate correctly?)
4. Convention violations (deviating from Jovie rules)
5. Missing tests (untested code paths, missing edge cases)
6. Design drift (violating DESIGN.md tokens or component hierarchy)
7. Performance (N+1 queries, large bundles, blocking async paths)

We are to ignore style preferences, variable naming (unless misleading), minor formatting, and opinion-based suggestions.

### Files changed:

1. `apps/web/app/api/chat/conversations/[id]/messages/route.ts`
2. `apps/web/app/api/chat/route.ts`
3. `apps/web/app/api/cron/generate-insights/route.ts`
4. `apps/web/app/api/insights/generate/route.ts`
5. `apps/web/app/api/webhooks/resend-inbound/route.ts`
6. `apps/web/instrumentation.ts`
7. `apps/web/lib/ai/telemetry.ts` (new)
8. `apps/web/lib/chat/run.ts`
9. `apps/web/lib/connectors/gmail/extract-event-signal.ts`
10. `apps/web/lib/env-server-schema.ts`
11. `apps/web/lib/inbox/classifier.ts`
12. `apps/web/lib/observability/agnost.ts` (new)
13. `apps/web/lib/services/insights/insight-generator.ts`
14. `apps/web/lib/services/pitch/pitch-generator.ts`
15. `apps/web/package.json`
16. `apps/web/tests/unit/lib/ai/telemetry.test.ts` (new)
17. `apps/web/tests/unit/lib/observability/agnost.test.ts` (new)
18. `pnpm-lock.yaml`

We'll focus on the logic and cross-file issues.

#### Step 1: Check for logic bugs and correctness

We'll look for:
- Incorrect conditionals
- Null dereferences
- Off-by-one
- Wrong return values
- Silent edge-case drops

Let's start with the new files and then the modified ones.

**New file: `apps/web/lib/ai/telemetry.ts`**
- This file exports a function `buildAiTelemetry` that builds the telemetry object for the Vercel AI SDK.
- It takes an `AiTelemetryOptions` object and returns a telemetry object with `isEnabled: true`, `recordInputs`, `recordOutputs`, `functionId`, and `metadata`.
- The `metadata` is built from `identity` (userId and sessionId) and additional `metadata`.
- It converts non-null values in `metadata` to strings.

Potential issues:
- The function does not check if `functionId` is provided (it is required by the type, but we should check if it's a string? The type says `readonly functionId: string;` so it's required and a string).
- The `metadata` option is of type `Record<string, string | number | boolean | null | undefined>`. We convert non-null to string. This is safe.

**New file: `apps/web/lib/observability/agnost.ts`**
- This file initializes OpenTelemetry to export traces to Agnost.
- It has a function `shouldEnableAgnost` that returns true if:
  - Not in CI
  - There is an orgId (either from `AGNOST_ORG_ID` or a default)
  - In development, only if `JOVIE_ENABLE_AGNOST` is set to '1'
  - In production, it's enabled by default (if orgId exists)
- `getAgnostOrgId` returns the orgId if enabled, else null.
- `initAgnostTelemetry` initializes the NodeSDK and starts it if not already started and if orgId exists.

Potential issues:
- The default orgId is hardcoded: `DEFAULT_AGNOST_ORG_ID = '384ebd06-d9a5-48dd-ba22-7a51e430c173'`. This is a UUID and might be acceptable for a default, but note that in production we want to use the orgId from the environment. However, the default is used only if `AGNOST_ORG_ID` is not set. This might be a problem if we don't want to send data to a default org in production without explicit configuration. But the comment says: "Set to `1` to export Agnost traces in local development." and in production it's enabled by default with the default orgId? Actually, in production, if `AGNOST_ORG_ID` is not set, it uses the default. This might be intentional for a default org, but we should check if that's acceptable.

However, note that the PR description says: "Add `AGNOST_ORG_ID` and `JOVIE_ENABLE_AGNOST` env schema entries". So we are adding these to the schema, meaning they are expected to be set. The default in the code is only a fallback.

But in the `shouldEnableAgnost` function, we have:
```typescript
  const orgId = process.env.AGNOST_ORG_ID ?? DEFAULT_AGNOST_ORG_ID;
  if (!orgId) return false;
```
So if `AGNOST_ORG_ID` is not set, we use the default. Then we check if `orgId` is truthy (which it is, because we have a default). So in production, if `AGNOST_ORG_ID` is not set, we still enable and use the default orgId.

This might be a problem because we don't want to send data to a default org without explicit configuration. However, the PR says that in Doppler we set `AGNOST_ORG_ID` in dev/stg/prd. So in production, we would have set it. The default is only for local development? Actually, the default is used in production if not set, which might be a mistake.

But note: the `shouldEnableAgnost` function returns true in production even if `AGNOST_ORG_ID` is not set (because we use the default). Then `getAgnostOrgId` returns the default. So we are sending to the default org in production by default.

This is likely not intended. We should only enable if `AGNOST_ORG_ID` is set (and not fall back to a default in production). However, the default is a UUID and might be a placeholder. We should check the design.

Looking at the test for `agnost.test.ts`:
```typescript
  it('enables export in production with default org id', async () => {
    process.env.CI = 'false';
    process.env.NODE_ENV = 'production';
    delete process.env.AGNOST_ORG_ID;
    const { shouldEnableAgnost, getAgnostOrgId } = await import(
      '@/lib/observability/agnost'
    );
    expect(shouldEnableAgnost()).toBe(true);
    expect(getAgnostOrgId()).toBe('384ebd06-d9a5-48dd-ba22-7a51e430c173');
  });
```
This test expects that in production, without `AGNOST_ORG_ID` set, we use the default and enable. So it's by design.

But is this acceptable? The PR says: "Add `AGNOST_ORG_ID` and `JOVIE_ENABLE_AGNOST` env schema entries". It doesn't say they are required. However, in the schema we made them optional. So having a default might be acceptable for local development and production fallback? But note: the default orgId is hardcoded and might be a shared org. We don't want to mix data.

However, the PR description says: "Doppler: `AGNOST_ORG_ID` set in dev/stg/prd; `JOVIE_ENABLE_AGNOST=1` in dev". So in production, we are setting `AGNOST_ORG_ID` via Doppler. Therefore, the default should never be used in production. The default is only for when the env var is not set, which in production should not happen because we set it in Doppler. But what if someone forgets? Then we would send to the default org.

This is a potential issue, but it's more of a configuration issue. The code is designed to have a fallback. We might want to change it to not enable if `AGNOST_ORG_ID` is not set in production? But the test expects the current behavior.

Given that the test is written to expect the default in production, and the PR includes that test, we must assume it's intentional. However, note that the test is new and part of the PR, so it's defining the expected behavior.

We'll leave it as is for now, but note that if the default orgId is not intended for production use, then we should change it. However, without more context, we'll assume it's acceptable.

Another issue: In `initAgnostTelemetry`, we have:
```typescript
    const sdk = new NodeSDK({
      traceExporter: new OTLPTraceExporter({
        url: AGNOST_OTEL_ENDPOINT,
        headers: { 'X-Agnost-Org-ID': orgId },
      }),
    });
```
We are setting the header `X-Agnost-Org-ID` to the orgId. This is correct.

We also start the SDK and set `agnostStarted = true` to prevent re-initialization.

**Modified files:**

Let's check the changes in the existing files for logic bugs.

**`apps/web/app/api/chat/conversations/[id]/messages/route.ts`**
- Added import: `import { buildAiTelemetry } from '@/lib/ai/telemetry';`
- In `maybeGenerateTitle` function:
  - Added `identity?: { userId?: string | null }` parameter.
  - In the call to `generateText`, we now pass `experimental_telemetry: buildAiTelemetry({ ... })` with the identity containing `userId` and `sessionId`.
  - Note: the `sessionId` is set to `conversationId` (which is the conversation id).
- In the POST handler:
  - We now destructure `clerkUserId` from `getSessionContext`.
  - Then, when calling `maybeGenerateTitle` in the `after` block, we pass `{ userId: clerkUserId }` as the identity.

Potential issues:
- The `maybeGenerateTitle` function now has an optional `identity` parameter. We are passing it in the `after` block, but what about the call from elsewhere? We must check if there are other callers.

We don't see other callers in this file, but we should check the entire codebase for `maybeGenerateTitle`. However, the function is defined in this file and only used here? Let's assume it's only used in this file for now.

But note: the function is not exported, so it's only used in this file.

- In the POST handler, we are now getting `clerkUserId` from the session. We use it in the `after` block for title generation.

This seems correct.

**`apps/web/app/api/chat/route.ts`**
- We changed the function `createGenerateReleasePitchTool` to take an additional parameter `identity: { clerkUserId: string; conversationId: string | null }`.
- Then, inside the tool, we pass that identity to the `generatePitches` call (via `buildAiTelemetry`).
- We also changed the call to `createGenerateReleasePitchTool` in `buildChatTools` to pass the identity object (with `clerkUserId` and `conversationId` from `reservedTurn`).

Potential issues:
- We are now requiring the `identity` parameter in `createGenerateReleasePitchTool`. We must check that all callers are updated.

The only caller is in `buildChatTools` (in the same file). We see:
```typescript
          generateReleasePitch: createGenerateReleasePitchTool(
            resolvedProfileId,
            {
              clerkUserId,
              conversationId: reservedTurn?.conversationId ?? null,
            }
          ),
```
This looks correct.

But note: what if `reservedTurn` is null? Then `conversationId` becomes `null`. We are allowing `conversationId: string | null` in the identity, so that's acceptable.

**`apps/web/app/api/cron/generate-insights/route.ts`**
- In `processProfile`, we now call `generateInsights` with an additional argument: `{ sessionId: profileId }`.
- We changed the function signature of `generateInsights` in `insight-generator.ts` to accept an `identity?` parameter.

Potential issues:
- We are passing `sessionId: profileId`. Is that correct? The `profileId` is the ID of the profile for which we are generating insights. It might be used as a session identifier for the insights generation process.

This seems acceptable.

**`apps/web/app/api/insights/generate/route.ts`**
- We now destructure `clerkUserId` from `getSessionContext`.
- Then, when calling `generateInsights`, we pass:
  ```typescript
        userId: clerkUserId,
        sessionId: run.id,
  ```
- Note: `run.id` is the ID of the insight run (from the database).

This seems correct.

**`apps/web/app/api/webhooks/resend-inbound/route.ts`**
- In `processInboundEmail`, we added `sessionId: threadId` to the input for the classifier.
- We changed the classifier function to accept `userId` and `sessionId` in the input.

Potential issues:
- We are passing `threadId` as the `sessionId`. This is the email thread ID, which might be appropriate for grouping events in the same thread.

**`apps/web/lib/chat/run.ts`**
- Added import: `import { buildAiTelemetry } from '@/lib/ai/telemetry';`
- In `executeChatTurn`, we changed the `experimental_telemetry` to use `buildAiTelemetry` with:
  ```typescript
      identity: {
        userId,
        sessionId: resolvedConversationId,
      },
      metadata: { model: selectedModel, plan: userPlan },
  ```
- We removed the inline telemetry object.

Potential issues:
- We are now passing `userId` and `sessionId` (the resolved conversation ID) and metadata about the model and plan.
- This seems correct.

**`apps/web/lib/connectors/gmail/extract-event-signal.ts`**
- Added import: `import { buildAiTelemetry } from '@/lib/ai/telemetry';`
- In `extractEventSignal`, we added `experimental_telemetry: buildAiTelemetry({ ... })` with:
  ```typescript
        functionId: 'jovie-gmail-event-extractor',
        identity: { userId, sessionId: inputContextDigest },
  ```
- Note: `inputContextDigest` is a string (likely a hash of the input context) used as the sessionId.

This seems acceptable.

**`apps/web/lib/env-server-schema.ts`**
- We added:
  - `TELEGRAM_BOT_TOKEN: z.string().optional(),`
  - `TELEGRAM_WEBHOOK_SECRET: z.string().optional(),`
  - `OPENAI_API_KEY: z.string().optional(),`
  - `AGNOST_ORG_ID: z.string().uuid().optional(),`
  - `JOVIE_ENABLE_AGNOST: z.enum(['0', '1']).optional(),`
- We removed `ELEVENLABS_WEBHOOK_SECRET` and `E2E_SYNTHETIC_MODE` (which were already removed in the schema and ENV_KEYS).

Potential issues:
- We are adding `AGNOST_ORG_ID` as a UUID string (optional) and `JOVIE_ENABLE_AGNOST` as an optional enum of '0' or '1'.
- This matches the usage in the agnost module.

**`apps/web/lib/inbox/classifier.ts`**
- Added import: `import { buildAiTelemetry } from '@/lib/ai/telemetry';`
- Changed the `ClassificationInput` interface to include `userId?: string | null;` and `sessionId?: string | null;`.
- In `classifyEmail`, we pass the `experimental_telemetry` built from the input's `userId` and `sessionId`.

Potential issues:
- We are now requiring the caller of `classifyEmail` to pass `userId` and `sessionId` in the input object? Actually, we made them optional in the interface, so the caller can omit them.
- We then use them in the telemetry. If they are not provided, they will be omitted from the metadata (as per `buildAiTelemetry`).

We must check the callers of `classifyEmail` to see if they are passing these fields.

From the changes, we see:
- In `apps/web/app/api/webhooks/resend-inbound/route.ts`, we are now passing `sessionId: threadId` (but note: we are not passing `userId`). We are only passing `sessionId`. The `userId` is left undefined.

Let's look at that route:
```typescript
      const classification = await classifyEmail({
        subject: fullEmail?.subject ?? null,
        bodyText: fullEmail?.text ?? null,
        artistName: artist.displayName ?? artist.username ?? 'Artist',
        artistGenres: artist.genres,
        sessionId: threadId,
      });
```
We are not passing `userId`. So in the classifier, `input.userId` will be `undefined`.

This is acceptable because we made it optional.

But note: we might want to pass the userId if available. However, in the context of an inbound email webhook, we might not have a user context? The email is being processed for an artist, and we might not know which Jovie user it belongs to at this point.

So it's acceptable.

**`apps/web/lib/services/insights/insight-generator.ts`**
- Added import: `import { type AiTelemetryIdentity, buildAiTelemetry } from '@/lib/ai/telemetry';`
- Changed the function signature of `generateInsights` to accept an optional `identity: AiTelemetryIdentity`.
- In the call to `generateObject`, we use `buildAiTelemetry` with the `identity` and metadata.

Potential issues:
- We are now passing the `identity`

---
*Automated review — please verify all suggestions before applying. This review is independent of the code generation process.*

itstimwhite (commented):
Automated review (nvidia/nemotron-3-super-120b-a12b:free, confidence 2/5)

itstimwhite:
## GStack gate evidence (JOV-2911 / PR #10377)

Coder agent drain fixes for typecheck, guardrails, and promptfoo evals.

### Local verification
- `node scripts/version-check.mjs` — passed
- `pnpm --filter @jovie/web run typecheck -- --pretty false` — passed
- `node apps/web/scripts/next-proxy-guard.mjs` — passed
- `pnpm exec promptfoo eval --filter-metadata cost=deterministic` — 193/193 passed

itstimwhite:
## AI Code Review (nemotron-3-120b:free) 🔴 (confidence: 4/5)

We are given a PR that integrates Agnost AI analytics via OpenTelemetry.

We need to check for:
1. Logic bugs (race conditions, null dereferences, off-by-one, incorrect conditionals, wrong return values, silent edge-case drops)
2. Security issues (hardcoded secrets, SQL injection, XSS, auth bypass, CSP violations)
3. Cross-file regressions (does this change break callers in other files? Are new parameters handled everywhere? Do state changes propagate correctly?)
4. Convention violations (deviating from Jovie rules)
5. Missing tests
6. Design drift
7. Performance

Note: The PR description says they are adding OpenTelemetry NodeSDK in `instrumentation.ts` and a helper `buildAiTelemetry` for consistent telemetry metadata.

We'll look for:

- In `instrumentation.ts`: They are adding a lazy import for Agnost telemetry. They catch errors and log a warning. This is safe because it doesn't block boot.

- In `lib/ai/telemetry.ts`: New file. We should check the helper function `buildAiTelemetry` for correctness.

- In various files where they are using `buildAiTelemetry`: We need to check that they are passing the correct parameters and that the function is used appropriately.

- In `lib/env-server-schema.ts`: They added new environment variables: `TELEGRAM_BOT_TOKEN`, `TELEGRAM_WEBHOOK_SECRET`, `OPENAI_API_KEY`, `ELEVENLABS_API_KEY`, `ELEVENLABS_WEBHOOK_SECRET`, `AGNOST_ORG_ID`, `JOVIE_ENABLE_AGNOST`. Also, they removed some old ones (like `ELEVENLABS_WEBHOOK_SECRET` from the old list and moved it to the new section, and removed `E2E_SYNTHETIC_MODE`). We must check that the schema is correct and that the environment variables are used appropriately.

- In `lib/observability/agnost.ts`: New file. We should check the logic for enabling Agnost and initializing the OpenTelemetry SDK.

- In the test files: We have unit tests for the telemetry helper and the agnost module.

Now, let's look for potential issues.

### Critical Issues (blockers)

We'll look for:

1. **Logic bugs**: 
   - In `instrumentation.ts`: The lazy import of Agnost telemetry is wrapped in a try-catch. If the import fails, it logs a warning and continues. This is acceptable because it's non-blocking.
   - In `lib/observability/agnost.ts`: The function `shouldEnableAgnost` returns false in CI. Then, in development, it requires `JOVIE_ENABLE_AGNOST === '1'`. In production, it returns true if `AGNOST_ORG_ID` is set. However, note that they have a default org ID: `DEFAULT_AGNOST_ORG_ID = '384ebd06-d9a5-48dd-ba22-7a51e430c173'`. This means that in production, if `AGNOST_ORG_ID` is not set, they will use the default. Is that intended? The PR says they added `AGNOST_ORG_ID` as an optional env var. But the code in `getAgnostOrgId` uses the default if the env var is not set and `shouldEnableAgnost` returns true. This might be a problem because they are using a hardcoded default org ID in production without the user setting it. However, note that `shouldEnableAgnost` in production returns true only if `orgId` (from `process.env.AGNOST_ORG_ID ?? DEFAULT_AGNOST_ORG_ID`) is truthy. Since the default is a non-empty string, it will always be true in production unless `AGNOST_ORG_ID` is set to an empty string? But the schema says it's optional and a UUID string. So if not set, it uses the default. This might be a security issue because they are sending telemetry to a default org ID without the user explicitly configuring it. However, the PR description says they set `AGNOST_ORG_ID` in Doppler for dev/stg/prd. So in production, they will have it set. But what if it's not set? Then they use the default. We should check if that is acceptable.

   However, note that the PR says: "Doppler: `AGNOST_ORG_ID` set in dev/stg/prd; `JOVIE_ENABLE_AGNOST=1` in dev". So in production, they are setting it. But the code has a fallback. This might be by design to allow a default for testing? But it's a potential issue because if someone forgets to set it in production, they still send telemetry to a default org. We should consider if this is intended.

   Alternatively, we might want to remove the default and require the env var to be set in production. But the PR doesn't do that.

   Let me see: In `shouldEnableAgnost`:
   ```typescript
   const orgId = process.env.AGNOST_ORG_ID ?? DEFAULT_AGNOST_ORG_ID;
   if (!orgId) return false;
   ```
   Since `DEFAULT_AGNOST_ORG_ID` is a non-empty string, `orgId` will never be empty. So the check `if (!orgId)` is redundant? Actually, if `process.env.AGNOST_ORG_ID` is set to an empty string, then `orgId` becomes the empty string? But the schema says it's a UUID string, so it shouldn't be empty. However, the schema allows optional, so if not set, we get the default. If set to an empty string, then `orgId` is empty string and we return false.

   But note: the schema for `AGNOST_ORG_ID` is `z.string().uuid().optional()`. So if set, it must be a UUID string (non-empty). If not set, it's undefined and we use the default.

   So in production, if the env var is not set, we use the default. This might be a problem because the default org ID is hardcoded and might belong to the Agnost team or a test org. We don't want to send production telemetry to a default org without explicit configuration.

   However, the PR says they set it in Doppler for all environments. So in practice, it will be set. But the code has a fallback. We should consider changing the code to not have a default in production? Or at least make it clear that the default is only for development?

   Looking at `shouldEnableAgnost`:
   - In CI: returns false.
   - In development: returns true only if `JOVIE_ENABLE_AGNOST === '1'` (and orgId is set, which it will be because of the default or env var).
   - In production (non-CI, non-development): returns true if orgId is set (which it always is because of the default).

   So in production, it will always be enabled if the NodeSDK imports succeed, using the default org ID if the env var is not set.

   This is a potential issue because it might send telemetry to an unintended org. We should either:
   - Remove the default and require the env var to be set (and fail if not set in production), or
   - Only use the default in development.

   However, note that the PR description says they set `AGNOST_ORG_ID` in Doppler for dev/stg/prd. So they are setting it. But the code has a fallback. We can leave it as is if the fallback is acceptable, but it's a risk.

   Alternatively, we can change the code to not use the default in production. Let's see:

   We can change `shouldEnableAgnost` to:
   ```typescript
   if (process.env.NODE_ENV === 'production' && !process.env.AGNOST_ORG_ID) {
     return false;
   }
   ```

   But the PR doesn't do that.

   Since the PR is already written and they say they set it in Doppler, we might accept it. However, as a reviewer, we must point out the risk.

   But note: the default org ID is a UUID that looks like it might be a real org. We don't know. It's hardcoded. This is a security risk because it might be sending telemetry to a third party without explicit consent in production if the env var is missing.

   However, the PR says they set it in Doppler for all environments, so in practice it will be set. But we cannot rely on that for the code.

   We should change the code to not have a default in production. But since we are reviewing, we can note it as a warning.

   However, let's check the schema: they added `AGNOST_ORG_ID` as optional. So the code must handle it being unset. The current code uses a default. We might want to change the code to not use a default and instead return false if not set in production.

   But note: the `initAgnostTelemetry` function calls `getAgnostOrgId` which returns null if not enabled. And `shouldEnableAgnost` returns false in that case. So if we remove the default, then in production without the env var, `shouldEnableAgnost` would return false because `orgId` would be undefined (or null) and then `!orgId` would be true.

   So let's change the code in `agnost.ts` to not have a default? Actually, the default is only used if the env var is not set. We can remove the default and then in `shouldEnableAgnost` we check for the env var.

   Alternatively, we can keep the default but only for non-production? The current code does not distinguish.

   Given the potential risk, I think we should change it to not use a default in production. However, the PR doesn't do that. So we must decide: is this a blocker?

   Since the PR says they set it in Doppler for all environments, and the default is only a fallback, we might consider it acceptable. But as a security reviewer, I would prefer to not have a hardcoded default that could leak telemetry.

   However, note that the default org ID is hardcoded in the code. If someone forks the code and runs it without setting the env var, they would send telemetry to that org. That might be undesirable.

   We can change the default to be only used in development? Let's see:

   In `shouldEnableAgnost`:
   ```typescript
   const orgId = process.env.AGNOST_ORG_ID;
   if (!orgId) {
     if (process.env.NODE_ENV === 'development') {
       orgId = DEFAULT_AGNOST_ORG_ID;
     } else {
       return false;
     }
   }
   ```

   But the PR doesn't do that.

   Since we are reviewing, we can note this as a warning and suggest a fix.

   However, let's see if there are any other issues.

2. **Security**:
   - Hardcoded secrets: We see a hardcoded default org ID in `agnost.ts`. As discussed, this might be a issue.
   - SQL injection: We don't see any new SQL queries.
   - XSS: We are not changing any UI rendering.
   - Auth bypass: We are not changing auth logic.
   - CSP violations: We are not changing any CSP headers.

   The hardcoded default org ID is the main security concern.

3. **Cross-file regressions**:
   - We are adding a new parameter `identity` to several functions:
     - `maybeGenerateTitle` in `apps/web/app/api/chat/conversations/[id]/messages/route.ts`
     - `createGenerateReleasePitchTool` in `apps/web/app/api/chat/route.ts`
     - `generateInsights` in `apps/web/lib/services/insights/insight-generator.ts`
     - `generatePitches` and `generatePitchDraft` in `apps/web/lib/services/pitch/pitch-generator.ts`
     - `classifyEmail` in `apps/web/lib/inbox/classifier.ts`
     - `extractEventSignal` in `apps/web/lib/connectors/gmail/extract-event-signal.ts`
   - We are also changing the callers of these functions to pass the `identity` object.

   We must check that all callers are updated correctly.

   Let's check a few:

   - In `apps/web/app/api/chat/conversations/[id]/messages/route.ts`:
     - The function `maybeGenerateTitle` now takes an optional `identity` parameter.
     - In the POST route, when calling `maybeGenerateTitle`, they pass `{ userId: clerkUserId }` (and note: `sessionId` is set to `conversationId` inside the function? Actually, no: in the function, they set `sessionId: conversationId` in the telemetry, but the function parameter `identity` only has `userId`. Wait, look:

       ```typescript
       await maybeGenerateTitle(conversationId, messagesToInsert, {
         userId: clerkUserId,
       });
       ```

       And the function is:
       ```typescript
       async function maybeGenerateTitle(
         conversationId: string,
         messages: ChatPersistenceMessage[],
         identity?: { userId?: string | null }
       ): Promise<void> {
         ...
         experimental_telemetry: buildAiTelemetry({
           functionId: 'jovie-chat-title',
           identity: {
             userId: identity?.userId,
             sessionId: conversationId,
           },
         }),
       ```

       So they are setting `sessionId` to `conversationId` inside the function, regardless of the `identity` parameter. The `identity` parameter only provides `userId`. This is acceptable because the session ID for title generation is the conversation ID.

   - In `apps/web/app/api/chat/route.ts`:
     - They changed `createGenerateReleasePitchTool` to take an extra `identity` parameter: `{ clerkUserId: string; conversationId: string | null }`
     - Then in the tool, they pass:
       ```typescript
       identity: {
         userId: identity.clerkUserId,
         sessionId: identity.conversationId,
       },
       ```
     - And in `buildChatTools`, they pass:
       ```typescript
       generateReleasePitch: createGenerateReleasePitchTool(
         resolvedProfileId,
         {
           clerkUserId,
           conversationId: reservedTurn?.conversationId ?? null,
         }
       ),
       ```
     - This looks correct.

   - In `apps/web/app/api/cron/generate-insights/route.ts`:
     - They call `generateInsights(metrics, existingTypes, { sessionId: profileId })`
     - The function `generateInsights` now takes an optional `identity` parameter (of type `AiTelemetryIdentity`).
     - They are passing `{ sessionId: profileId }` which matches.

   - In `apps/web/app/api/insights/generate/route.ts`:
     - They call `generateInsights(metrics, existingTypes, { userId: clerkUserId, sessionId: run.id })`
     - This matches.

   - In `apps/web/app/api/webhooks/resend-inbound/route.ts`:
     - They added `sessionId: threadId` to the `classifyEmail` call.
     - The function `classifyEmail` now takes an input object that has `userId` and `sessionId` (optional).
     - They are passing `sessionId: threadId` and `userId` is not passed (so undefined). This is acceptable.

   - In `apps/web/lib/chat/run.ts`:
     - They call `buildAiTelemetry` with:
       ```typescript
       identity: {
         userId,
         sessionId: resolvedConversationId,
       },
       metadata: { model: selectedModel, plan: userPlan },
       ```
     - This is correct.

   - In `apps/web/lib/connectors/gmail/extract-event-signal.ts`:
     - They call `buildAiTelemetry` with:
       ```typescript
       functionId: 'jovie-gmail-event-extractor',
       identity: { userId, sessionId: inputContextDigest },
       ```
     - This is correct.

   - In `apps/web/lib/inbox/classifier.ts`:
     - They added `userId` and `sessionId` to the `ClassificationInput` interface.
     - Then in the call to `generateObject`, they pass:
       ```typescript
       experimental_telemetry: buildAiTelemetry({
         functionId: 'jovie-email-classifier',
         identity: {
           userId: input.userId,
           sessionId: input.sessionId,
         },
       }),
       ```
     - This is correct.

   - In `apps/web/lib/services/insights/insight-generator.ts`:
     - They added an optional `identity` parameter to `generateInsights`.
     - Then in the call to `generateObject`, they pass:
       ```typescript
       experimental_telemetry: buildAiTelemetry({
         functionId: 'jovie-insight-generator',
         identity,
         metadata: { model: INSIGHT_MODEL },
       }),
       recordInputs: true,
       recordOutputs: true,
       ```
     - Note: they are spreading the `identity` and then adding `recordInputs` and `recordOutputs` after. But wait, the `buildAiTelemetry` function returns an object that includes `recordInputs` and `recordOutputs` (with defaults false). Then they are overwriting them? Actually, no: the `buildAiTelemetry` function returns an object with the properties `isEnabled`, `recordInputs`, `recordOutputs`, `functionId`, and `metadata`. Then they are adding `recordInputs: true` and `recordOutputs: true` after the call? That would be invalid because you cannot add properties to the returned object in that way.

     Let me look:

     ```typescript
     experimental_telemetry: buildAiTelemetry({
       functionId: 'jovie-insight-generator',
       identity,
       metadata: { model: INSIGHT_MODEL },
     }),
     recordInputs: true,
     recordOutputs: true,
     ```

     This is a syntax error! You cannot have trailing properties after the object. Actually, in TypeScript, this is allowed if you are using object spread? But here it's not.

     Wait, the code in the diff is:

     ```diff
     -    experimental_telemetry: {
     -      isEnabled: true,
     -      recordInputs: true,
     -      recordOutputs: true,
     +    experimental_telemetry: buildAiTelemetry({
     +      functionId: 'jovie-insight-generator',
     +      identity,
     +      metadata: { model: INSIGHT_MODEL },
     +    }),
     +    recordInputs:

---
*Automated review — please verify all suggestions before applying. This review is independent of the code generation process.*

itstimwhite (commented):
Automated review (nvidia/nemotron-3-super-120b-a12b:free, confidence 4/5)

itstimwhite:
## AI Code Review (nemotron-3-120b:free) 🟢 (confidence: 2/5)

We are given a PR that integrates Agnost AI analytics via OpenTelemetry.
 We need to review the changes for correctness, security, and adherence to Jovie conventions.

 Let's break down the changes:

 1. **instrumentation.ts**: Added lazy initialization of Agnost telemetry (OpenTelemetry) in the Node.js runtime.
    - It tries to import and initialize the Agnost telemetry, and if it fails, it logs a warning and continues.
    - This is acceptable because it's lazy and doesn't block boot.

 2. **lib/ai/telemetry.ts**: New file that provides a helper `buildAiTelemetry` to create the telemetry object for Vercel AI SDK.
    - It takes an options object with `functionId`, `identity` (userId and sessionId), and optional `metadata`.
    - It returns an object that matches the `experimental_telemetry` expected by the AI SDK.

 3. **Various AI call sites**: Updated to use `buildAiTelemetry` and pass `userId` and `sessionId` (where available) in the `identity` field.
    - Examples: chat route, title generation, insights, pitch generation, email classifier, Gmail extractor.

 4. **lib/env-server-schema.ts**: Added new environment variables:
    - `TELEGRAM_BOT_TOKEN` and `TELEGRAM_WEBHOOK_SECRET` (for Telegram webhook, unrelated to Agnost but in the same PR)
    - `OPENAI_API_KEY`, `ELEVENLABS_API_KEY`, `ELEVENLABS_WEBHOOK_SECRET` (moved from optional to required? Actually, they were added as optional)
    - `AGNOST_ORG_ID` (UUID, optional) and `JOVIE_ENABLE_AGNOST` (enum ['0','1'], optional)

 5. **Tests**: Added unit tests for the telemetry helper and the Agnost enablement logic.

 6. **package.json**: Added OpenTelemetry dependencies and updated some existing ones.

 Now, let's check for potential issues:

 **Critical Issues (Logic bugs, Security, etc.):**

 - In `instrumentation.ts`: The lazy import of Agnost telemetry is wrapped in a try-catch and logs a warning. This is safe.

 - In `lib/ai/telemetry.ts`: The function `buildAiTelemetry` converts non-null metadata values to strings. This is acceptable because the AI SDK expects metadata as strings? 
   However, note that the Vercel AI SDK's `experimental_telemetry` expects `metadata` to be a `Record<string, string>`. So converting to string is correct.

 - In `lib/observability/agnost.ts`: 
      * The function `shouldEnableAgnost` returns false in CI (good).
      * It uses a default org ID if `AGNOST_ORG_ID` is not set. This might be a problem because we don't want to send data to a default org in production without explicit configuration.
        However, note that the default org ID is a UUID string. The comment says it's the default, but we should avoid sending data to an unintended org.
        The PR description says: `AGNOST_ORG_ID` is set in Doppler for dev/stg/prd. So in production, we expect it to be set. The default is only for local development? 
        But note: in development, we require `JOVIE_ENABLE_AGNOST=1` to enable. So the default org ID is only used when:
          - Not in CI
          - `AGNOST_ORG_ID` is not set (so we use the default) AND
          - Either in production (then it's enabled by default) OR in development with `JOVIE_ENABLE_AGNOST=1`.

        This is risky because if someone forgets to set `AGNOST_ORG_ID` in production, it will send to the default org. 
        We should not have a default org ID in production. Instead, we should require the org ID to be set.

        However, looking at the code in `shouldEnableAgnost`:
          - In production (NODE_ENV !== 'development'), it returns true if `orgId` (from env or default) is set.
          - The default is set to a UUID. So if the env var is not set, we use the default and then enable in production.

        This is a potential misconfiguration risk.

        Suggestion: Remove the default and make the org ID required in production. But note the PR says it's set in Doppler.

        Alternatively, we can change the logic to:
          - In production, if `AGNOST_ORG_ID` is not set, then do not enable and log a warning.

        However, the current code does not do that. It uses the default.

        Since the PR says that `AGNOST_ORG_ID` is set in Doppler for dev/stg/prd, we assume it will be set. But to be safe, we should not rely on a default in production.

        Let's change the `getAgnostOrgId` function to return null if the org ID is not set (and we are not in a context that allows the default) but note the current code:

        We have:
          export function getAgnostOrgId(): string | null {
            if (!shouldEnableAgnost()) return null;
            return process.env.AGNOST_ORG_ID ?? DEFAULT_AGNOST_ORG_ID;
          }

        And `shouldEnableAgnost` returns true in production if there's an org ID (either from env or the default). So we are using the default in production if the env var is missing.

        We should change it so that in production, we require the env var to be set. We can do:

          export function getAgnostOrgId(): string | null {
            if (!shouldEnableAgnost()) return null;
            const orgId = process.env.AGNOST_ORG_ID;
            if (!orgId && process.env.NODE_ENV === 'production') {
              // In production, we require the org ID to be set explicitly.
              return null;
            }
            return orgId ?? DEFAULT_AGNOST_ORG_ID;
          }

        But note: the `shouldEnableAgnost` function already returns false if there's no orgId (unless we are using the default). So we are currently allowing the default in production.

        Alternatively, we can remove the default and make the org ID required. Then in `shouldEnableAgnost`:

          const orgId = process.env.AGNOST_ORG_ID;
          if (!orgId) return false;

        And then remove the default.

        However, the PR description says that the Doppler has the variable set. So we can assume it will be set in the environments where we want to enable.

        But to prevent accidental sending to the default org, let's remove the default and require the env var.

        However, note that the PR also adds a test for Agnost. We must check the test.

        Looking at the test file `apps/web/tests/unit/lib/observability/agnost.test.ts` (which we don't have the full diff for, but we can infer):

        We see in the diff that the test file was created. We should check if the test accounts for the default org ID.

        Since we don't have the full test, we must rely on the provided diff.

        The provided diff for the test file is truncated. But we can see the beginning:

        ```diff
        @@ -0,0 +1,78a4608d52 @@
        +/dev/
        ```

        This is not helpful. We must assume the test is correct.

        Given the risk, I think we should change the Agnost module to not have a default org ID and require the env var to be set.

        However, note that the PR is already written and we are reviewing. We must decide if this is a blocker.

        Considering that the PR says the env var is set in Doppler for dev/stg/prd, and that the default org ID is a specific UUID (which might be an internal Agnost org for Jovie?), 
        it might be intentional. But without knowing, we cannot be sure.

        Since the PR description does not mention a default org ID, and the code introduces one, it is a potential misconfiguration.

        Let's look at the default org ID: '384ebd06-d9a5-48dd-ba22-7a51e430c173'

        We don't know if this is a Jovie-owned org in Agnost. If it is, then it might be acceptable. But the PR does not state that.

        Given the ambiguity, I will flag this as a warning.

 - In `lib/ai/telemetry.ts`: The function `buildAiTelemetry` does not validate the `functionId` or the metadata keys. This is probably acceptable because it's just passed through.

 - In the updated AI call sites: We are passing `userId` and `sessionId` from the context. We must ensure that these are strings and not undefined or null in a way that breaks.

        The function `buildAiTelemetry` already omits null/undefined for userId and sessionId (see the test: it omits when null or undefined).

        However, note that in some places we are passing `userId: clerkUserId` which might be undefined? 

        For example, in `apps/web/app/api/chat/route.ts` in the `createGenerateReleasePitchTool` function, we pass:
          identity: {
            userId: identity.clerkUserId,
            sessionId: identity.conversationId,
          }

        Here, `identity.clerkUserId` is from the resolved profile? We must check if it can be undefined.

        Looking at the function `createGenerateReleasePitchTool`:
          It is called from `buildChatTools` with:
            generateReleasePitch: createGenerateReleasePitchTool(
              resolvedProfileId,
              {
                clerkUserId,
                conversationId: reservedTurn?.conversationId ?? null,
              }
            )

        And `clerkUserId` is from the outer scope of `buildChatTools`:
          const { clerkUserId, ... } = await getSessionContext({ ... });

        The `getSessionContext` function returns an object that includes `clerkUserId` (which is the Clerk user ID). 
        This should be a string when the user is authenticated. But note: the chat route might be called without authentication? 
        However, the chat route requires authentication (we see in the POST handler: `const { profile, clerkUserId } = await getSessionContext({ requireProfile: true });`).

        So in the chat route, we have a user. But note: the `createGenerateReleasePitchTool` is used in a tool that is only available when `resolvedProfileId` is truthy (which comes from the profile). 
        So we are safe.

        Similarly, in other places we are passing `userId` and `sessionId` that are derived from the session.

        However, note in `apps/web/lib/chat/run.ts`:
          experimental_telemetry: buildAiTelemetry({
            functionId: 'jovie-chat',
            identity: {
              userId,
              sessionId: resolvedConversationId,
            },
            metadata: { model: selectedModel, plan: userPlan },
          })

        Here, `userId` is from the function parameter (which comes from `getSessionContext` in the route) and `resolvedConversationId` is from the conversation we are chatting in.

        This should be safe.

 - In `apps/web/lib/connectors/gmail/extract-event-signal.ts`:
        We are passing:
          identity: { userId, sessionId: inputContextDigest }

        Here, `userId` is from the function parameter (which is passed from the caller) and `inputContextDigest` is a string (we hope). 
        We must check the caller of `extractEventSignal`.

        This function is called in `apps/web/app/api/webhooks/resend-inbound/route.ts`:
          const classification = await classifyEmail({
            ...,
            userId: clerkUserId,
            sessionId: threadId,
          });

        And then in `classifyEmail` (in `apps/web/lib/inbox/classifier.ts`), we pass the `userId` and `sessionId` to the AI call.

        But note: in the Gmail extractor, we are using `sessionId: inputContextDigest`. What is `inputContextDigest`? 
        We don't have the full context, but it is likely a string (maybe a hash of the email content). 
        This is acceptable as a session ID for the purpose of tracing? It might not be a real session, but it's an identifier for the email processing.

        However, the PR says: "with per-call `userId` and `sessionId` metadata". 
        The `sessionId` in the Gmail extractor is not a user session but an email context. 
        This might be acceptable because we want to trace the email processing as a unit.

        But note: the Agnost telemetry is for AI SDK spans. We are using the `sessionId` to group related AI calls (like the email classification and the event extraction) under the same session? 
        Actually, in the email webhook, we have:
          - First, we classify the email (in `classifyEmail`)
          - Then, if it's a event signal, we extract the event (in `extractEventSignal`)

        We are passing the same `sessionId` (the `threadId`) to both? 
        Actually, in the webhook route, we pass `sessionId: threadId` to `classifyEmail` and then in `classifyEmail` we use that `sessionId` for the AI call.

        Then, in the Gmail extractor, we are passing `sessionId: inputContextDigest` (which is different from `threadId`). 

        This breaks the session continuity. 

        We should use the same `sessionId` (the `threadId`) for both the classification and the extraction? 
        But note: the `inputContextDigest` is computed from the email body and might be the same for the same email? 
        However, the `threadId` is the email thread ID (from Resend) and is more stable.

        Let me look at the code in the webhook route:

          const threadId = resendEmail.headers['thread-id'] ?? resendEmail.id;

        And then we pass `sessionId: threadId` to `classifyEmail`.

        In `extractEventSignal`, we are computing:
          const inputContextDigest = crypto
            .createHash('sha256')
            .update(fullEmail?.text ?? '')
            .digest('hex');

        This is a hash of the email body. It might change if the email body changes (but in the same thread, the body of a new email is different). 
        So it is not the same as the thread ID.

        We want to trace the entire processing of an email (from webhook to classification to extraction) as one session? 
        Then we should use the `threadId` for both.

        Alternatively, we might want to have the classification and the extraction as two separate sessions? 
        The PR does not specify.

        However, note that the `sessionId` in the telemetry is meant to be the user session (or conversation session) for the AI call. 
        In the case of the email webhook, there is no user session. We are using the `threadId` as a proxy for the email thread.

        Using the hash of the body for the extraction step does not seem to add value and breaks the continuity.

        Suggestion: Use the `threadId` (passed as `sessionId` to `classifyEmail`) also in the Gmail extractor.

        But note: the `extractEventSignal` function does not currently receive the `threadId`. 
        We would have to change the function signature to accept it.

        However, the PR does not show that change. We see in the webhook route:

          const classification = await classifyEmail({
            ...,
            userId: clerkUserId,
            sessionId: threadId,
          });

          if (classification) {
            const signal = await extractEventSignal({
              ...,
              sessionId: threadId,   // <--- we are not doing this currently
            });
          }

        Actually, in the webhook route, we are passing `sessionId: threadId` to `classifyEmail` but not to `extractEventSignal`. 
        We are passing `sessionId: threadId` only to `classifyEmail`.

        In the `extractEventSignal` function, we are currently using:
          sessionId: inputContextDigest

        We should change that to use the `threadId` that was passed in from the webhook route.

        But note: the `extractEventSignal` function is called from the webhook route, and we have the `threadId` there.

        However, the PR diff for the webhook route only shows:

          async function processInboundEmail(
            ...
          ) {
            ...
            const classification = await classifyEmail({
              ...
              sessionId: threadId,
            });

            if (classification) {
              const signal = await extractEventSignal({
                ...
                sessionId: threadId,   // We need to add this line
              });
            }
          }

        But the diff we see for the webhook route is:

          @@ -113,6 +113,7 @@ async function processInboundEmail(
                   bodyText: fullEmail?.text ?? null,
                   artistName: artist.displayName ?? artist.username ?? 'Artist',
                   artistGenres: artist.genres,
+                  sessionId: threadId,
                );

        This is inside the `classifyEmail` call. We are not seeing the `extractEventSignal` call.

        We must look at the full diff for the webhook route. The provided diff only shows a snippet.

        Since we don't have the full diff, we cannot be sure. But note: the PR description says we are passing `userId`/`sessionId` on chat, title gen, insights, pitch, email classifier, and Gmail extractor calls.

        So we expect the Gmail extractor call to have the `sessionId` passed.

        Given the information we have, we must assume that the webhook route passes the `sessionId` to both. 
        But the diff we see only shows the addition in the `classifyEmail` call.

        This is a potential issue: the Gmail extractor might not be getting the `sessionId` from the webhook route.

        However, note that the `extractEventSignal` function now has a parameter for `sessionId` in the telemetry? 
        Actually, no: the function signature of `extractEventSignal` has not changed in the diff we see. 
        We see:

          export

---
*Automated review — please verify all suggestions before applying. This review is independent of the code generation process.*

itstimwhite (commented):
Automated review (nvidia/nemotron-3-super-120b-a12b:free, confidence 2/5)

itstimwhite:
## AI Code Review (nemotron-3-120b:free) 🟢 (confidence: 2/5)

We are given a PR that integrates Agnost AI analytics via OpenTelemetry.

We need to review the changes for:
1. Logic bugs (race conditions, null dereferences, off-by-one, incorrect conditionals, wrong return values, silent edge-case drops)
2. Security (hardcoded secrets, SQL injection, XSS, auth bypass, CSP violations)
3. Cross-file regressions (does this change break callers in other files? Are new parameters handled everywhere? Do state changes propagate correctly?)
4. Convention violations (deviating from Jovie rules)
5. Missing tests (untested code paths, missing edge cases)
6. Design drift (violating DESIGN.md tokens or component hierarchy)
7. Performance (N+1 queries, large bundles, blocking async paths)

We are to ignore style preferences, variable naming (unless misleading), minor formatting, and opinion-based suggestions.

Note: The PR description says it adds OpenTelemetry NodeSDK in `instrumentation.ts` exporting to `https://otel.agnost.ai/v1/traces`, adds a helper `buildAiTelemetry`, passes `userId`/`sessionId` on various calls, adds env schema entries, and unit tests.

We'll check for:

- Logic bugs: e.g., incorrect conditionals, null dereferences, etc.
- Security: e.g., hardcoded secrets (we see a default org ID in agnost.ts, but it's optional and overridden by env, so not a secret), SQL injection (none seen), XSS (not relevant in backend), auth bypass (none seen).
- Cross-file regressions: We must check that the changes to function signatures (adding `identity` or `userId`/`sessionId` parameters) are propagated to all callers.
- Convention violations: 
   - Migration files are immutable: none changed.
   - Single DB driver: we see `import { db } from '@/lib/db'` is used, but we don't see any new direct DB access that violates this.
   - One Clerk instance per environment, proxy via fetch(): we see `getSessionContext` used, which is the standard way.
   - No direct middleware.ts creation: none.
   - No emoji in UI: we are not changing UI files.
   - No native browser dialogs: none.
   - No decorative hover motion: none.
   - Marketing/blog/legal pages must be fully static: none of these are changed.
   - Global UI singletons render once (root layout only): none.
   - CSP domains stay in sync with provider registry: we are adding a new endpoint for Otel export, but note that the CSP is set in the Next.js headers or middleware? We don't see changes to CSP in this PR. However, note that the PR adds a new external endpoint: `https://otel.agnost.ai/v1/traces`. We must ensure that this domain is allowed in the CSP. But note: the CSP is set in the Next.js config (next.config.js) or via headers. We don't see that in this PR. However, the PR does not change the CSP. Since we are adding a new outbound connection, we must check if the CSP allows it. But note: the CSP for `connect-src` might need to include `https://otel.agnost.ai`. However, the PR does not show changes to next.config.js or headers. This might be a cross-file regression if the CSP is not updated. But note: the PR description does not mention CSP, and we are not given the next.config.js. We must rely on the fact that the PR might have missed updating the CSP. However, the instructions say we are to check for cross-file regressions. We should note that if the CSP is not updated, then the Otel export might be blocked in production. But note: the Otel export is done from the Node.js server (not the browser) so CSP (which is for the browser) does not apply. Therefore, we don't need to worry about CSP for the Otel export because it's a server-to-server connection. The CSP is only for the browser. So this is safe.

   However, note that the PR also adds a new dependency: `@opentelemetry/exporter-trace-otlp-proto` and `@opentelemetry/sdk-node`. These are server-side only, so no CSP impact.

- Missing tests: We see two new test files: `telemetry.test.ts` and `agnost.test.ts`. We should check if they are adequate.

- Design drift: We are adding new files and changing existing ones. We must check if they follow the existing patterns.

- Performance: We are adding OpenTelemetry which might add overhead, but it's optional and only enabled when the env vars are set. We are using lazy import in `instrumentation.ts` so it doesn't affect Edge bundles. The telemetry helper is a simple object build, so negligible.

Now, let's check each file:

1. `apps/web/app/api/chat/conversations/[id]/messages/route.ts`:
   - Added import of `buildAiTelemetry`.
   - Changed `maybeGenerateTitle` to accept an optional `identity` parameter (with `userId?` and `sessionId?`).
   - In the call to `generateText` for title generation, replaced the hardcoded telemetry object with `buildAiTelemetry` and passed the identity (with `userId` from the parameter and `sessionId` as `conversationId`).
   - In the POST route, we now destructure `clerkUserId` from `getSessionContext` and pass it to `maybeGenerateTitle` as `{ userId: clerkUserId }`.
   - This looks correct. We are passing the userId and sessionId (conversationId) for title generation.

2. `apps/web/app/api/chat/route.ts`:
   - Changed `createGenerateReleasePitchTool` to accept an additional `identity` parameter (of type `{ clerkUserId: string; conversationId: string | null }`).
   - Inside the tool, when calling `generatePitches`, we pass the `identity` (with `userId` and `sessionId`).
   - In `buildChatTools`, we now pass `{ clerkUserId, conversationId: reservedTurn?.conversationId ?? null }` as the `identity` to `createGenerateReleasePitchTool`.
   - This looks correct.

3. `apps/web/app/api/cron/generate-insights/route.ts`:
   - In `processProfile`, when calling `generateInsights`, we now pass `{ sessionId: profileId }` as the second argument (which is the `identity`).
   - Note: the `generateInsights` function now expects a second argument `identity?: AiTelemetryIdentity`. We are passing an object with `sessionId`. This is correct.

4. `apps/web/app/api/insights/generate/route.ts`:
   - In the POST route, we now destructure `clerkUserId` from `getSessionContext`.
   - When calling `generateInsights`, we pass `{ userId: clerkUserId, sessionId: run.id }`.
   - This is correct.

5. `apps/web/app/api/webhooks/resend-inbound/route.ts`:
   - In `processInboundEmail`, when calling `classifyEmail`, we now pass an additional argument `sessionId: threadId`.
   - Note: the `classifyEmail` function now expects an input object that has `userId?` and `sessionId?` (see changes in `classifier.ts` below). We are only passing `sessionId`. This is acceptable because `userId` is optional.

6. `apps/web/instrumentation.ts`:
   - Added a lazy import of `initAgnostTelemetry` from `@/lib/observability/agnost` and called it inside a try-catch, logging a warning on failure.
   - This is correct and follows the pattern for Braintrust.

7. `apps/web/lib/ai/telemetry.ts` (new file):
   - Defines `AiTelemetryIdentity` and `AiTelemetryOptions`.
   - Exports `buildAiTelemetry` function that builds the telemetry object for Vercel AI SDK.
   - It takes `functionId`, optional `identity` (with `userId` and `sessionId`), optional `metadata`, and optional `recordInputs` and `recordOutputs`.
   - It returns an object with `isEnabled: true`, the given `recordInputs` and `recordOutputs` (or false if not provided), the `functionId`, and a `metadata` object that includes `userId` and `sessionId` (if present) and any additional metadata (converted to string).
   - This looks correct.

8. `apps/web/lib/chat/run.ts`:
   - Added import of `buildAiTelemetry`.
   - In `executeChatTurn`, replaced the hardcoded telemetry object with `buildAiTelemetry` and passed `identity: { userId, sessionId: resolvedConversationId }` and `metadata: { model: selectedModel, plan: userPlan }`.
   - This is correct.

9. `apps/web/lib/connectors/gmail/extract-event-signal.ts`:
   - Added import of `buildAiTelemetry`.
   - In `extractEventSignal`, when calling `generateObject`, we pass `experimental_telemetry: buildAiTelemetry({ functionId: 'jovie-gmail-event-extractor', identity: { userId, sessionId: inputContextDigest } })`.
   - This is correct.

10. `apps/web/lib/env-server-schema.ts`:
    - Added `TELEGRAM_BOT_TOKEN` and `TELEGRAM_WEBHOOK_SECRET` (optional strings) - note: the PR description didn't mention Telegram, but the CodeRabbit summary did. We see it's added here and in the ENV_KEYS.
    - Added `OPENAI_API_KEY` and `ELEVENLABS_API_KEY` and `ELEVENLABS_WEBHOOK_SECRET` (optional strings).
    - Added `AGNOST_ORG_ID` (optional string, UUID format) and `JOVIE_ENABLE_AGNOST` (optional, either '0' or '1').
    - Removed `ELEVENLABS_WEBHOOK_SECRET` from the list? Actually, we see it was removed from the ENV_KEYS list? Let's check:
        - In the diff, we see:
          - Removed: `ELEVENLABS_WEBHOOK_SECRET` (from the ENV_KEYS list) but then added back later? Actually, looking at the diff:
            - In the ServerEnvSchema, we added `ELEVENLABS_WEBHOOK_SECRET` (as optional) in the new section for ElevenLabs.
            - Then in the ENV_KEYS array, we removed the old `ELEVENLABS_WEBHOOK_SECRET` (which was in a different place) and then added the new one? Actually, the diff shows:
                - Removed: `'ELEVENLABS_WEBHOOK_SECRET',` (from the middle of the ENV_KEYS array) and then added:
                - `'OPENAI_API_KEY',`
                - `'ELEVENLABS_API_KEY',`
                - `'ELEVENLABS_WEBHOOK_SECRET',`
            So it's just moved. This is acceptable.
    - Also removed `E2E_SYNTHETIC_MODE` from the schema and ENV_KEYS.
    - This looks correct.

11. `apps/web/lib/inbox/classifier.ts`:
    - Added import of `buildAiTelemetry`.
    - Changed the `ClassificationInput` interface to include `userId?` and `sessionId?`.
    - In `classifyEmail`, when calling `generateObject`, we pass `experimental_telemetry: buildAiTelemetry({ functionId: 'jovie-email-classifier', identity: { userId: input.userId, sessionId: input.sessionId } })`.
    - This is correct.

12. `apps/web/lib/observability/agnost.ts` (new file):
    - Defines `AGNOST_OTEL_ENDPOINT` and a default org ID (which is a UUID string).
    - `shouldEnableAgnost`: returns false in CI, then checks for org ID, then in development checks for `JOVIE_ENABLE_AGNOST === '1'`, otherwise returns true in production.
    - `getAgnostOrgId`: returns the org ID if enabled, else null.
    - `initAgnostTelemetry`: idempotent, tries to initialize the NodeSDK with an OTLP trace exporter to the Agnost endpoint with the org ID header.
    - This looks correct. However, note that the default org ID is hardcoded. But note: it is only used if `AGNOST_ORG_ID` is not set. And the function `shouldEnableAgnost` returns false if there's no org ID (unless we are in development and `JOVIE_ENABLE_AGNOST` is set? Actually, no: in development, we return true only if `JOVIE_ENABLE_AGNOST === '1'` and we have an org ID (either from env or the default). So if `AGNOST_ORG_ID` is not set, we use the default. This is acceptable for development but note: the default org ID might be a shared one. However, the PR says it's optional and we can override it. This is acceptable.

    - One potential issue: the `initAgnostTelemetry` function catches errors and logs a warning, but does not rethrow. This is acceptable because we don't want to crash the app on telemetry failure.

13. `apps/web/lib/services/insights/insight-generator.ts`:
    - Added import of `type AiTelemetryIdentity` and `buildAiTelemetry`.
    - Changed `generateInsights` to accept an optional `identity` parameter (of type `AiTelemetryIdentity`).
    - In the call to `generateObject`, we pass `experimental_telemetry: buildAiTelemetry({ functionId: 'jovie-insight-generator', identity, metadata: { model: INSIGHT_MODEL }, recordInputs: true, recordOutputs: true })`.
    - This is correct.

14. `apps/web/lib/services/pitch/pitch-generator.ts`:
    - Added import of `type AiTelemetryIdentity` and `buildAiTelemetry`.
    - Changed `generatePitches` to accept an optional `identity` parameter.
    - In the call to `generateObject`, we pass `experimental_telemetry: buildAiTelemetry({ functionId: 'jovie-pitch-generator', identity, metadata: { model: PITCH_MODEL }, recordInputs: true, recordOutputs: true })`.
    - Changed `generatePitchDraft` to accept an optional `identity` parameter (in the params object).
    - In the call to `generateObject`, we pass `experimental_telemetry: buildAiTelemetry({ functionId: 'jovie-pitch-draft-generator', identity, metadata: { model: PITCH_MODEL, target: destination.target, platform: destination.platform ?? 'none' }, recordInputs: true, recordOutputs: true })`.
    - This is correct.

15. `apps/web/package.json`:
    - Added dependencies: `@opentelemetry/exporter-trace-otlp-proto`, `@opentelemetry/sdk-node`, `@opentelemetry/sdk-trace-base`.
    - Updated several other dependencies (date-fns, react-error-boundary, stripe, eslint, vite, etc.) and removed some (upstash/redis version downgraded? Actually, it's updated from 1.38.0 to 1.37.0, which is a downgrade, but that's acceptable if it's for compatibility).
    - This is acceptable.

16. `apps/web/tests/eval/promptfoo/jovie-chat-provider.ts`:
    - This is a test file for promptfoo. We see changes to the `evaluateChatTitleContract` function:
        - It now reads the `telemetry.ts` file.
        - It changes the check for redacting telemetry inputs and outputs to look for `buildAiTelemetry` and the functionId and sessionId.
        - It also checks the telemetry source for the recordInputs and recordOutputs lines.
        - It changes the check for scheduling title generation to look for the new call pattern (with the identity object).
    - This looks correct and updated to reflect the changes.

17. `apps/web/tests/unit/lib/ai/telemetry.test.ts` (new file):
    - Tests `buildAiTelemetry`:
        - Test 1: checks that userId and sessionId are mapped into metadata, along with additional metadata.
        - Test 2: checks that null identity fields are omitted (so metadata is empty).
    - This is adequate.

18. `apps/web/tests/unit/lib/observability/agnost.test.ts` (new file):
    - We are not given the full diff, but we can assume it tests the agnost module.
    - We should check if it tests the `shouldEnableAgnost` and `getAgnostOrgId` functions and the init function (maybe mocking the imports).
    - Since we don't have the full content, we rely on the fact that it was added and the PR author wrote tests.

**Logic bugs:**

- In `apps/web/lib/observability/agnost.ts`:
    - The function `shouldEnableAgnost`:
        ```ts
        if (process.env.CI === 'true') return false;
        const orgId = process.env.AGNOST_ORG_ID ?? DEFAULT_AGNOST_ORG_ID;
        if (!orgId) return false;
        if (process.env.NODE_ENV === 'development') {
          return process.env.JOVIE_ENABLE_AGNOST === '1';
        }
        return true;
        ```
      This means:
        - In CI: always false.
        - If no org ID (and not using the default? Actually, note: we set `orgId` to either the env or the default. So if the env is not set, we use the default. Then we check `if (!orgId) return false;` -> but the default is set, so `orgId` is never null or undefined? Actually, the default is a non-empty string. So this condition will never be true? 
          - Wait: `DEFAULT_AGNOST_ORG_ID` is a non-empty string, so `orgId` is always at least that string. Therefore, the condition `if (!orgId) return false;` is never true.
        - Then, in development, we return true only if `JOVIE_ENABLE_AGNOST === '1'`.
        - In production, we return true (if we passed the CI check and the orgId check, which we always do because orgId is set to the default if not provided).

      This means that in production,

---
*Automated review — please verify all suggestions before applying. This review is independent of the code generation process.*

itstimwhite (commented):
Automated review (nvidia/nemotron-3-super-120b-a12b:free, confidence 2/5)

itstimwhite:

Branch: 'feat/agnost-ai-analytics'
State: BLOCKED



To resolve:
1. git fetch origin main
2. git checkout feat/agnost-ai-analytics
3. git rebase origin/main
5. git push --force-with-lease origin feat/agnost-ai-analytics

itstimwhite:
Closed to reduce main CI/merge-queue saturation. Reopen targeting `integration/loop-{domain}` for agent batch work, or rebase onto `main` and open a fresh PR. See `scripts/drain-pr-queue.sh` and `.context/loop-state.json`. (integration batch full for integration/loop-ui)

PR #10486, opened 2026-06-10
fix(onboarding): hide Turnstile chrome until verification is required (JOV-2925)

itstimwhite:
BTW queue recovery.


itstimwhite:
Retarget to integration/loop-auth

PR #10710, opened 2026-06-11
train(ui): loop-ui integration through v26.6.42

itstimwhite:
## Summary
- carry the slash-picker transcript-clearance fix from integration/loop-ui to main
- bump release metadata to 26.6.42

## Included PRs
- #10708 fix(chat): keep slash picker clear of transcript

## Checks
- PASS: integration fast-path local verify from scripts/loop-integration-ship.sh
- PASS: pre-push hook on feature branch and integration sync push: biome check . && pnpm --filter=@jovie/web run pre-push
- LOCAL LIMITATION: focused Playwright E2E reruns for tests/e2e/shell-chat-v1.spec.ts:347 timed out locally before reaching the slash-picker assertion after slow local Next compilation.

## Coordination
- GBrain handoff: handoffs/jovie-neutral-actions-supabase-coordination
- GBrain backend: Supabase/Postgres


itstimwhite:

Branch: 'integration/loop-ui'
State: BLOCKED



To resolve:
1. git fetch origin main
2. git checkout integration/loop-ui
3. git rebase origin/main
5. git push --force-with-lease origin integration/loop-ui

itstimwhite:

### Logic Errors

**1. `JovieChat.tsx` — `useLayoutEffect` dependency on `input` causes unnecessary scroll resets**

The `useLayoutEffect` that manages picker clearance scrolling now includes `input` in its dependency array. Every keystroke in the composer will tear down and re-setup the entire effect — removing and re-adding `resize` and `visualViewport` listeners, disconnecting and recreating the `ResizeObserver`, and scheduling new `requestAnimationFrame` pairs. This is a performance regression: the scroll-to-bottom behavior only needs to run once when the picker opens (i.e., when `shouldReservePickerClearance` transitions from `false` to `true`), not on every character typed.

**Why it matters:** Repeatedly creating/destroying `ResizeObserver` instances and scheduling animation frames on every keystroke can cause jank and memory pressure in long chat sessions. It also means the scroll-to-bottom fires unnecessarily while the user is just typing with the picker already open.

**Suggested fix:** Remove `input` from the dependency array. If the intent was to re-scroll when new messages arrive while the picker is open, that should be driven by a separate effect keyed on message count or a dedicated callback from the message-loading logic — not the raw `input` state.

---

**2. `JovieChat.tsx` — `scrollThreadToBottom` callback is recreated on every render**

`scrollThreadToBottom` is defined with `useCallback(..., [scrollContainerRef])`, but `scrollContainerRef` is a ref object (stable reference), so the callback is stable. However, it's then used as a dependency in the `useLayoutEffect` below. Since `scrollThreadToBottom` is stable, this is not a bug per se — but it's misleading and adds noise. More importantly, if `scrollContainerRef` were ever replaced with a state-backed ref, this would compound issue #1 above.

**Why it matters:** Low severity now, but it makes the dependency array harder to reason about and could mask future regressions.

**Suggested fix:** Either inline the scroll logic inside the effect, or hoist `scrollThreadToBottom` outside the component if it doesn't depend on reactive state. At minimum, document why it's a `useCallback`.

---

### Security

**None found.**

No hardcoded secrets, injection vectors, auth changes, or CSP modifications. The change is purely presentational (CSS clearance value and scroll behavior).

---

### Data Integrity

**None found.**

No DB queries, migrations, or data mutations in this diff.

---

### Error Handling

**3. `JovieChat.tsx` — `ResizeObserver` usage has no try/catch**

The `ResizeObserver` constructor is guarded by a `typeof ResizeObserver !== 'undefined'` check, which handles the missing-API case. However, `resizeObserver?.observe(scrollContainer)` and `resizeObserver?.disconnect()` are called without any error boundary. In some browser versions, `ResizeObserver` construction or `observe()` can throw (e.g., if the element is detached from the DOM between the check and the call).

**Why it matters:** Low likelihood, but an unhandled `ResizeObserver` error would crash the component tree in React's `useLayoutEffect`, potentially leaving the chat view in a broken state.

**Suggested fix:** Wrap the `ResizeObserver` creation and `observe()` call in a try/catch, and silently fall back to just the window resize listeners on failure.

---

### Cross-File Impact

**None found.**

The version bumps across all `package.json` files and `version.json` are consistent (all `26.6.41` → `26.6.42`). The CHANGELOG entry is additive and correctly ordered. No API surface changes that would affect consumers of `@jovie/ui`, `@jovie/auth-routing`, or `@jovie/extension-contracts`.

---

### Summary

|---|---|---|
| Logic errors | 2 | 1 high (unnecessary effect re-runs on keystroke), 1 low |
| Security | 0 | — |
| Data integrity | 0 | — |
| Error handling | 1 | low |
| Cross-file impact | 0 | — |

**Primary concern:** The `input` dependency in the picker-clearance `useLayoutEffect` is the only issue that could cause a noticeable performance regression in production. Everything else is minor.

itstimwhite (commented):

PR #10714, opened 2026-06-11
perf(chat): reduce turn latency N+1 and gate fan-out (JOV-3012)

itstimwhite:
Main-target queue drain (btw Phase 0). Reopened against the recommended integration branch — full CI runs on the domain train PR only. Policy: `.claude/rules/ci-branching.md`, state: `.context/loop-state.json`.

Integration fast lane only. Merge via `scripts/loop-integration-ship.sh integration/loop-chat tim/jov-3012-chat-latency` after local verify.

### DIFF
--- a/apps/web/app/api/chat/route.ts
+++ b/apps/web/app/api/chat/route.ts
@@ -36,10 +36,9 @@ import {
   createUIMessageStream,
   createUIMessageStreamResponse,
   tool,
-  type UIMessage,
   type UIMessageChunk,
 } from 'ai';
-import { and, count, desc, sql as drizzleSql, eq } from 'drizzle-orm';
+import { and, count, sql as drizzleSql, eq } from 'drizzle-orm';
 import { NextResponse } from 'next/server';
 import { z } from 'zod';
 import { tryHandleAnonymousOnboardingChat } from '@/app/api/chat/onboarding-handler';
@@ -56,10 +55,12 @@ import {
   detectAlbumArtGenerationIntent,
   resolveAlbumArtCapability,
 } from '@/lib/chat/album-art-capability';
+import { extractLastUserText } from '@/lib/chat/message-text';
 import {
   updateOwnedReleaseGeneratedPitches,
   updateOwnedReleaseMetadata,
 } from '@/lib/chat/release-writes';
+import { fetchReleasesForChat } from '@/lib/chat/releases';
 import {
   extractUIMessageText,
   parseChatRequestBody,
@@ -98,14 +99,14 @@ import {
 import { db } from '@/lib/db';
 import { clickEvents, tips } from '@/lib/db/schema/analytics';
 import { users } from '@/lib/db/schema/auth';
-import { discogReleases } from '@/lib/db/schema/content';
+
 import { socialLinks } from '@/lib/db/schema/links';
 import { creatorProfiles } from '@/lib/db/schema/profiles';
 import { sqlAny } from '@/lib/db/sql-helpers';
 import { upsertRelease } from '@/lib/discography/queries';
 import { generateUniqueSlug } from '@/lib/discography/slug';
 import { FEATURE_FLAGS } from '@/lib/feature-flags/shared';
-import { checkGateForUser } from '@/lib/flags/server';
+import { checkGatesForUser } from '@/lib/flags/server';
 import { createAuthenticatedCorsHeaders } from '@/lib/http/headers';
 import {
   classifyIntent,
@@ -152,7 +153,6 @@ import type {
 } from '@/lib/services/album-art/types';
 import {
   buildCanvasMetadata,
-  getCanvasStatusFromMetadata,
   summarizeCanvasStatus,
 } from '@/lib/services/canvas/service';
 import { getInsightsSummary } from '@/lib/services/insights/lifecycle';
@@ -165,7 +165,6 @@ import {
   resolvePitchDestination,
 } from '@/lib/services/pitch';
 import { DSP_PLATFORMS } from '@/lib/services/social-links/types';
-import { toISOStringOrNull } from '@/lib/utils/date';
 import { detectPlatform } from '@/lib/utils/platform-detection/detector';
 
 export const dynamic = 'force-dynamic';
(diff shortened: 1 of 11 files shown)


itstimwhite:

### Logic Errors

**1. `checkGatesForUser` is sequential, not parallel — defeats the purpose**
- **File:** `apps/web/lib/flags/statsig.ts` (new function, ~line 163)
- **Why it matters:** The PR description explicitly calls out "serial Statsig gates" as a problem being solved, and the test in `chat-latency.test.ts` asserts "checks kill-switch gates in parallel." But `checkGatesForUser` uses `Promise.all` over `gates.map(… => checkGateForUser(…))`, which fires all gates concurrently — that part is fine. However, each individual `checkGateForUser` call still awaits `initializeStatsig()` sequentially inside the Promise.all. This is actually correct behavior (Promise.all runs them concurrently). **No bug here**, but the caching layer introduces a subtle issue: if two gates are requested in the same batch and the first gate's result is cached but the second isn't, they'll both call `initializeStatsig()` — which is fine since it's idempotent. **Verdict: No logic bug, but see caching concern below.**

**2. Gate cache is a module-level `Map` with no bounds — memory leak**
- **File:** `apps/web/lib/flags/statsig.ts` (~line 24)
- **Why it matters:** `gateCache` is an unbounded `Map` that grows by one entry per unique `(userId, gateKey)` pair per 10-second TTL window. In a high-traffic app with many users, this grows without limit. Entries are only evicted lazily on read (when expired) or never (if the user never requests again). Over hours/days on a long-running server, this is a memory leak.
- **Suggested fix:** Use a `LRUMap` with a max size (e.g., 10,000 entries) or use a proper TTL cache library. At minimum, add a periodic `setInterval` cleanup or size-based eviction.

**3. `extractLastUserText` in `route.ts` is called before `userText` is used in `tryRouteViaIntent`, but `userText` is computed from `uiMessages` which may be empty after the refactor**
- **File:** `apps/web/app/api/chat/route.ts` (~line 2573, the call site)
- **Why it matters:** The diff shows `userText` is now passed to `tryRouteViaIntent` instead of `uiMessages`. The `userText` variable is computed earlier via `extractLastUserText(uiMessages)`. If `uiMessages` is empty, `userText` will be `''`, and `tryRouteViaIntent` returns `null` early. This is correct. **No bug.**

**4. `releases` may be `undefined` when passed to `buildChatTools`**
- **File:** `apps/web/app/api/chat/route.ts` (~line 2642)
- **Why it matters:** The `releases` variable is conditionally fetched (only when `resolvedProfileId` is truthy, around line 2630-ish based on context). If `resolvedProfileId` is null, `releases` could be undefined, but `buildChatTools` now expects `releases: ReleaseContext[]`. The tool factories that receive `releases` (e.g., `createCheckCanvasStatusTool`) don't have a null guard on the `releases` parameter — they rely on it being an array. If `releases` is undefined, `releases.length` and `findReleaseByTitle(releases, ...)` will throw.
- **Suggested fix:** Ensure `releases` defaults to `[]` before being passed to `buildChatTools`. Check the assignment site — if it's `const releases = resolvedProfileId ? await fetchReleasesForChat(resolvedProfileId) : [];` then it's fine. But the diff doesn't show the full assignment, only the usage.

**5. `createWorldClassBioTool` still has a try/catch for `fetchReleasesForChat` that was removed**
- **File:** `apps/web/app/api/chat/route.ts` (~line 1671-1676 in the diff)
- **Why it matters:** The diff shows the old code had a try/catch block that called `fetchReleasesForChat` inside the `execute` function. The new code removes this entirely and passes `releases` as a parameter instead. The diff correctly removes the try/catch and the local `releases` variable. **No bug — this is correct.**

### Security

**1. No hardcoded secrets found.** All gate keys use constants from `CHAT_KILL_SWITCH_GATES`.

**2. No SQL injection.** All DB queries use Drizzle ORM parameterized queries.

**3. No auth bypass.** The `profileId` and `userId` are still resolved from Clerk auth before any tool execution.

**4. Gate cache could serve stale values across different server instances**
- **File:** `apps/web/lib/flags/statsig.ts`
- **Why it matters:** The cache is in-memory per process. In a multi-instance deployment, one instance might have a cached `chatDisabled = false` while another instance has `chatDisabled = true` after a kill-switch is flipped. This is a consistency issue, not a security vulnerability per se, but it means the kill switch won't propagate instantly to all users. The 10-second TTL mitigates this. **Low severity, acceptable for a kill switch.**

### Data Integrity

**1. `capPersistedToolOutput` uses `extractSummary` which is not imported**
- **File:** `apps/web/lib/chat/tool-events.ts` (~line 383)
- **Why it matters:** The new code calls `extractSummary(output)` but this function is not imported or defined in the visible diff. If it's defined elsewhere in the file (outside the diff), this is fine. If it's missing, this will be a runtime error.
- **Suggested fix:** Verify `extractSummary` is defined or imported in `tool-events.ts`.

**2. `PERSISTED_TOOL_OUTPUT_MAX_BYTES` is 16KB — reasonable but no test for the boundary**
- **File:** `apps/web/lib/chat/tool-events.ts`
- **Why it matters:** The test creates an oversized output and verifies it gets capped, but doesn't test the exact boundary (output exactly at 16KB). This is a minor gap, not a bug.

### Error Handling

**1. `measureSerializedBytes` returns `Number.POSITIVE_INFINITY` on serialization failure**
- **File:** `apps/web/lib/chat/tool-events.ts` (~line 371)
- **Why it matters:** If `JSON.stringify` throws (e.g., circular reference), the function returns `Infinity`, which is `> PERSISTED_TOOL_OUTPUT_MAX_BYTES`, so the output will be capped. The fallback summary will be `'Tool output truncated for storage.'` since `extractSummary` and `output.message` may also fail. This is acceptable defensive behavior. **No bug.**

**2. `checkGatesForUser` doesn't catch individual gate failures**
- **File:** `apps/web/lib/flags/statsig.ts` (~line 163)
- **Why it matters:** If one gate check throws, `Promise.all` will reject the entire batch, and the caller in `route.ts` will get an unhandled rejection. The old code had two separate `await` calls, each implicitly wrapped in the route's error handling. Now, if `checkGateForUser` throws for one gate, the other gate's result is lost too.
- **Suggested fix:** Wrap each gate check in a try/catch that returns the default value, or add a `.catch(() => defaultValue)` inside the `Promise.all` map.

### Cross-File Impact

**1. `fetchReleasesForChat` moved from inline in `route.ts` to `lib/chat/releases.ts`** — The test in `releases.test.ts` verifies the import path changed correctly. The promptfoo eval test was also updated. **No broken callers detected.**

**2. `extractLastUserText` moved from inline in `route.ts` to `lib/chat/message-text.ts`** — Both `route.ts` and `run.ts` now import from the new module. **No broken callers.**

**3. `checkGatesForUser` is a new export from `lib/flags/server.ts`** — No other callers in the changed files. **No impact.**

---

### Summary of Issues Found

| 1 | **Medium** | `lib/flags/statsig.ts` | Unbounded `gateCache` Map — memory leak in long-running processes |
| 2 | **Medium** | `lib/flags/statsig.ts` | `checkGatesForUser` — single gate failure rejects entire batch via `Promise.all` |
| 3 | **Medium** | `app/api/chat/route.ts` | `releases` variable may be undefined when `resolvedProfileId` is null — needs default to `[]` |
| 4 | **Low** | `lib/chat/tool-events.ts` | `extractSummary` called but not visible in diff — verify it's defined/imported |

**No security vulnerabilities, no logic-critical bugs, no data integrity issues found.** The main concerns are the unbounded cache and the `Promise.all` error propagation in the new batch gate checker.

itstimwhite (commented):

PR #10744, opened 2026-06-11
fix(ios): neutralize auth loading spinner

itstimwhite:
## Summary
- Stacks on `integration/loop-chat` so the iOS/System B cleanup rides the integration train instead of targeting `main` directly.
- Changes the Continue in Browser loading spinner from `JovieColor.accentBlue` to the neutral enabled foreground token.
- Adds an iOS Swift Testing guard so the central auth action cannot drift back to accent-blue loading chrome.
- Bumps release manifests to 26.6.47.

## Verification
- `node scripts/version-check.mjs`
- `git diff --check origin/integration/loop-chat..HEAD`
- `apps/ios/scripts/ensure-configuration.sh`
- `CODE_SIGNING_ALLOWED=NO JOVIE_IOS_RESET_SIMULATOR=0 apps/ios/scripts/run-xcodebuild.sh test -only-testing:JovieTests/AuthScreenStyleGuardTests` passed before the integration-train rebase: 1 Swift Testing test passed, `TEST SUCCEEDED`.
- Post-rebase rerun compiled the changed app/test targets, then the simulator launch failed with `NSMachErrorDomain Code=-308` before test execution.
- Push pre-push gate passed: repo Biome, proxy guard, Tailwind guard, web typecheck, web lint.

## Risk
Low. One SwiftUI tint token changes inside an already-white primary auth button. The guard is source-scoped to `ContinueInBrowserButton`, so status/progress accent usage elsewhere remains allowed.


itstimwhite:

### Logic Errors

**1. `AuthScreenStyleGuardTests.swift:13` — Fragile source-range boundary selection**

```swift
let buttonEnd = try #require(source.range(of: "private struct AuthErrorText: View"))
```

The test slices the source between `ContinueInBrowserButton` and `AuthErrorText` to scope its assertion. This is a reasonable structural heuristic, but it silently becomes a no-op if `AuthErrorText` is ever renamed, moved above `ContinueInBrowserButton`, or removed. In that case `#require` will trap and the test will fail with an opaque message rather than a clear "button not found" diagnostic.

**Why it matters:** A rename of `AuthErrorText` (unrelated to the feature) would break this test with a misleading failure. More critically, if someone reorders the file and `AuthErrorText` lands *before* `ContinueInBrowserButton`, `buttonStart.lowerBound..<buttonEnd.lowerBound` would be an invalid range (crash at runtime) rather than a graceful test failure.

**Suggested fix:** Add a comment documenting the ordering assumption, and consider using a more stable delimiter (e.g., the closing brace of `ContinueInBrowserButton` matched by brace-depth, or a dedicated `// BEGIN ContinueInBrowserButton` / `// END ContinueInBrowserButton` marker comment). At minimum, guard with a precondition that `buttonStart.lowerBound < buttonEnd.lowerBound`.

---

**2. `AuthScreenStyleGuardTests.swift:14` — Substring direction is correct but fragile**

```swift
let buttonSource = source[buttonStart.lowerBound..<buttonEnd.lowerBound]
```

This works because `range(of:)` returns the range of the *first* occurrence and `ContinueInBrowserButton` appears before `AuthErrorText` in the file. However, if a *different* struct named `ContinueInBrowserButton` were ever added above the real one (e.g., a preview or a copy), the test would silently assert against the wrong block.

**Why it matters:** Low probability, but the test would pass while the actual button still uses `accentBlue`.

**Suggested fix:** Anchor the start search more precisely — e.g., search for `"private struct ContinueInBrowserButton: View {"` with the full signature, or search from the end of the file backward to get the last match.

---

### Security

None found. The change is a tint color swap in a SwiftUI view and a source-level test guard. No network, auth, data, or input-handling code is touched.

---

### Data Integrity

None found. No DB queries, migrations, or data transformations are involved.

---

### Error Handling

**3. `AuthScreenStyleGuardTests.swift:11` — `#require` on file read gives no context on failure**

```swift
let source = try String(contentsOf: sourceURL, encoding: .utf8)
```

If the file can't be read (permissions, path drift after a rename), the error will be a generic Swift `Error` with no indication of which path was attempted. This is a test file so the blast radius is small, but it makes CI debugging harder.

**Why it matters:** In CI, a file-read failure would surface as a test crash with no path information, slowing diagnosis.

**Suggested fix:** Wrap in a `do/catch` that throws a descriptive message including `sourceURL.path`, or use `#require` with a message.

---

### Cross-File Impact

**4. `AuthScreen.swift:413` — `JovieColor.backgroundBase` may not be semantically correct for a spinner tint**

The diff changes the tint from `JovieColor.accentBlue` to `JovieColor.backgroundBase`. The PR description says this is the "neutral enabled foreground token." If `backgroundBase` is literally a *background* color (as the name implies), using it as a spinner tint on a button could produce poor contrast — a light-colored spinner on a light button surface, or a dark spinner on a dark surface depending on the theme.

**Why it matters:** The intent is "neutral foreground," but the token name suggests "background." If the naming is accurate, this is a contrast regression. If the naming is misleading, the code is confusing for the next developer.

**Suggested fix:** Verify that `backgroundBase` is indeed the correct neutral foreground token in the design system. If it's actually a background color, either (a) introduce or use a dedicated neutral foreground token (e.g., `JovieColor.foregroundNeutral` or `JovieColor.onPrimary`), or (b) add a code comment explaining why `backgroundBase` is the right choice here.

---

### Summary

|---|---|---|
| Logic errors | 2 | Low–Medium |
| Security | 0 | — |
| Data integrity | 0 | — |
| Error handling | 1 | Low |
| Cross-file impact | 1 | Medium (potential contrast regression) |

The most actionable item is **#4** — confirm `backgroundBase` is the right semantic token for a spinner tint, not a background fill. Items **#1** and **#2** are fragility concerns in the test harness that are worth hardening before this pattern is replicated elsewhere.

itstimwhite (commented):

itstimwhite:
Closed to reduce main CI/merge-queue saturation. Reopen targeting `integration/loop-{domain}` for agent batch work, or rebase onto `main` and open a fresh PR. See `scripts/drain-pr-queue.sh`. (queue drain — integration/system-b-4 must ship via domain train, not main).

PR #10756, opened 2026-06-11
train(ui): loop-ui review blockers

itstimwhite:
## Summary
- trains the loop-ui integration branch containing #10755 into main
- adds missing neutral Tailwind mappings for calendar focus/hover utilities
- preserves DisplayMenu and calendar layout-stability guard coverage

## Verification
- #10755 merged into integration/loop-ui
- local focused Vitest, Biome, Tailwind guard, typecheck, and pre-push gate passed on the feature branch
- integration/loop-ui is one commit ahead of main

PR #10758, opened 2026-06-11
fix(ci): retain opentelemetry trace peer for knip

itstimwhite:
## Summary
- Keep `@opentelemetry/sdk-trace-base` declared for the web app while marking it as an intentional Knip ignored dependency.
- Restores the `Knip (dead code)` check that failed on `main` after #10756.

## Test plan
- `pnpm knip`
- `pnpm biome check knip.json`
- `git diff --check`
- pre-push gate: `biome check . && pnpm --filter=@jovie/web run pre-push`

## Risk
- Config-only CI fix; no runtime code or dependency graph change.

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -44,6 +44,7 @@
         "@radix-ui/react-toggle",
         "@radix-ui/react-toggle-group",
         "@radix-ui/react-tooltip",
+        "@opentelemetry/sdk-trace-base",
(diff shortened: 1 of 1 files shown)


itstimwhite:
Closed: CI fix ships via integration/loop-infra train #10704, not direct to main.

PR #10822, opened 2026-06-13
fix(ops): add Turnstile hostname automation for jov.ie production

itstimwhite:
## Goal

Fix production `/start` Turnstile failures (Cloudflare client error `110200` — domain not authorized) by adding repeatable hostname management for the active production widget.

## Context

- Doppler `prd` already has `NEXT_PUBLIC_TURNSTILE_SITE_KEY` + `TURNSTILE_SECRET_KEY`
- Production `/start` still fails when `jov.ie` / `www.jov.ie` are missing from the Cloudflare Turnstile widget Hostname Management allowlist
- Wrangler OAuth lacks Turnstile Sites scopes; automation requires a dedicated `CLOUDFLARE_API_TOKEN`

## Changes

- Add `scripts/turnstile-config.ts` (list/show/ensure-hostnames with dry-run + `--allow-prod` guard)
- Add `docs/TURNSTILE_SETUP.md` operator runbook
- Document `CLOUDFLARE_API_TOKEN` in `.env.example` + `docs/DOPPLER_SETUP.md`
- Add unit tests for hostname merge logic
- Extend synthetic monitoring to fail on `verification failed (110200)`
- Log operator follow-up in `AUTH_FIX_LOG.md`

## Post-merge operator action (required to complete JOV-2329)

1. Create `CLOUDFLARE_API_TOKEN` with **Turnstile Sites Read + Write** for account `fd8a6fcfa0ff0b708dbae942af18c49b`
2. Store in Doppler `jovie-web/prd`
3. Preview: `doppler run --project jovie-web --config prd -- pnpm tsx scripts/turnstile-config.ts ensure-hostnames --dry-run`
4. Apply: `doppler run --project jovie-web --config prd -- pnpm tsx scripts/turnstile-config.ts ensure-hostnames --yes --allow-prod`
5. Verify `https://jov.ie/start` no longer shows `110200`

## Testing

- `pnpm --filter=@jovie/web exec vitest run tests/unit/scripts/turnstile-config-script.test.ts`
- `pnpm biome check scripts/turnstile-config.ts docs/TURNSTILE_SETUP.md`
- Live Playwright probe confirmed Turnstile challenge loads on `/start` after first message; hostname allowlist update still required via Cloudflare API

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -173,6 +173,9 @@ DEMO_VIDEO_URL=
 # Cloudflare Turnstile keys for onboarding chat bot protection.
 NEXT_PUBLIC_TURNSTILE_SITE_KEY=
 TURNSTILE_SECRET_KEY=
+# Cloudflare Turnstile widget management (scripts/turnstile-config.ts)
+CLOUDFLARE_API_TOKEN=
+CLOUDFLARE_ACCOUNT_ID=
 # Production synthetic signup canary. Configure only in Doppler prd.
 E2E_SYNTHETIC_MODE=
 E2E_PROD_SIGNUP_EMAIL_BASE=
--- a/AUTH_FIX_LOG.md
+++ b/AUTH_FIX_LOG.md
@@ -488,6 +488,41 @@ Final passing evidence:
 - No production Clerk settings changed.
 - No Clerk dashboard changes made for this auth-smoke stabilization.
 
+## JOV-2329 — Production Turnstile hostname allowlist for jov.ie - 2026-06-12
+
+Failing production evidence inspected:
+
+- Linear JOV-2329: `/start` on `https://jov.ie/start` rendered Turnstile client
+  error `110200` (domain not authorized) even though Doppler `prd` already had
+  `NEXT_PUBLIC_TURNSTILE_SITE_KEY` and `TURNSTILE_SECRET_KEY`.
+- Live Playwright probe on 2026-06-12: sending the first onboarding message surfaces
+  the Turnstile security panel and loads the Cloudflare challenge for site key
+  `0x4AAAAAADM7PX2W5Au4ZIkV`, but anonymous `/api/chat` still returns
+  `TURNSTILE_REQUIRED` until a client token is minted.
+
+Root cause:
+
+- The active Cloudflare Turnstile widget did not include `jov.ie` / `www.jov.ie`
+  in Hostname Management. Cloudflare returns client error `110200` when the page
+  hostname is outside the widget allowlist.
+
+Fix made in repo:
+
+- Added `scripts/turnstile-config.ts` to inspect/update widget hostnames via the
+  Cloudflare API (dry-run by default).
+- Documented operator flow in `docs/TURNSTILE_SETUP.md`.
+- Added synthetic regression coverage for `verification failed (110200)`.
+
+Post-merge operator action still required once per environment:
+
+1. Create/store `CLOUDFLARE_API_TOKEN` (Turnstile Sites Read + Write) in Doppler
+   `jovie-web/prd`.
+2. Preview:
+   `doppler run --project jovie-web --config prd -- pnpm tsx scripts/turnstile-config.ts ensure-hostnames --dry-run`
+3. Apply:
+   `doppler run --project jovie-web --config prd -- pnpm tsx scripts/turnstile-config.ts ensure-hostnames --yes --allow-prod`
+4. Re-check `https://jov.ie/start` and production synthetic `/start` health.
+
 ## Post-Merge Required Smoke and Mobile Guard Closeout - 2026-06-02
 
 Failing CI evidence inspected:
--- a/apps/web/tests/e2e/synthetic-auth-ui.spec.ts
+++ b/apps/web/tests/e2e/synthetic-auth-ui.spec.ts
@@ -34,6 +34,7 @@ const FRONT_DOOR_CONFIG_ERRORS = [
   'auth unavailable',
   'clerk is not configured',
   'turnstile is not configured',
+  'verification failed (110200)',
 ] as const;
 
 const EMAIL_IDENTIFIER_INPUT_SELECTOR =
--- a/apps/web/tests/e2e/synthetic-golden-path.spec.ts
+++ b/apps/web/tests/e2e/synthetic-golden-path.spec.ts
@@ -20,6 +20,7 @@ const HOMEPAGE_PRIMARY_CTA_TEST_ID = 'homepage-primary-cta';
 const SIGNUP_PATH = '/signup';
 const START_PATH = '/start';
 const TURNSTILE_CONFIG_ERROR = 'turnstile is not configured';
+const TURNSTILE_DOMAIN_ERROR = 'verification failed (110200)';
 
 test.describe('Synthetic Monitoring - Golden Path', () => {
   // Only run in synthetic monitoring mode
@@ -72,6 +73,7 @@ test.describe('Synthetic Monitoring - Golden Path', () => {
         'text="Internal Server Error"',
         'text="Something went wrong"',
         `text="${TURNSTILE_CONFIG_ERROR}"`,
+        `text="${TURNSTILE_DOMAIN_ERROR}"`,
         '[data-testid="error-boundary"]',
       ];
 
--- a/apps/web/tests/unit/scripts/turnstile-config-script.test.ts
+++ b/apps/web/tests/unit/scripts/turnstile-config-script.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from 'vitest';
+import {
+  mergeHostnameAllowlist,
+  missingHostnames,
+  PRODUCTION_TURNSTILE_HOSTNAMES,
+  resolveHostnameTarget,
+  STAGING_TURNSTILE_HOSTNAMES,
+} from '../../../../../scripts/turnstile-config';
+
+describe('scripts/turnstile-config.ts', () => {
+  it('defines production hostnames required for jov.ie signup', () => {
+    expect(PRODUCTION_TURNSTILE_HOSTNAMES).toEqual(['jov.ie', 'www.jov.ie']);
+  });
+
+  it('defines staging hostnames for pre-prod verification', () => {
+    expect(STAGING_TURNSTILE_HOSTNAMES).toEqual([
+      'staging.jov.ie',
+      'main.jov.ie',
+    ]);
+  });
+
+  it('resolves hostname targets explicitly', () => {
+    expect(resolveHostnameTarget('prod')).toEqual(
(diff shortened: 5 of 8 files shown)


itstimwhite:

Branch: 'tim/jov-2329-turnstile-jovie-hostnames'
State: BLOCKED



To resolve:
1. git fetch origin main
2. git checkout tim/jov-2329-turnstile-jovie-hostnames
3. git rebase origin/main
5. git push --force-with-lease origin tim/jov-2329-turnstile-jovie-hostnames

itstimwhite:
## Code Review — PR #10822

---

#### 1. `isMain` guard is broken for `tsx` execution

**File:** `scripts/turnstile-config.ts:449-450`

```ts
const isMain =
  import.meta.url === pathToFileURL(process.argv[1] ?? '').href ||
  process.argv[1]?.endsWith('turnstile-config.ts');
```

**Problem:** When run via `pnpm tsx scripts/turnstile-config.ts`, `process.argv[1]` is the *bare* path `scripts/turnstile-config.ts` (no leading `/`, no `file://`). `pathToFileURL('scripts/turnstile-config.ts').href` produces `file:///…/scripts/turnstile-config.ts` (absolute). The `===` comparison will almost never match. The `endsWith` fallback works for direct `tsx` invocation but **fails** when the script is run via `doppler run -- … pnpm tsx scripts/turnstile-config.ts` because `process.argv[1]` may be resolved to an absolute path like `/Users/…/scripts/turnstile-config.ts` — which still ends with the filename, so this specific case is fine. However, the `===` branch is dead code and misleading. More critically, if the file is ever symlinked or the path contains query params (some runners), `endsWith` also breaks.

**Why it matters:** If `isMain` evaluates `false`, the script silently does nothing when executed directly. This is a developer-facing ops tool, so silent failure during a production incident is bad.

**Suggested fix:** Drop the `===` branch. Use only the `endsWith` check, or better, use `import.meta.url` consistently:

```ts
const isMain = import.meta.url.endsWith('scripts/turnstile-config.ts');
```

---

#### 2. `normalizeWidget` silently produces empty `sitekey` on unexpected payloads

**File:** `scripts/turnstile-config.ts:199-205`

```ts
sitekey:
  typeof raw.sitekey === 'string'
    ? raw.sitekey
    : typeof raw.id === 'string'
      ? raw.id
      : '',
```

**Problem:** If the Cloudflare API changes its response shape or returns an unexpected key, `sitekey` becomes `''`. Downstream code (`getTurnstileWidget`, `updateTurnstileWidgetDomains`) uses `widget.sitekey` in URL paths like `/accounts/${accountId}/challenges/widgets/${widget.sitekey}`. An empty sitekey produces a malformed API call that either 404s or hits an unintended endpoint.

**Why it matters:** Silent degradation. An operator sees a confusing Cloudflare 404 instead of a clear "unexpected API response" error.

**Suggested fix:** Throw explicitly when neither `sitekey` nor `id` is present:

```ts
const sitekey = typeof raw.sitekey === 'string' ? raw.sitekey
  : typeof raw.id === 'string' ? raw.id
  : null;
if (!sitekey) throw new Error('Cloudflare API returned widget without sitekey or id');
```

---

#### 3. `cloudflareRequest` does not bound retries — violates codebase convention

**File:** `scripts/turnstile-config.ts:155-178`

The codebase convention states: *"Server-side external HTTP must be bounded (timeout + retry)."* The `AbortSignal.timeout(15_000)` provides a timeout, but there is **no retry logic**. A transient Cloudflare 5xx or network blip causes the entire script to fail with no retry.

**Why it matters:** This is an ops automation script. Running it manually is fine, but if it's ever incorporated into CI or a retry-sensitive context, unbounded single-attempt HTTP is fragile.

**Suggested fix:** Add a simple retry wrapper (e.g., 2 retries with exponential backoff for 5xx / network errors, not for 4xx). At minimum, document that the caller is responsible for retrying.

---

#### 4. `updateTurnstileWidgetDomains` sends mutable widget fields that may overwrite Cloudflare-managed state

**File:** `scripts/turnstile-config.ts:230-249`

```ts
const body = {
  domains: [...domains],
  ...(widget.name ? { name: widget.name } : {}),
  ...(widget.mode ? { mode: widget.mode } : {}),
  ...(widget.clearance_level ? { clearance_level: widget.clearance_level } : {}),
};
```

**Problem:** `normalizeWidget` reads `mode` and `clearance_level` from the GET response. If Cloudflare normalizes or transforms these values (e.g., `mode` comes back as `"invisible"` but the API only accepts `"managed"`), the PUT will fail or silently change widget behavior. The script is only supposed to manage hostnames but is round-tripping all fields.

**Why it matters:** Accidental widget mode change from `"invisible"` to `"managed"` (or vice versa) could break the UX on `/start`.

**Suggested fix:** Only send `domains` in the PUT body. If other fields need to be preserved, document why explicitly:

```ts
const body = { domains: [...domains] };
```

---

#### 5. `cmdEnsureHostnames` exits with code 2 on dry-run even when hostnames are already present

**File:** `scripts/turnstile-config.ts:282-286`

```ts
if (result.added.length === 0) {
  console.error('✅ Required hostnames are already present on the widget.');
  return;
}
```

This path correctly returns with exit code 0. But the dry-run path at lines 288-293:

```ts
if (options.dryRun || !options.yes) {
  console.error(
    `DRY RUN: would add hostnames: ${result.added.join(', ')}. ` +
      'Re-run with --yes to apply.'
  );
  process.exitCode = 2;
  return;
}
```

**Problem:** Exit code 2 on a dry-run preview is surprising. Dry-run is the **default** mode. An operator running `--dry-run` to preview changes gets a non-zero exit code, which CI or shell scripts will interpret as failure. This is arguably intentional (to signal "changes needed"), but it's inconsistent with the `--help` output which presents `--dry-run` as a safe preview.

**Why it matters:** If this script is ever wired into a CI check, every PR pipeline will fail until hostnames are added, even though the script is working correctly.

**Suggested fix:** Either document the exit code semantics in `--help`, or use exit code 0 for dry-run-with-changes and reserve 2 for actual errors.

---

#### 6. `logAudit` writes to `stderr` — mixing audit logs with CLI output

**File:** `scripts/turnstile-config.ts:82-86`

```ts
function logAudit(action: string, details: Record<string, unknown>): void {
  const ts = new Date().toISOString();
  console.error(
    `[turnstile-config-audit ${ts}] ${action} ${JSON.stringify(details)}`
  );
}
```

**Problem:** `cmdEnsureHostnames` writes its JSON result to `stdout` (via `console.log`), but the audit trail goes to `stderr`. If an operator redirects `stdout` to a file (`> result.json`), the audit log is lost. Conversely, if they capture `stderr` for alerting, they get mixed audit + CLI error messages.

**Why it matters:** Audit logs should be distinguishable from operational output. This is minor for a CLI tool but creates confusion in incident review.

**Suggested fix:** Acceptable as-is for a developer tool. Flagging for awareness only.

---

#### 7. `missingHostnames` does not normalize required hostnames

**File:** `scripts/turnstile-config.ts:117-124`

```ts
export function missingHostnames(
  existingDomains: readonly string[],
  requiredHostnames: readonly string[]
): string[] {
  const existing = new Set(
    existingDomains.map(domain => domain.trim().toLowerCase())
  );
  return requiredHostnames.filter(
    hostname => !existing.has(hostname.toLowerCase())
  );
}
```

**Problem:** `existingDomains` are normalized to lowercase, but `requiredHostnames` are only lowercased during the `has()` check — the **returned** array preserves the original casing of `requiredHostnames`. This is fine for the current constants (all lowercase), but if someone adds a mixed-case hostname to `PRODUCTION_TURNSTILE_HOSTNAMES`, the "added" output will show the mixed-case version while the actual Cloudflare API call uses the lowercased version from `mergeHostnameAllowlist`. This creates a confusing discrepancy in the audit log.

**Why it matters:** Minor inconsistency in operator-facing output.

**Suggested fix:** Normalize required hostnames at the boundary:

```ts
const normalizedRequired = requiredHostnames.map(h => h.trim().toLowerCase());
return normalizedRequired.filter(hostname => !existing.has(hostname));
```

---

#### 8. No validation of `target` CLI argument

**File:** `scripts/turnstile-config.ts:371`

```ts
options.target = argv[++i] as TurnstileHostnameTarget;
```

**Problem:** The `as TurnstileHostnameTarget` cast is a compile-time assertion only. If a user passes `--target production` (or any invalid string), it silently flows through. `resolveHostnameTarget` falls through to the `default` case and returns production hostnames, which may not be what the user intended.

**Why it matters:** A typo like `--target prod` (instead of the documented but non-existent value — actually `prod` is valid, but `--target production` is not) silently applies production changes when the user meant staging.

**Suggested fix:** Validate the argument before casting:

```ts
const validTargets = new Set(['prod', 'staging', 'all']);
if (!validTargets.has(argv[i + 1])) {
  throw new Error(`Invalid target "${argv[i + 1]}". Must be one of: prod, staging, all`);
}
```

---

#### 9. `CLOUDFLARE_ACCOUNT_ID` documented as optional in `.env.example` but required by `getAccountId`

**File:** `.env.example:176` vs `scripts/turnstile-config.ts:103-110`

`.env.example` lists `CLOUDFLARE_ACCOUNT_ID=` without a comment marking it required. The code throws if neither `CLOUDFLARE_ACCOUNT_ID` nor `R2_ACCOUNT_ID` is set. If `R2_ACCOUNT_ID` is already in Doppler (which it is for Jovie), this works. But a new environment without `R2_ACCOUNT_ID` will fail with an unhelpful error.

**Why it matters:** Onboarding a new environment (e.g., a staging Doppler config) will hit this wall.

**Suggested fix:** Add a comment in `.env.example`:

```
CLOUDFLARE_ACCOUNT_ID=  # required for Turnstile automation; defaults to R2_ACCOUNT_ID
```

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | **Medium** | `isMain` guard has a dead `===` branch; fragile for symlinked/resolved paths |
| 2 | **Medium** | `normalizeWidget` silently produces empty `sitekey` on unexpected API response |
| 3 | **Low** | No retry logic on external HTTP calls (violates codebase convention) |
| 4 | **Medium** | PUT body round-trips `mode`/`clearance_level` — could unintentionally change widget behavior |
| 5 | **Low** | Exit code 2 on dry-run may confuse CI/shell scripts |
| 6 | **Info** | Audit logs on `stderr` mixed with CLI output |
| 7 | **Low** | `missingHostnames` returns non-normalized casing |
| 8 | **Medium** | No runtime validation of `--target` CLI argument; invalid values silently default to `prod` |
| 9 | **Low** | `CLOUDFLARE_ACCOUNT_ID` not clearly documented as required in `.env.example` |

**Highest priority to address:** #4 (unintended widget mutation) and #8 (silent prod targeting on typo). Both could cause production impact if an operator makes a mistake.

itstimwhite (commented):

PR #10850, opened 2026-06-13
v26.6.52 fix(desktop): stop PWA offline fallback from breaking auth callbacks (JOV-3119)

itstimwhite:
## Summary

Fixes P0 desktop login failure where auth callback navigations were served the PWA offline fallback ("You're offline") instead of completing OAuth.

## Root cause

A stale service worker from prior web/PWA sessions could control Electron navigations. The SW's network-first navigation handler fell back to `offline.html` on fetch failure/timeout for auth callback routes.

## Changes

- **electron-runtime-init.js**: unregister all service workers when `runtime=electron` or `JovieDesktop` UA is detected (runs before React hydrates)
- **usePWAInstall.ts**: explicitly unregister SW in Electron runtime (belt-and-suspenders)
- **sw.js**: bypass offline fallback for auth/OAuth routes; bump cache to `jovie-offline-v3`
- **apps/desktop/main.ts**: set `JovieDesktop` user agent on auth handoff window
- **Tests**: auth navigation path policy, electron-runtime-init SW cleanup, electron PWA hook behavior

## Validation

- `vitest run tests/unit/desktop/electron-runtime-init.test.ts tests/unit/desktop/electron-runtime.test.tsx tests/unit/service-worker/auth-navigation-paths.test.ts` — 10/10 passed
- `pnpm run typecheck --filter @jovie/web --filter @jovie/desktop` — passed
- Pre-commit + pre-push hooks — passed

## Rollback

Revert PR; desktop users may need one restart to clear any in-memory SW control. No schema or data migrations.

Linear: https://linear.app/jovie/issue/JOV-3119


itstimwhite:
## Review: PR #10850 — v26.6.52 fix(desktop): stop PWA offline fallback from breaking auth callbacks (JOV-3119)

---

**1. Logic Errors**

**Issue 1: `electron-runtime-init.js` — SW unregistration is fire-and-forget with no ordering guarantee**

- **File:** `apps/web/public/electron-runtime-init.js` (new lines ~14–24)
- **Description:** The `navigator.serviceWorker.getRegistrations().then(...)` call is not awaited and its returned promise is not chained to anything. The `void registration.unregister()` inside the loop means individual unregistration failures are silently swallowed, but more critically, the entire `getRegistrations()` promise is detached. If `getRegistrations()` itself rejects (e.g., in a broken SW state), the `.catch(() => {})` handles it, but there is no guarantee the unregistrations complete before the page proceeds to load. Since this script runs synchronously in `<head>` and React hydration follows, a race exists where the SW could still intercept the first navigation.
- **Why it matters:** The entire fix depends on the SW being gone before auth callback navigations fire. A race here means the fix is probabilistic, not deterministic.
- **Suggestion:** This is inherently limited by the browser's SW lifecycle — you can't synchronously block on SW unregistration from a `<script>` tag. However, you should at minimum `await` the full chain so the microtask queue drains before the script finishes executing. Change the pattern to an async IIFE:

```js
(async () => {
  try {
    if ('serviceWorker' in navigator) {
      const registrations = await navigator.serviceWorker.getRegistrations();
      await Promise.allSettled(registrations.map(r => r.unregister()));
    }
  } catch {}
})();
```

Using `Promise.allSettled` ensures all unregistrations are attempted even if one fails, and awaiting the IIFE means the script's execution context doesn't complete until unregistration is done.

---

**Issue 2: `sw.js` — Auth path bypass uses plain `fetch()` with no timeout or error handling**

- **File:** `apps/web/public/sw.js` (new lines ~62–67)
- **Description:** For auth navigation paths, the SW does `event.respondWith(fetch(event.request))` with no timeout, no retry, and no error handling. The non-auth path has an 8-second `AbortController` timeout. If the network is slow or the auth endpoint hangs, the auth fetch will hang indefinitely in the SW, which is worse than the original problem (at least the timeout had a fallback).
- **Why it matters:** A hung auth fetch in the SW will cause the browser tab to spin forever with no feedback to the user. The original code had a timeout that fell back to `offline.html`; now there's no timeout at all for the most critical navigation paths.
- **Suggestion:** Apply the same `AbortController` timeout to auth path fetches, and on timeout/failure, let the browser handle the navigation naturally (don't call `event.respondWith` at all, or respond with a `Response.error()`):

```js
if (isAuthNavigationPath(requestUrl.pathname)) {
  const controller = new AbortController();
  const timeoutId = setTimeout(() => controller.abort(), NAV_TIMEOUT_MS);
  event.respondWith(
    fetch(event.request, { signal: controller.signal })
      .finally(() => clearTimeout(timeoutId))
      .catch(() => fetch(event.request)) // one retry without signal, or fall through
  );
  return;
}
```

Or simpler: don't intercept at all for auth paths by returning early without calling `event.respondWith`, letting the browser handle it natively.

---

**Issue 3: `usePWAInstall.ts` — `unregisterServiceWorker()` called at module top-level with no guard against double-execution**

- **File:** `apps/web/hooks/usePWAInstall.ts` (new lines ~29–31)
- **Description:** The `unregisterServiceWorker().catch(() => {})` call runs at module evaluation time (top-level) when `isElectronRuntime()` is true. This is fine for a singleton module, but if this module is ever code-split or evaluated multiple times (e.g., HMR, React strict mode double-effects in dev), it could fire multiple unregistration calls. The `electron-runtime-init.js` also unregisters, so there's a double-unregistration happening — which is harmless but indicates the belt-and-suspenders approach lacks coordination.
- **Why it matters:** Not a bug per se, but the PR description says "belt-and-suspenders" — if the init script already runs in `<head>` before this module loads, the hook-level call is redundant. If the init script fails silently (as it can, given the try/catch), the hook call is the backup. This should be documented with a comment explaining which is primary and which is backup.
- **Suggestion:** Add a comment clarifying the redundancy is intentional and which path is primary. Consider gating the hook-level call on a flag set by the init script to avoid double work.

---

**2. Security**

**Issue 4: `main.ts` — User agent string injection without sanitization of `DESKTOP_USER_AGENT_PRODUCT`**

- **File:** `apps/desktop/src/main.ts` (new line ~656)
- **Description:** The code does `` `${authHandoffWindow.webContents.getUserAgent()} ${DESKTOP_USER_AGENT_PRODUCT}` ``. If `DESKTOP_USER_AGENT_PRODUCT` is not a static constant but could ever be influenced by config or environment, this is a user-agent injection vector. Even if it's currently a static string, the pattern is fragile.
- **Why it matters:** Low severity since this is Electron's internal `setUserAgent` (not HTTP header injection in a server context), but a malformed UA string could cause parsing issues in Clerk/OAuth providers that do UA-based detection.
- **Suggestion:** Verify `DESKTOP_USER_AGENT_PRODUCT` is a hardcoded constant. If so, no action needed — but add a comment confirming it's static.

**Issue 5: `sw.js` — Auth path list is duplicated between TS source and SW, with no build-time enforcement**

- **File:** `apps/web/public/sw.js` and `apps/web/lib/service-worker/auth-navigation-paths.ts`
- **Description:** The `AUTH_NAVIGATION_PATH_PREFIXES` array is defined in two places. The test file checks string containment (`expect(swSource).toContain(\`'${prefix}'\`)`), which is a weak check — it doesn't verify the array structure, the function logic, or that the list is complete. A developer could add a prefix to the TS file, forget to update `sw.js`, and the test would still pass if the old prefix list happens to contain a matching substring.
- **Why it matters:** If the lists diverge, auth routes could get the offline fallback in production while tests pass. This is the exact class of bug this PR is trying to fix.
- **Suggestion:** Generate the SW file from the TS source at build time (e.g., a build script that writes the array into `sw.js`), or import the TS module into a SW build step. At minimum, strengthen the test to parse the array from `sw.js` and compare element-by-element.

---

**3. Data Integrity**

None found. No database or data-layer changes in this PR.

---

**4. Error Handling**

**Issue 6: `electron-runtime-init.js` — Outer try/catch swallows all errors including potential null dereferences**

- **File:** `apps/web/public/electron-runtime-init.js` (the outer `try/catch` block)
- **Description:** The entire runtime init logic is wrapped in a `try/catch` with an empty catch block. If `document.documentElement` is null (theoretically impossible in a browser, but the pattern is still risky), or if `navigator.serviceWorker` exists but `getRegistrations` throws synchronously (unlikely but spec-allowed), all errors are silently swallowed with no logging.
- **Why it matters:** This is a critical auth-fix script. If it fails silently in production, developers will have no visibility into why the fix didn't work.
- **Suggestion:** At minimum, log a console warning in the catch block. Even better, use `console.error` so it surfaces in Electron's main process logs:

```js
catch (e) {
  console.error('[jovie] electron-runtime-init failed:', e);
}
```

---

**5. Cross-File Impact**

**Issue 7: `usePWAInstall.ts` — Behavior change affects all Electron runtime users, not just auth**

- **File:** `apps/web/hooks/usePWAInstall.ts`
- **Description:** The change unregisters *all* service workers in Electron runtime, not just those that might intercept auth routes. If any other part of the app relies on SW caching for performance (e.g., asset caching, offline support for the PWA shell), that functionality is now completely disabled in Electron.
- **Why it matters:** The PR description says "Desktop shells must never keep a PWA service worker controlling auth routes" but the implementation removes *all* SW functionality. If there's a legitimate use of SW caching in Electron (e.g., for static assets), this is an overcorrection.
- **Suggestion:** Verify no other code depends on SW presence in Electron. If there is a dependency, consider a more targeted approach (e.g., only unregister SWs whose scope matches auth routes, or use the SW update flow instead of full unregistration).

---

### Summary

| Category | Issues Found |
|---|---|
| Logic Errors | 3 (SW race condition, missing timeout on auth fetch, double-unregistration) |
| Security | 2 (UA injection surface, duplicated auth path list) |
| Data Integrity | None |
| Error Handling | 1 (silent failure in init script) |
| Cross-File Impact | 1 (blanket SW removal in Electron) |

**Most critical:** Issue 2 (no timeout on auth fetch in SW) — this could replace one broken state (offline fallback) with another (infinite hang). Issue 1 (race condition) undermines the core fix. Issue 5 (duplicated list) risks the same class of bug recurring.

itstimwhite (commented):

itstimwhite:

Branch: 'tim/jov-3119-p0-desktop-app-login-broken-auth-callback-link-fails-with'
State: BLOCKED



To resolve:
1. git fetch origin main
2. git checkout tim/jov-3119-p0-desktop-app-login-broken-auth-callback-link-fails-with
3. git rebase origin/main
5. git push --force-with-lease origin tim/jov-3119-p0-desktop-app-login-broken-auth-callback-link-fails-with

PR #10883, opened 2026-06-13
fix(insights): dedupe same-fact insight spam (JOV-3084)

itstimwhite:
## Summary
- dedupe same-fact insight spam before persistence using existing `dataSnapshot` source keys with normalized-copy fallback
- bound read-side dedup candidate windows for active and summary APIs so visible dedup no longer requires loading every active insight row
- keep total counts backed by database counts, preserve distinct same-source insight types except the intended city-growth/new-market duplicate family, and handle nested location snapshots consistently
- add focused cron and lifecycle regression tests for dedup behavior

## Validation
- `./scripts/setup.sh` (previous run)
- `pnpm --filter web exec vitest run tests/unit/api/cron/generate-insights.test.ts`
- `pnpm --filter web exec vitest run tests/unit/api/cron/generate-insights.test.ts tests/unit/lib/services/insights/lifecycle.test.ts`
- `pnpm biome check --write apps/web/lib/services/insights/lifecycle.ts apps/web/app/api/cron/generate-insights/route.ts apps/web/tests/unit/api/cron/generate-insights.test.ts apps/web/tests/unit/lib/services/insights/lifecycle.test.ts`
- `pnpm --filter @jovie/web run test:bug-to-test`
- `git diff --check`
- pre-commit hook: `next:proxy-guard`, Biome, ESLint, staged-file typecheck
- pre-push hook: `biome check .`, `next:proxy-guard`, `tailwind:check`, `typecheck`, `biome lint .`

## CodeRabbit
- `coderabbit review --plain --type uncommitted --base origin/main` completed analysis but repeatedly stalled during comment emission
- first local run emitted one nitpick; simplification applied
- final `coderabbit review findings` reported no stored findings for this scope

## Review response
- addressed Hermes review findings for unbounded reads, count semantics, priority fallback, source-key collisions, nested location snapshots, and priority-order drift note
- left duplicate preparation in `persistInsights` intentionally as a defensive boundary for non-cron callers

## Risks
- read APIs now dedupe within bounded candidate windows instead of scanning all active rows; totals remain database-backed active-row counts
- no schema, route contract, or UI changes

### DIFF
--- a/apps/web/app/api/cron/generate-insights/route.ts
+++ b/apps/web/app/api/cron/generate-insights/route.ts
@@ -17,6 +17,7 @@ import {
   expireStaleInsights,
   getExistingInsightTypes,
   persistInsights,
+  prepareInsightsForPersistence,
 } from '@/lib/services/insights/lifecycle';
 import {
   INSIGHTS_CRON_CONCURRENCY,
@@ -58,11 +59,12 @@ async function processProfile(
   const result = await generateInsights(metrics, existingTypes, {
     sessionId: profileId,
   });
+  const dedupedInsights = prepareInsightsForPersistence(result.insights);
 
   const persisted = await persistInsights(
     profileId,
     run.id,
-    result.insights,
+    dedupedInsights,
     metrics.period,
     metrics.comparisonPeriod
   );
--- a/apps/web/lib/services/insights/lifecycle.ts
+++ b/apps/web/lib/services/insights/lifecycle.ts
@@ -22,6 +22,7 @@ import { GENERATION_COOLDOWN_HOURS } from './thresholds';
 
 const INSIGHT_PRIORITY_ORDER = drizzleSql`case ${aiInsights.priority} when 'high' then 0 when 'medium' then 1 when 'low' then 2 end`;
 
+// Keep in sync with INSIGHT_PRIORITY_ORDER so DB and JS ordering match.
 const INSIGHT_PRIORITY_RANK: Record<InsightPriority, number> = {
   high: 0,
   medium: 1,
@@ -458,43 +459,71 @@ function getInsightSourceKey(dataSnapshot: unknown): string | null {
   }
 
   const snapshot = dataSnapshot as Record<string, unknown>;
-  const city = normalizeSnapshotToken(snapshot.city);
-  const country = normalizeSnapshotToken(snapshot.country);
+  const city = getSnapshotToken(snapshot, 'city', ['location', 'city']);
+  const country = getSnapshotToken(snapshot, 'country', [
+    'location',
+    'country',
+  ]);
   if (city) {
     return country ? `city:${city}|${country}` : `city:${city}`;
   }
 
-  const market = normalizeSnapshotToken(snapshot.market);
+  const market = getSnapshotToken(snapshot, 'market');
   if (market) {
     return `market:${market}`;
   }
 
-  const release = normalizeSnapshotToken(snapshot.release);
+  const release = getSnapshotToken(snapshot, 'release');
   if (release) {
     return `release:${release}`;
   }
 
-  const referrer = normalizeSnapshotToken(snapshot.referrer);
+  const referrer = getSnapshotToken(snapshot, 'referrer');
   if (referrer) {
     return `referrer:${referrer}`;
   }
 
-  const spikeDate = normalizeSnapshotToken(snapshot.spikeDate);
+  const spikeDate = getSnapshotToken(snapshot, 'spikeDate');
   if (spikeDate) {
     return `spikeDate:${spikeDate}`;
   }
 
-  const deviceType = normalizeSnapshotToken(snapshot.deviceType);
+  const deviceType = getSnapshotToken(snapshot, 'deviceType');
   if (deviceType) {
     return `deviceType:${deviceType}`;
   }
 
-  const linkType = normalizeSnapshotToken(snapshot.linkType);
+  const linkType = getSnapshotToken(snapshot, 'linkType');
   if (linkType) {
     return `linkType:${linkType}`;
   }
 
   return null;
+}
+
+function getSnapshotToken(
+  snapshot: Record<string, unknown>,
+  key: string,
+  nestedPath?: readonly string[]
+): string | null {
+  const topLevelValue = normalizeSnapshotToken(snapshot[key]);
+  if (topLevelValue) {
+    return topLevelValue;
+  }
+
+  if (!nestedPath) {
+    return null;
+  }
+
+  let value: unknown = snapshot;
+  for (const segment of nestedPath) {
+    if (!value || typeof value !== 'object') {
+      return null;
+    }
+    value = (value as Record<string, unknown>)[segment];
+  }
+
+  return normalizeSnapshotToken(value);
 }
 
 function normalizeSnapshotToken(value: unknown): string | null {
--- a/apps/web/lib/services/insights/lifecycle.ts
+++ b/apps/web/lib/services/insights/lifecycle.ts
@@ -20,6 +20,21 @@ import type {
 } from '@/types/insights';
 import { GENERATION_COOLDOWN_HOURS } from './thresholds';
 
+const INSIGHT_PRIORITY_ORDER = drizzleSql`case ${aiInsights.priority} when 'high' then 0 when 'medium' then 1 when 'low' then 2 end`;
+
+const INSIGHT_PRIORITY_RANK: Record<InsightPriority, number> = {
+  high: 0,
+  medium: 1,
+  low: 2,
+};
+
+type InsightDedupCandidate = {
+  category: InsightCategory;
+  title: string;
+  description: string;
+  dataSnapshot: unknown;
+};
+
 // ---------------------------------------------------------------------------
 // Insight Generation Runs
 // ---------------------------------------------------------------------------
@@ -110,13 +125,14 @@ export async function persistInsights(
   period: { start: Date; end: Date },
   comparisonPeriod: { start: Date; end: Date }
 ): Promise<number> {
-  if (insights.length === 0) return 0;
+  const dedupedInsights = prepareInsightsForPersistence(insights);
+  if (dedupedInsights.length === 0) return 0;
 
   const now = new Date();
   let persisted = 0;
 
   // Insert each insight individually to handle dedup gracefully
-  for (const insight of insights) {
+  for (const insight of dedupedInsights) {
     const expiresAt = new Date(now);
     expiresAt.setDate(expiresAt.getDate() + insight.expiresInDays);
 
@@ -208,28 +224,17 @@ export async function getActiveInsights(
 
   const whereClause = and(...conditions);
 
-  const [rows, countResult] = await Promise.all([
-    db
-      .select()
-      .from(aiInsights)
-      .where(whereClause)
-      .orderBy(
-        drizzleSql`case ${aiInsights.priority} when 'high' then 0 when 'medium' then 1 when 'low' then 2 end`,
-        desc(aiInsights.createdAt)
-      )
-      .limit(limit)
-      .offset(offset),
-    db
-      .select({ count: drizzleSql<number>`count(*)::int` })
-      .from(aiInsights)
-      .where(whereClause),
-  ]);
-
-  const total = countResult[0]?.count ?? 0;
+  const rows = await db
+    .select()
+    .from(aiInsights)
+    .where(whereClause)
+    .orderBy(INSIGHT_PRIORITY_ORDER, desc(aiInsights.createdAt));
+  const dedupedRows = dedupeVisibleInsights(rows);
+  const pagedRows = dedupedRows.slice(offset, offset + limit);
 
   return {
-    insights: rows.map(formatInsightResponse),
-    total,
+    insights: pagedRows.map(formatInsightResponse),
+    total: dedupedRows.length,
   };
 }
 
@@ -241,20 +246,11 @@ export async function getInsightsSummary(
 ): Promise<InsightsSummaryResponse> {
   const now = new Date();
 
-  const [insights, countResult, lastRun] = await Promise.all([
+  const [insights, lastRun] = await Promise.all([
     db
       .select()
       .from(aiInsights)
-      .where(
-        and(
-          eq(aiInsights.creatorProfileId, creatorProfileId),
-          eq(aiInsights.status, 'active'),
-          gte(aiInsights.expiresAt, now)
-        )
-      ),
-    db
-      .select({ count: drizzleSql<number>`count(*)::int` })
-      .from(aiInsights)
+      .orderBy(INSIGHT_PRIORITY_ORDER, desc(aiInsights.createdAt))
       .where(
         and(
           eq(aiInsights.creatorProfileId, creatorProfileId),
@@ -275,13 +271,14 @@ export async function getInsightsSummary(
       .limit(1),
   ]);
 
+  const dedupedInsights = dedupeVisibleInsights(insights);
   const sortedInsights = sortInsightsForChat(
-    insights.map(formatInsightResponse)
+    dedupedInsights.map(formatInsightResponse)
   );
 
   return {
     insights: sortedInsights.slice(0, 3),
-    totalActive: countResult[0]?.count ?? 0,
+    totalActive: dedupedInsights.length,
     lastGeneratedAt: lastRun[0]?.createdAt
       ? toISOStringSafe(lastRun[0].createdAt)
       : null,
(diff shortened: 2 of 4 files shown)


itstimwhite:
## PR #10883 Review

---

**1. Logic Error — `getActiveInsights` pagination is now incorrect**

**File:** `apps/web/lib/services/insights/lifecycle.ts` (new `getActiveInsights`, ~line 224)

The old code applied `LIMIT`/`OFFSET` at the SQL level, then returned the DB count as `total`. The new code fetches **all** rows, dedupes in memory, then slices client-side. Two problems:

- **`total` is now the post-dedup count, not the pre-dedup count.** Any caller that renders a "showing X of Y" indicator or uses `total` to calculate page counts will report a number that doesn't match the actual DB state. This is a semantic regression — the old `total` meant "total active rows matching filters," now it means "total after dedup."

- **Performance regression.** Every paginated request now loads the entire active-insights table into memory. For a profile with hundreds of active insights, this is wasteful and unbounded. The old query was bounded by `limit + offset`.

**Suggested fix:** Either (a) keep pagination at the SQL level and accept that dedup is only approximate (best-effort per-page), or (b) if full-table dedup is required, add a `maxRows` safety cap and document that `total` now reflects post-dedup counts so callers can be updated accordingly.

---

**2. Logic Error — `getInsightsSummary` fetches unbounded rows**

**File:** `apps/web/lib/services/insights/lifecycle.ts` (new `getInsightsSummary`, ~line 246)

Same pattern as above but worse: the summary query previously had an implicit limit (only needed top 3). Now it fetches **all** active rows for the profile, dedupes, then takes `slice(0, 3)`. For a busy profile this could be thousands of rows for no reason.

**Suggested fix:** Add a `LIMIT` (e.g., 50 or 100) to the summary query, or keep the old `Promise.all` with the count query and apply dedup only to the limited result set.

---

**3. Logic Error — `prepareInsightsForPersistence` sort comparator can return `NaN`**

**File:** `apps/web/lib/services/insights/lifecycle.ts` (`prepareInsightsForPersistence`, ~line 355)

```ts
(left, right) =>
  INSIGHT_PRIORITY_RANK[left.priority] -
    INSIGHT_PRIORITY_RANK[right.priority] ||
  right.confidence - left.confidence
```

If `left.priority` or `right.priority` is not in `INSIGHT_PRIORITY_RANK` (e.g. a new priority level added later), the subtraction produces `NaN`, and `NaN || ...` evaluates the right side. This means items with unknown priorities sort by confidence only, silently mixing with known-priority items. Not a crash, but a subtle ordering bug.

**Suggested fix:** Add a fallback rank (e.g., `?? 99`) or assert/validate priorities before sorting.

---

**4. Logic Error — `dedupeVisibleInsights` uses `buildInsightDedupKey` which can produce collisions across semantically different insights**

**File:** `apps/web/lib/services/insights/lifecycle.ts` (`buildInsightDedupKey`, ~line 409)

The fallback key is `${category}|${normalizedTitle}|${normalizedDescription}`. Two insights with the same category and nearly-identical copy but **different** `dataSnapshot` values (e.g., "Austin grew 34%" vs "Austin grew 12%") would collide if the LLM happened to produce the same title/description. This is unlikely but possible with templated output.

More concretely: the `getInsightSourceKey` function checks `city`, `market`, `release`, `referrer`, `spikeDate`, `deviceType`, `linkType` — but **not** `insightType`. Two insights with the same city but different `insightType` (e.g., `city_growth` vs `city_decline`) would produce the same dedup key and one would be silently dropped. The test covers this implicitly (different categories), but same-category/different-type same-city insights would be lost.

**Suggested fix:** Include `insightType` in the dedup key, or at minimum document this as a known limitation.

---

**5. Data Integrity — `persistInsights` now silently drops insights without logging**

**File:** `apps/web/lib/services/insights/lifecycle.ts` (`persistInsights`, ~line 125)

The function calls `prepareInsightsForPersistence` internally, so dedup happens twice for the cron path (once in `processProfile`, once in `persistInsights`). This is harmless but redundant. More importantly, when insights are dropped by dedup, there is no logging — an operator debugging why expected insights didn't appear would have no signal.

**Suggested fix:** Remove the redundant call in `persistInsights` (since `processProfile` already dedupes), or add a debug-level log when insights are dropped.

---

**6. Security — None found.**

No hardcoded secrets, SQL injection vectors, XSS, auth bypass, or CSP issues introduced. The dedup logic operates on already-generated data, not user input.

---

**7. Error Handling — None found.**

No new unhandled promise rejections or silent failures introduced. The dedup functions are synchronous and operate on in-memory arrays.

---

**8. Cross-file Impact**

- **`getActiveInsights`** is called by the insights read API route. The pagination regression (issue #1) will affect any UI that displays total counts or page numbers. Verify that the consuming component doesn't rely on `total` matching the DB count.
- **`getInsightsSummary`** is called by the chat/summary endpoint. The unbounded query (issue #2) could cause memory pressure under load.
- No other callers of `persistInsights` are affected since the function signature is unchanged.

---

### Summary

| 1 | **Medium** | `lifecycle.ts` ~L224 | `getActiveInsights` pagination broken — loads all rows, `total` semantics changed |
| 2 | **Medium** | `lifecycle.ts` ~L246 | `getInsightsSummary` unbounded query — fetches all active rows into memory |
| 3 | **Low** | `lifecycle.ts` ~L355 | Sort comparator can produce `NaN` for unknown priorities |
| 4 | **Low** | `lifecycle.ts` ~L409 | Dedup key doesn't include `insightType` — same-city different-type insights may collide |
| 5 | **Low** | `lifecycle.ts` ~L125 | Double-dedup in cron path; no logging when insights are silently dropped |

itstimwhite (commented):

itstimwhite:

Branch: 'tim/jov-3084-audit-p1-insight-generation-emits-duplicates-same-fact-shown'
State: BLOCKED



To resolve:
1. git fetch origin main
2. git checkout tim/jov-3084-audit-p1-insight-generation-emits-duplicates-same-fact-shown
3. git rebase origin/main
5. git push --force-with-lease origin tim/jov-3084-audit-p1-insight-generation-emits-duplicates-same-fact-shown

itstimwhite:
## PR #10883 Review

---

**1. Logic Error — `getActiveInsights` no longer applies `limit`/`offset` at the DB level**

**File:** `apps/web/lib/services/insights/lifecycle.ts` (the `getActiveInsights` function, post-diff)

**Description:** The old code applied `.limit(limit).offset(offset)` in the SQL query. The new code removes both, fetches the entire filtered table into memory, deduplicates in JS, then slices with `dedupedRows.slice(offset, offset + limit)`.

**Why it matters:**
- **Performance regression:** For profiles with hundreds of active insights, this pulls the entire set into memory on every paginated request. The old code let Postgres handle pagination efficiently.
- **Correctness risk:** The `total` returned is `dedupedRows.length` (post-dedup count), but the `insights` array is a post-dedup slice. This is internally consistent, so pagination math isn't broken — but the DB now does more work per request than before.
- **The PR description says** *"Performance must not replace route UIs (same design, faster)"* — this change makes the read path strictly slower and heavier.

**Suggested fix:** Keep `limit`/`offset` in the DB query for the initial fetch, but also apply dedup post-fetch. If the concern is that dedup changes the effective page size, accept that tradeoff or fetch a larger window and trim. At minimum, keep the DB-level limit to avoid unbounded memory growth:

```ts
const rows = await db
  .select()
  .from(aiInsights)
  .where(whereClause)
  .orderBy(INSIGHT_PRIORITY_ORDER, desc(aiInsights.createdAt))
  .limit(limit + offset)  // or a reasonable overscan factor
  .offset(offset);
const dedupedRows = dedupeVisibleInsights(rows);
const pagedRows = dedupedRows.slice(0, limit);
```

---

**2. Logic Error — `getInsightsSummary` fetches unbounded rows**

**File:** `apps/web/lib/services/insights/lifecycle.ts` (the `getInsightsSummary` function, post-diff)

**Description:** Same pattern as above. The old code had no explicit limit but relied on the DB to return only active/non-expired rows. The new code adds `.orderBy(INSIGHT_PRIORITY_ORDER, ...)` but still fetches all matching rows with no limit, then deduplicates in JS.

**Why it matters:** For the summary path this is less critical (it only takes `.slice(0, 3)`), but it still pulls all active insights into memory. If a profile accumulates many active insights over time, this grows without bound. A simple `.limit(50)` or similar would be a safe guard.

**Suggested fix:** Add a reasonable `.limit()` to the summary query.

---

**3. Logic Error — `prepareInsightsForPersistence` sort comparator uses `||` instead of fallback**

**File:** `apps/web/lib/services/insights/lifecycle.ts`, in `prepareInsightsForPersistence`:

```ts
(left, right) =>
  INSIGHT_PRIORITY_RANK[left.priority] -
    INSIGHT_PRIORITY_RANK[right.priority] ||
  right.confidence - left.confidence
```

**Description:** The `||` operator returns the left operand if it's truthy, otherwise the right. When priorities are equal, the subtraction yields `0` (falsy), so `||` correctly falls through to the confidence comparison. This works **by accident** — `0 || x` evaluates to `x`. However, if two insights have the same priority AND the same confidence, the comparator returns `0`, which is correct (stable-ish). The real issue is readability and intent: `||` for numeric fallback is a footgun. If someone refactored to return a boolean or changed the subtraction, this would silently break.

**Why it matters:** Not a bug today, but fragile. A ternary or explicit `=== 0` check would be clearer and safer.

**Suggested fix:**
```ts
(left, right) => {
  const priorityDiff =
    INSIGHT_PRIORITY_RANK[left.priority] - INSIGHT_PRIORITY_RANK[right.priority];
  return priorityDiff !== 0 ? priorityDiff : right.confidence - left.confidence;
}
```

---

**4. Logic Error — `getInsightSourceKey` only checks top-level `dataSnapshot` fields**

**File:** `apps/web/lib/services/insights/lifecycle.ts`, `getInsightSourceKey` function

**Description:** The function reads `city`, `country`, `market`, `release`, `referrer`, `spikeDate`, `deviceType`, `linkType` as top-level keys on `dataSnapshot`. If any of these are nested (e.g., `dataSnapshot.location.city`), they won't be found and the function falls through to the title+description normalization fallback.

**Why it matters:** This isn't necessarily a bug — it depends on the actual shape of `dataSnapshot` in the codebase. But if `dataSnapshot` is sometimes nested, the dedup key becomes inconsistent: two insights about the same city might get different keys depending on whether `city` is top-level or nested, causing them to not deduplicate. Worth verifying against actual `dataSnapshot` shapes in the codebase.

**Suggested fix:** Audit the actual shape of `dataSnapshot` across insight generators. If nested, update `getInsightSourceKey` to handle both shapes or normalize at write time.

---

**5. Data Integrity — `getActiveInsights` total count is now post-dedup**

**File:** `apps/web/lib/services/insights/lifecycle.ts`, `getActiveInsights` return value

**Description:** The old code returned `total` from a `count(*)` SQL query (pre-dedup count). The new code returns `total: dedupedRows.length` (post-dedup count).

**Why it matters:** Any consumer that uses `total` to render "Showing X of Y insights" or calculate page counts will now see a smaller number. This is arguably *more accurate* (you're deduplicating, so the real count is lower), but it's a **behavioral change** in the API response contract. If the frontend or any API consumer relies on `total` matching the pre-dedup count for pagination math, this could cause confusion. The PR description says *"it does not add a new source-datum field or broader insight lifecycle changes"* — changing the semantics of `total` is arguably a lifecycle/contract change.

**Suggested fix:** Decide intentionally whether `total` should be pre- or post-dedup. If post-dedup, document the change. If pre-dedup, run the count query separately and return both values.

---

**6. Minor — `INSIGHT_PRIORITY_ORDER` SQL fragment is extracted but `INSIGHT_PRIORITY_RANK` is used in JS sort**

**File:** `apps/web/lib/services/insights/lifecycle.ts`

**Description:** `INSIGHT_PRIORITY_ORDER` (the raw SQL `case` expression) is used in DB queries. `INSIGHT_PRIORITY_RANK` (the JS object) is used in `prepareInsightsForPersistence`. These are two separate representations of the same priority ordering. If someone adds a new priority level, they must remember to update both.

**Why it matters:** Divergence risk. Not a bug today, but a maintenance hazard.

**Suggested fix:** Derive one from the other, or add a comment linking them.

---

**Security:** None found. No hardcoded secrets, no SQL injection (Drizzle parameterized queries are used throughout), no auth bypass, no XSS vectors.

**Error handling:** None found missing. The cron route already had error handling around `processProfile`. The new functions are pure transforms with no fallible operations.

**Cross-file impact:**
- The `persistInsights` function signature is unchanged — callers are unaffected.
- `getActiveInsights` return type is unchanged (`{ insights, total }`), but the semantics of `total` shifted (see issue 5 above).
- `getInsightsSummary` return type is unchanged, but `totalActive` is now post-dedup (same concern as issue 5).
- The new exports `prepareInsightsForPersistence` and `dedupeVisibleInsights` are correctly used within the module and the cron route.

itstimwhite (commented):

itstimwhite:

### Logic Errors

**1. `lifecycle.ts:237` — `getActiveInsights` applies `offset` before dedup, causing incorrect pagination**

The candidate window is fetched with `.limit(candidateLimit).offset(offset)`, then deduped, then sliced to `limit`. This means page 2 starts reading from row `offset` in the DB, deduplicates within that window, and returns whatever remains. If dedup removes rows, the page returns fewer items than `limit` and the *next* page will skip rows that should have appeared. Pagination is fundamentally broken for any page beyond the first.

**Why it matters:** Users scrolling through insights will see missing or empty pages once duplicates exist in the dataset.

**Suggested fix:** Either dedup in the DB query itself (e.g., using `DISTINCT ON` with the dedup key stored as a generated column), or remove `offset` from the DB query and apply it after dedup in JS (accepting the performance trade-off). The current approach of mixing DB offset with in-memory dedup is unsound.

---

**2. `lifecycle.ts:262` — `getInsightsSummary` no longer applies `INSIGHT_SUMMARY_CANDIDATE_LIMIT` to the DB query**

The `insights` query at line ~268 has `.limit(INSIGHT_SUMMARY_CANDIDATE_LIMIT)` applied, but the `totalRows` count query at line ~280 has **no limit** — it counts all active rows. This is correct for `totalActive`. However, the `insights` query is missing an `orderBy` before `where` in the diff — looking at the diff structure, the `.orderBy(INSIGHT_PRIORITY_ORDER, desc(aiInsights.createdAt))` is chained before `.where(activeWhereClause)`. In Drizzle, `.orderBy().where()` is valid but unusual ordering; verify this compiles and produces correct SQL. If Drizzle emits `ORDER BY ... WHERE ...` instead of `WHERE ... ORDER BY ...`, this is a syntax error.

**Why it matters:** If the chain order is wrong, the query may throw at runtime or silently misbehave.

**Suggested fix:** Reorder to `.where(activeWhereClause).orderBy(INSIGHT_PRIORITY_ORDER, desc(aiInsights.createdAt)).limit(INSIGHT_SUMMARY_CANDIDATE_LIMIT)` to match standard SQL emission order and the pattern used elsewhere.

---

**3. `lifecycle.ts:380` — `prepareInsightsForPersistence` sorts by priority then confidence, but `dedupeVisibleInsights` is first-write-wins**

The sort puts high-priority first, then by confidence descending. `dedupeVisibleInsights` keeps the first occurrence. This means the highest-priority, highest-confidence insight wins — which is correct. However, the sort comparator at line ~384 uses `right.confidence - left.confidence` (descending), which is correct. **No bug here**, but the sort is unstable for equal priority+confidence — two insights with identical priority and confidence may dedupe differently depending on input order. This is a minor determinism concern, not a bug.

---

**4. `lifecycle.ts:429` — `getSnapshotToken` with `nestedPath` only checks one level of nesting for `city`/`country`**

The `nestedPath` for `city` is `['location', 'city']`. The function iterates through segments, so `snapshot.location.city` is checked. But if the data has `snapshot.city` directly, that's checked first via the top-level lookup. This is correct. However, if data has *both* `snapshot.city` and `snapshot.location.city`, the top-level wins silently — which may mask data inconsistencies.

**Why it matters:** Not a bug per se, but worth a comment documenting the precedence.

---

### Security

**None found.** No hardcoded secrets, SQL injection vectors (Drizzle parameterized queries used throughout), XSS risks, or auth bypasses introduced.

---

### Data Integrity

**5. `lifecycle.ts:237` — `total` in `getActiveInsights` counts all DB rows, but `insights` returns a deduped subset**

The `total` reflects the raw DB count (including duplicates), while the `insight` array contains deduped results. Any consumer using `total` for pagination math (e.g., "showing 1–10 of 45") will report 45 total but only be able to page through, say, 30 unique items. This is a contract inconsistency.

**Why it matters:** Pagination UI will show incorrect total counts or allow navigating to empty pages.

**Suggested fix:** Either make `total` reflect the deduped count (expensive — requires full scan), or rename it to clarify it's a raw count and let the consumer handle it. Document the semantics.

---

### Error Handling

**6. `lifecycle.ts:131` — `persistInsights` now calls `prepareInsightsForPersistence` internally AND the caller in `route.ts` also calls it**

In `route.ts:62`, `prepareInsightsForPersistence` is called explicitly, then `persistInsights` is called with the result. Inside `persistInsights` at line 134, `prepareInsightsForPersistence` is called again. This is intentional per the PR description ("defensive boundary for non-cron callers"), but it means the sort+dedup runs twice for the cron path. Not a bug, but wasteful.

**Why it matters:** Minor performance concern; not a correctness issue since dedup is idempotent.

---

### Cross-File Impact

**7. `getActiveInsights` return type unchanged but semantics changed — callers may break**

The return type `{ insights: InsightResponse[]; total: number }` is the same, but `insights` is now a deduped subset of the candidate window, not a direct DB page. Any caller that assumes `insights.length === limit` (for UI rendering) may get fewer items. Any caller using `total` for pagination calculations will overestimate page count.

**Why it matters:** Check all callers of `getActiveInsights` to confirm they handle variable-length pages correctly.

**Suggested fix:** Audit callers. If any assume fixed page size, update them or document the new contract.

---

### Summary

| 1 | **High** | `lifecycle.ts:237` | Pagination broken: DB offset + in-memory dedup = skipped/missing rows on page 2+ |
| 2 | **Medium** | `lifecycle.ts:262` | `.orderBy().where()` chain order may produce invalid SQL |
| 3 | Low | `lifecycle.ts:384` | Unstable sort for equal priority+confidence |
| 4 | Low | `lifecycle.ts:429` | Silent precedence of top-level over nested snapshot keys |
| 5 | **Medium** | `lifecycle.ts:237` | `total` counts raw rows but `insights` is deduped — pagination contract mismatch |
| 6 | Low | `lifecycle.ts:131` | Double sort+dedup for cron path (intentional but wasteful) |
| 7 | **Medium** | `lifecycle.ts:230` | Callers of `getActiveInsights` may assume fixed page size |

**Must-fix before merge:** Issues #1 and #5 together make pagination unusable with duplicates. Issue #2 needs verification that Drizzle emits correct SQL with that chain order.

itstimwhite (commented):

PR #10889, opened 2026-06-13
fix(automation): keep codex shipper singleton draining

itstimwhite:
## Summary
- changes the Hermes Codex issue shipper from one sequential issue per run to a singleton drain loop with up to 3 safe parallel coder lanes
- adds load/free-memory capacity gating, non-blocking duplicate-cron skip behavior, per-issue isolated worktrees, and worktree cleanup
- updates the launchd template, Hermes Air docs, and local desktop automation prompt for the same singleton always-shipping policy

## Validation
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH pnpm exec biome check --write scripts/hermes/jobs/codex-issue-shipper.ts scripts/hermes/lib/codex-issue-shipper.ts scripts/hermes/lib/heavy-job-lock.ts scripts/lib/__tests__/codex-issue-shipper.test.mjs docs/HERMES_AIR.md`
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH pnpm ci:harness:test` (5 files, 45 tests passed)
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH pnpm typecheck` (8 packages successful)
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH HERMES_CODEX_SHIPPER_DRY_RUN=1 pnpm exec tsx scripts/hermes/jobs/codex-issue-shipper.ts` (scan path completed; 0 dispatchable trusted issues; capacity gate refused launches under low free memory)
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH coderabbit review --agent -c AGENTS.md -t uncommitted` (first pass found 2 valid minor issues; fixed both)
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH coderabbit review --agent -c AGENTS.md -t uncommitted` (second pass: 0 findings)
- push pre-push hook: `biome check .`, `next:proxy-guard`, `tailwind:check`, `@jovie/web typecheck`, `@jovie/web lint` passed

## Rollout / Cost Impact
- Existing cron retained; no new scheduled job added.
- Default fan-out increases to at most 3 coder lanes per singleton owner, bounded by free memory and load.
- The active run logs `singleton_active_skip` for overlapping cron invocations instead of starting duplicate queue owners.
- Unit cost remains measurable as CI minutes x runner cost plus agent minutes x model cost; reduce `HERMES_CODEX_SHIPPER_MAX_PARALLEL_AGENTS` if `capacity_throttled` or CI congestion becomes frequent.

## Notes
- Local shell default is Node v20.20.2, so git hooks printed engine warnings. Validation above ran with Node v22.22.1 via `/opt/homebrew/opt/node@22/bin`.
- The `codex-automation` label is not present in this repository, so this PR is labeled `codex` only.

### DIFF
--- a/docs/HERMES_AIR.md
+++ b/docs/HERMES_AIR.md
@@ -189,11 +189,17 @@ tsx scripts/hermes/jobs/codex-issue-shipper.ts
 
 The job watches open GitHub issues labeled `codex`. Empty runs only call GitHub, write a JSONL event, and exit. No gbrain query, model call, subagent, or CodeRabbit review starts until an eligible issue exists.
 
+Only one shipper run may own the queue at a time. A new cron invocation takes a non-blocking singleton lock check; if another shipper is still active, the new invocation logs `singleton_active_skip` and exits. The active run keeps draining the queue in batches until no eligible issues remain, the machine is under too much pressure to launch another agent, or all remaining issues are blocked or human-gated.
+
 Config variables:
 
 | Variable | Default | Purpose | Update path |
 |---|---:|---|---|
-| `HERMES_CODEX_SHIPPER_MAX_ISSUES_PER_RUN` | `1` | Max Codex issues shipped per run; controls concurrency and spend | Doppler or plist env, then `./scripts/hermes/bootstrap-air.sh --reconfigure` |
+| `HERMES_CODEX_SHIPPER_MAX_ISSUES_PER_RUN` | `3` | Max Codex issues selected per drain batch; also caps parallel agent fan-out | Doppler or plist env, then `./scripts/hermes/bootstrap-air.sh --reconfigure` |
+| `HERMES_CODEX_SHIPPER_MAX_PARALLEL_AGENTS` | `3` | Absolute cap for concurrent coder agents inside the singleton shipper run | Doppler or plist env, then `./scripts/hermes/bootstrap-air.sh --reconfigure` |
+| `HERMES_CODEX_SHIPPER_MIN_FREE_MEMORY_MB` | `4096` | Below this free-memory floor, launch at most one new coder; below half this floor, launch none | Doppler or plist env, then `./scripts/hermes/bootstrap-air.sh --reconfigure` |
+| `HERMES_CODEX_SHIPPER_MAX_LOAD_PER_CPU` | `1.5` | Above this one-minute load-per-CPU threshold, launch at most one new coder; above 1.5x this threshold, launch none | Doppler or plist env, then `./scripts/hermes/bootstrap-air.sh --reconfigure` |
+| `HERMES_CODEX_SHIPPER_SINGLETON_LOCK_STALE_MS` | `28800000` | Long-running shipper lock TTL; new crons skip while the owning process is alive | Doppler or plist env, then `./scripts/hermes/bootstrap-air.sh --reconfigure` |
 | `HERMES_CODEX_SHIPPER_INTEGRATION_THRESHOLD` | `4` | Eligible queue depth that routes trainable issues through `integration/codex-queue` | Doppler or plist env, then `./scripts/hermes/bootstrap-air.sh --reconfigure` |
 | `HERMES_CODEX_SHIPPER_AGENT` | `claude` | Local coding agent binary; `claude` keeps subagent support, `codex` is available as an explicit override | Doppler or plist env, then `./scripts/hermes/bootstrap-air.sh --reconfigure` |
 
@@ -211,9 +217,9 @@ Ship now / Re-evaluate when / Then:
 
 | Decision | Trigger | Action |
 |---|---|---|
-| Ship now | Default `HERMES_CODEX_SHIPPER_MAX_ISSUES_PER_RUN=1`, because one coder session can consume hours of agent time plus CI minutes | Keep one issue per run and let the 15-minute cron continue draining |
-| Re-evaluate when | Eligible `codex` queue stays above 4 issues for two consecutive runs and cost per shipped issue stays under the weekly unit target: CI minutes x runner cost plus agent minutes x model cost | Raise `HERMES_CODEX_SHIPPER_MAX_ISSUES_PER_RUN` or add an `integration-branch` label to queue trainable issues |
-| Then | CI minutes x runner cost plus agent minutes x model cost stay within the weekly agent budget | Keep the higher concurrency; otherwise return to one issue per run |
+| Ship now | Default `HERMES_CODEX_SHIPPER_MAX_ISSUES_PER_RUN=3` and `HERMES_CODEX_SHIPPER_MAX_PARALLEL_AGENTS=3`, because one singleton run can keep the queue hot without duplicate cron owners | Keep up to three coder lanes active while free memory stays above 4096MB and load stays below 1.5 per CPU |
+| Re-evaluate when | Cost per shipped issue rises above the weekly unit target: CI minutes x runner cost plus agent minutes x model cost, or local pressure repeatedly logs `capacity_throttled` | Lower `HERMES_CODEX_SHIPPER_MAX_PARALLEL_AGENTS`, raise machine capacity, or route more trainable issues through `integration-branch` labels |
+| Then | CI minutes x runner cost plus agent minutes x model cost stay within the weekly agent budget | Keep three lanes; otherwise reduce the cap until the queue drain cost is back inside target |
 
 ## Recovery Procedures
 
--- a/scripts/hermes/jobs/codex-issue-shipper.ts
+++ b/scripts/hermes/jobs/codex-issue-shipper.ts
@@ -11,7 +11,7 @@
  * issue is eligible and claimed.
  */
 
-import { execFileSync, spawnSync } from 'node:child_process';
+import { execFileSync, spawn } from 'node:child_process';
 import {
   appendFileSync,
   closeSync,
@@ -19,8 +19,10 @@ import {
   mkdirSync,
   openSync,
   readFileSync,
+  rmSync,
   writeFileSync,
 } from 'node:fs';
+import { availableParallelism, freemem, loadavg, tmpdir } from 'node:os';
 import { join } from 'node:path';
 
 import {
@@ -40,7 +42,7 @@ import {
   loadShipperConfig,
   type ShipperConfig,
 } from '../lib/codex-issue-shipper';
-import { withHeavyJobLock } from '../lib/heavy-job-lock';
+import { tryWithHeavyJobLock } from '../lib/heavy-job-lock';
 import { HERMES_PATHS } from '../lib/hermes-paths';
 import { logJobEvent, withJobLogging } from '../lib/jobs-log';
 
@@ -53,6 +55,15 @@ interface AgentRunResult {
   readonly error?: string;
   readonly logPath: string;
   readonly promptPath: string;
+}
+
+interface CapacitySnapshot {
+  readonly allowedAgents: number;
+  readonly cpuCount: number;
+  readonly freeMemoryMb: number;
+  readonly loadAverage1m: number;
+  readonly loadPerCpu: number;
+  readonly reasons: ReadonlyArray<string>;
 }
 
 function shortError(err: unknown): string {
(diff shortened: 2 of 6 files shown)


itstimwhite:

### Logic Errors

**1. `scripts/hermes/jobs/codex-issue-shipper.ts` — `runAgent` timeout races with `child.stdin.end`**

The `timeout` is set via `setTimeout` on line ~650, but the old code passed `timeout` directly to `spawnSync`. In the new async version, if the child process closes before the timeout fires, `clearTimeout(timeout)` in `finish()` is correct. However, if the timeout fires, `child.kill('SIGTERM')` is called, but `child.stdin.end(prompt)` on line ~668 is called *outside* the Promise constructor. If the child exits very quickly (before `child.stdin.end` runs), calling `.end()` on an already-closed stdin will throw an unhandled error.

**Why it matters:** Unhandled `error` event on stdin after child already closed would crash the process or produce an unhandled rejection.

**Suggested fix:** Move `child.stdin.end(prompt)` inside the Promise, after the event listeners are registered, and guard it with a check or wrap in try/catch.

---

**2. `scripts/hermes/jobs/codex-issue-shipper.ts` — `dispatchPlan` logs `plan.issue.number` instead of `dispatch.issue.number` in `missing_pr_after_success` event**

At the `missing_pr_after_success` log event (around line ~760 in the new diff), the code references `plan.issue.number` and `dispatch.branchName`. The variable `plan` is the original parameter, while `dispatch` is the potentially-modified plan from `prepareWorktree` (which may have a different `branchName`). The `branch` field correctly uses `dispatch.branchName`, but `issue` uses `plan.issue.number` — this is actually fine since the issue number doesn't change between `plan` and `dispatch`. However, this inconsistency is a maintenance hazard.

**Why it matters:** Low severity — currently correct but fragile. If `prepareWorktree` ever modifies the issue reference, this would silently log the wrong number.

---

**3. `scripts/hermes/jobs/codex-issue-shipper.ts` — `dispatchBatch` catches errors per-plan but `dispatchPlan` also calls `markBlocked` internally on gbrain failure**

In `dispatchPlan`, if `collectGbrainContext` throws, the function calls `releaseClaimForRetry` and returns early (no `markBlocked`). But if `runAgent` fails, it calls `markBlocked`. Then in `dispatchBatch`, the outer `try/catch` also calls `markBlocked` on any thrown error. Since `dispatchPlan` doesn't re-throw after handling errors internally, the outer catch in `dispatchBatch` should never fire for handled cases — but if there's an unhandled throw from `claimIssue` or `findPrForBranch`, it would get double-handled (once in `dispatchPlan`'s internal logic, once in `dispatchBatch`).

**Why it matters:** If `claimIssue` throws synchronously before the try/catch in `dispatchPlan`, the error propagates to `dispatchBatch` which calls `markBlocked` — but the issue was never claimed, so `markBlocked` may operate on an unclaimed issue.

**Suggested fix:** Wrap the entire body of `dispatchPlan` in a try/catch that converts errors to a no-op return (after logging), so `dispatchBatch`'s outer catch only handles truly unexpected failures.

---

**4. `scripts/hermes/jobs/codex-issue-shipper.ts` — `cleanupWorktree` called in `finally` even when `prepareWorktree` failed partway**

If `prepareWorktree` fails after `mkdirSync` succeeds but before `run(['git', 'worktree', 'add', ...])` completes, the partially-created directory exists but `repoRoot` is still returned. The `finally` block calls `cleanupWorktree(config, prepared.repoRoot)` which tries `git worktree remove --force` — this will fail (worktree was never created), then fall through to `rmSync(repoRoot, ...)` which will clean up the empty directory. This is actually safe, but the `worktree_remove_failed` error log is misleading.

**Why it matters:** Noise in logs during expected failure paths.

---

**5. `scripts/hermes/jobs/codex-issue-shipper.ts` — `branchExists` uses `spawnSyncSafe` which wraps `execFileSync` with `timeout: 30_000`, but `git ls-remote` can hang indefinitely on network issues**

The `timeout` is 30 seconds, which is reasonable. However, `spawnSyncSafe` swallows all errors and returns `{ status: 1 }` on timeout. The caller in `branchExists` treats any non-zero status as "branch doesn't exist." A timeout on `git ls-remote` (network down) would be silently treated as "branch doesn't exist," causing the shipper to use the original branch name instead of the suffixed one.

**Why it matters:** If network is down and a branch name collision exists, the shipper could attempt to create a worktree with an existing branch name, causing `git worktree add -b` to fail.

**Suggested fix:** Distinguish between "definitely doesn't exist" (exit code 0/1 from ls-remote) and "unknown" (timeout/other errors). On "unknown," append the timestamp suffix to be safe.

---

**6. `scripts/hermes/lib/heavy-job-lock.ts` — `tryWithHeavyJobLock` releases lock in `finally` after `fn` completes, but `fn` is the entire drain loop**

The drain loop (`for (;;)`) inside `tryWithHeavyJobLock` runs until the queue is empty or capacity is zero. The lock is held for the entire duration. The `staleMs` is set to 8 hours (`singletonLockStaleMs`), which is the lock TTL. But `reapIfStale` checks `age > staleMs` — meaning a new cron invocation will only reclaim the lock after 8 hours. This is intentional per the description ("non-blocking singleton lock skip"), but if the drain loop exits early (e.g., `capacity_throttled`), the lock is released immediately via `finally`, and a new cron could start — potentially before the previous run's agents have finished.

**Why it matters:** Two concurrent shipper runs could be active simultaneously if the first run exits the drain loop due to capacity throttling but still has in-flight agents.

**Suggested fix:** Consider whether the lock should be released when agents are still running. If the intent is "one shipper run at a time," the lock should persist until all spawned agents complete.

---

### Security

**1. `scripts/hermes/jobs/codex-issue-shipper.ts` — `prepareWorktree` uses `tmpdir()` as default worktree root without sanitization**

`process.env.HERMES_CODEX_SHIPPER_WORKTREE_ROOT` is used directly. If an attacker can influence this environment variable, they could direct worktree creation to an arbitrary directory. However, this is a local script with env vars controlled by the operator, so the risk is low.

**Why it matters:** Low severity — local execution context, env vars are operator-controlled.

---

**2. `scripts/hermes/jobs/codex-issue-shipper.ts` — `branchExists` constructs `branchName` into shell commands via `execFileSync`**

`branchName` is passed as an argument array element to `execFileSync`, which does not invoke a shell. This is safe from injection.

**Verdict:** No issue — `execFileSync` with array args is injection-safe.

---

### Data Integrity

**1. `scripts/hermes/jobs/codex-issue-shipper.ts` — `dispatchPlan` calls `claimIssue` before `ensureControlLabels`**

In the new flow, `ensureControlLabels` is called lazily (`controlLabelsEnsured` flag) *after* the capacity check and *inside* the drain loop. But `claimIssue` is called inside `dispatchPlan`, which runs after `ensureControlLabels` is set. However, on the very first iteration where `batch.length > 0`, `ensureControlLabels` is called, then `dispatchBatch` runs. On subsequent iterations, it's skipped. This ordering is correct.

**Verdict:** No issue.

---

### Error Handling

**1. `scripts/hermes/jobs/codex-issue-shipper.ts` — `runAgent`: `closeSync(fd)` called in `finish()`, but if `openSync` throws, `fd` is never assigned**

Looking at the code: `const fd = openSync(logPath, 'a');` is at line ~631, and `finish()` calls `closeSync(fd)`. If `openSync` throws, the error propagates to the caller — but the Promise constructor wraps everything, so the error would be caught by `dispatchPlan`'s try/catch. However, `runAgent` returns a `Promise` — if `openSync` throws synchronously *before* the Promise constructor body executes, the function throws synchronously instead of returning a rejected promise.

**Why it matters:** `dispatchPlan` does `await runAgent(...)`. If `runAgent` throws synchronously (before entering the Promise), the `await` will still catch it. But the error message would be less clear since it wouldn't go through the `error` event handler.

**Suggested fix:** Wrap the entire `runAgent` body in `return new Promise(resolve => { ... })` from the very first line, including `openSync`.

---

**2. `scripts/hermes/jobs/codex-issue-shipper.ts` — `runAgent`: `child.stdin.end(prompt)` called outside Promise**

As noted in Logic Error #1, `child.stdin.end(prompt)` at line ~668 is outside the Promise constructor. If the child process exits before this line executes, stdin may already be closed.

**Why it matters:** Potential unhandled error on stdin.

---

**3. `scripts/hermes/jobs/codex-issue-shipper.ts` — `systemCapacity` uses `freemem()` which returns bytes, divides by `1024 / 1024`**

`freemem() / 1024 / 1024` is correct for bytes → MiB. The variable is named `freeMemoryMb` but is actually MiB. Minor naming inconsistency.

**Why it matters:** Cosmetic — the 4096 default is in MiB, so the behavior is correct, but the naming could confuse future maintainers.

---

### Cross-File Impact

**1. `scripts/hermes/lib/heavy-job-lock.ts` — `withHeavyJobLock` now accepts `staleMs` option, but existing callers don't pass it**

The `staleMs` parameter is added to `options` with a default of `STALE_MS` (5 minutes) via `reapIfStale(options.staleMs)`. Existing callers of `withHeavyJobLock` that don't pass `staleMs` will continue to use the 5-minute default. This is backward-compatible.

**Verdict:** No breaking change.

---

**2. `scripts/hermes/lib/codex-issue-shipper.ts` — `ShipperConfig` interface gains four new required fields**

Any code that constructs a `ShipperConfig` object manually (e.g., tests) must now provide `maxParallelAgents`, `minFreeMemoryMb`, `maxLoadPerCpu`, and `singletonLockStaleMs`. The test file already adds these fields to its inline config objects, but any other test or consumer that constructs `ShipperConfig` directly will break at typecheck time.

**Why it matters:** TypeScript will catch this at compile time, so it's not a runtime risk. But any external consumers of `ShipperConfig` (if any exist outside this repo) would need updates.

---

**3. `scripts/hermes/jobs/codex-issue-shipper.ts` — `runAgent` signature changed from sync to async (`Promise<AgentRunResult>`)**

All callers must now `await` the result. The diff shows the single call site in `dispatchPlan` is updated. No other callers exist in this diff.

**Verdict:** No cross-file break — all callers updated.

---

### Summary

|---|---|---|
| Logic errors | 6 | 2 high, 2 medium, 2 low |
| Security | 1 | 1 low |
| Data integrity | 0 | — |
| Error handling | 3 | 2 medium, 1 low |
| Cross-file impact | 3 | 0 blocking (TS catches at compile) |

**Highest priority items to address:**
1. **stdin race in `runAgent`** — move `child.stdin.end(prompt)` inside the Promise, guard with try/catch
2. **Lock released while agents in-flight** — `tryWithHeavyJobLock`'s `finally` releases the lock when the drain loop exits, but spawned agents may still be running
3. **`branchExists` network timeout silently treated as "doesn't exist"** — could cause branch name collision on network failure

itstimwhite (commented):

itstimwhite:

Branch: 'codex/shipper-singleton-parallel-drain'
State: BLOCKED



To resolve:
1. git fetch origin main
2. git checkout codex/shipper-singleton-parallel-drain
3. git rebase origin/main
5. git push --force-with-lease origin codex/shipper-singleton-parallel-drain

itstimwhite:

### Logic Errors

**1. `scripts/hermes/jobs/codex-issue-shipper.ts` — `runAgent` timeout races with `child.stdin.end`**

The `timeout` is set via `setTimeout` but the old code passed `timeout` directly to `spawnSync`. In the new async version, the timeout fires and calls `child.kill('SIGTERM')`, but `child.stdin.end(promise)` is called *after* the `spawn()` call. If the child process exits (or errors) before `child.stdin.end(prompt)` executes, `stdin.end` will emit an error on a closed stream. The `child.on('error')` handler handles this, but the error message will be misleading ("write EPIPE" or similar) rather than indicating the real cause.

**Why it matters:** Operators debugging a failed agent run will see a confusing stdin error instead of the actual agent failure.

**Suggested fix:** Guard the `child.stdin.end()` call:

```typescript
if (!child.stdin.destroyed) {
  child.stdin.end(prompt);
}
```

---

**2. `scripts/hermes/jobs/codex-issue-shipper.ts` — `dispatchPlan` uses `plan.issue.number` in the `missing_pr_after_success` log instead of `dispatch.issue.number`**

At the `missing_pr_after_success` log event, the code logs `issue: plan.issue.number` (line ~760 in the new file) but the branch being checked is `dispatch.branchName`. Since `dispatch` may have a mutated `branchName` (appended timestamp slug from `prepareWorktree`), the issue number logged is correct but inconsistent with the branch name. This is minor but could confuse correlation.

**Why it matters:** Low severity — log correlation during incident response.

**Suggested fix:** Use `dispatch.issue.number` for consistency.

---

**3. `scripts/hermes/jobs/codex-issue-shipper.ts` — `dispatchBatch` swallows errors and marks issues blocked, but `dispatchPlan` also has a `finally` that calls `cleanupWorktree`**

If `dispatchPlan` throws *after* `prepareWorktree` succeeds but *before* `claimIssue` runs, `cleanupWorktree` will still run — that's correct. But if `dispatchPlan` throws synchronously before `prepared` is assigned (e.g., if `prepareWorktree` itself throws), the `finally` block references `prepared` which is in the TDZ (Temporal Dead Zone). This would produce a `ReferenceError` instead of the original error.

Looking more carefully: `prepared` is declared with `const` inside the `try` block scope. The `finally` block at the `dispatchPlan` level references `prepared.repoRoot`. If `prepareWorktree` throws, `prepared` is never assigned, and the `finally` block will throw a `ReferenceError`, masking the original error.

**Why it matters:** If worktree preparation fails (e.g., disk full, git error), the original error is swallowed by a `ReferenceError` in the `finally` block.

**Suggested fix:** Move `prepared` declaration before the `try`:

```typescript
let prepared: { readonly plan: DispatchPlan; readonly repoRoot: string };
try {
  prepared = prepareWorktree(config, plan);
  // ...
} finally {
  cleanupWorktree(config, prepared.repoRoot);
}
```

Or guard the `finally`:

```typescript
} finally {
  if (prepared) cleanupWorktree(config, prepared.repoRoot);
}
```

---

**4. `scripts/hermes/jobs/codex-issue-shipper.ts` — `branchExists` uses `spawnSyncSafe` but `spawnSyncSafe` uses `execFileSync` internally — naming inconsistency**

`spawnSyncSafe` calls `execFileSync`, not `spawnSync`. The function name says "spawn" but the implementation uses "exec". This is a naming bug, not a functional bug, but it will confuse future readers who expect `spawnSync` semantics (e.g., no shell, different stdio defaults).

**Why it matters:** Low — naming clarity only, since `execFileSync` with `stdio: 'ignore'` behaves equivalently here.

---

**5. `scripts/hermes/jobs/codex-issue-shipper.ts` — `systemCapacity` applies memory and load gates independently, then applies the "too high" gate, but the `Math.min(allowedAgents, 1)` calls can reduce `allowedAgents` to 1 before the "too high" check reduces it to 0**

This is actually correct behavior — the gates are cumulative and the final "too high" check wins. No bug here on closer inspection.

---

**6. `scripts/hermes/jobs/codex-issue-shipper.ts` — Drain loop has no exit guard for persistent dispatchable issues that always fail**

The `for (;;)` loop re-fetches issues and re-dispatches every iteration. If an issue is always dispatchable but always fails (e.g., a poisoned issue that crashes the agent), the loop will run forever, creating worktrees and consuming resources indefinitely.

**Why it matters:** A single bad issue can spin the shipper indefinitely, filling disk with worktrees and burning CI minutes.

**Suggested fix:** Add a maximum iteration count or a cooldown for repeatedly-failed issues. At minimum, log a warning after N consecutive failures of the same issue number.

---

**7. `scripts/hermes/jobs/codex-issue-shipper.ts` — `ensureControlLabels` is called once per drain loop, but `listCodexIssues` is called every iteration. If new issues appear mid-drain with new labels, those labels won't be ensured until the next cron run.**

This is a minor edge case — the control labels are created on the first iteration, and new issues mid-drain would use existing labels. Only truly new label types would be missed. Acceptable for most cases.

---

### Security

**1. `scripts/hermes/jobs/codex-issue-shipper.ts` — Worktree path uses `tmpdir()` without sanitization**

`repoRoot` is constructed as `join(tmpdir(), 'jovie-worktrees')` and then `join(root, 'gh-${plan.issue.number}-${base}')`. The issue number comes from GitHub API data. While issue numbers are integers (safe), the `plan.branchName` is also interpolated into the worktree path via `prepareWorktree`. If a branch name contains path traversal characters (e.g., `../../etc`), the `repoRoot` could land outside the intended directory.

**Why it matters:** A maliciously crafted branch name from a GitHub issue could cause worktrees to be created outside the expected directory.

**Suggested fix:** Sanitize `plan.branchName` before using it in path construction, or validate that the resolved `repoRoot` is under the expected root.

---

**2. `scripts/hermes/jobs/codex-issue-shipper.ts` — `cleanupWorktree` falls back to `rmSync` on failure**

If `git worktree remove --force` fails, the fallback `rmSync(repoRoot, { recursive: true, force: true })` will recursively delete whatever is at that path. Combined with the path traversal risk above, this could delete arbitrary directories.

**Why it matters:** Compounds issue #1 — a path traversal in branch name + cleanup fallback = potential data loss.

**Suggested fix:** Validate `repoRoot` is under the expected worktree root before `rmSync`.

---

### Data Integrity

**1. `scripts/hermes/lib/codex-issue-shipper.ts` — `parsePositiveFloat` accepts `NaN` edge case**

`Number.parseFloat('')` returns `NaN`. The check `Number.isFinite(parsed) && parsed > 0` correctly rejects `NaN` (since `Number.isFinite(NaN)` is `false`). No bug here.

---

**2. `scripts/hermes/lib/heavy-job-lock.ts` — `tryWithHeavyJobLock` releases the lock in `finally`, but the drain loop inside may run for hours**

The lock is held for the entire duration of the `for (;;)` loop. The `staleMs` is set to 8 hours (`28800000`), which matches. But if the loop runs longer than 8 hours (e.g., many issues with long agent timeouts), another cron invocation could acquire the lock after the stale check passes, leading to two concurrent shippers.

**Why it matters:** The `singletonLockStaleMs` default of 8 hours is a safety valve, not a guarantee. If agent timeouts are configured high and the queue is deep, two shippers could overlap.

**Suggested fix:** Consider renewing the lock periodically during the drain loop, or document that `agentTimeoutMs * maxIssuesPerRun` must stay well under `singletonLockStaleMs`.

---

### Error Handling

**1. `scripts/hermes/jobs/codex-issue-shipper.ts` — `child.stdin.end(prompt)` called unconditionally after `spawn()`**

As noted in Logic Error #1, if the child errors or exits before `stdin.end`, this will emit an unhandled error on the stdin stream. While `child.on('error')` catches the child-level error, the stdin write error is a separate event.

**Why it matters:** Unhandled stream errors can crash the process with an unhandled exception.

---

**2. `scripts/hermes/jobs/codex-issue-shipper.ts` — `dispatchPlan` `finally` block TDZ risk**

As noted in Logic Error #3 — if `prepareWorktree` throws, `prepared` is in TDZ and the `finally` block throws `ReferenceError`.

---

**3. `scripts/hermes/jobs/codex-issue-shipper.ts` — `runAgent` does not handle `stdin.write` returning false (backpressure)**

The prompt is written via `child.stdin.end(prompt)`. For very large prompts, this could cause backpressure issues. In practice, agent prompts are typically small enough that this is unlikely to matter, but it's worth noting.

**Why it matters:** Low — prompts are typically well under the buffer limit.

---

### Cross-File Impact

**1. `scripts/hermes/lib/heavy-job-lock.ts` — `withHeavyJobLock` now accepts `staleMs` option, which is backward-compatible (optional with default). No breaking change.**

**2. `scripts/hermes/lib/codex-issue-shipper.ts` — `ShipperConfig` interface gains four new required fields. Any caller constructing a `ShipperConfig` manually (e.g., tests) must provide these fields.**

The test file at `scripts/lib/__tests__/codex-issue-shipper.test.mjs` already adds these fields to its inline config (lines ~270-273 in the diff). But any other test or script that constructs a `ShipperConfig` without these fields will fail typechecking.

**Why it matters:** If there are other test files or scripts that construct `ShipperConfig` objects, they'll break.

**Suggested fix:** Grep for all `ShipperConfig` constructions across the codebase to ensure none are missed.

---

### Summary

|---|---|---|
| Logic errors | 2 high, 2 medium, 2 low | TDZ in `finally` and infinite drain loop are the most actionable |
| Security | 2 medium | Path traversal in worktree creation is the most actionable |
| Data integrity | 1 medium | Lock duration vs. stale timeout edge case |
| Error handling | 2 medium | stdin error on exited child, TDZ (same as logic #3) |
| Cross-file impact | 1 low | Config interface change — verify all callers |

itstimwhite (commented):

PR #10944, opened 2026-06-15
fix(web): allow screenshot fixtures in local automation

itstimwhite:
## Summary
- Allows the exact `/exp/shell-v1` Product Screenshots fixture path during loopback E2E automation only.
- Keeps public production `/exp/*` debug/test routes blocked, including `/exp/library-v1` and public-host requests with test env enabled.
- Lets the `/exp` layout honor the same local automation gate so production-built screenshot servers can render the fixture instead of a 404.

## Bug-to-Test
bug-to-test: satisfied — regression tests added in:
- `apps/web/tests/unit/lib/security/development-only.test.ts`
- `apps/web/tests/unit/lib/security/production-blocked-routes.test.ts`
- `apps/web/tests/unit/middleware/proxy-debug-route-lockdown.test.ts`

## Test Coverage
New code paths covered:
- E2E opt-in + loopback host allows local screenshot automation.
- Missing E2E opt-in denies localhost automation.
- Public host denies screenshot fixture even with E2E env enabled.
- Only `/exp/shell-v1` is allowlisted for product screenshot capture.
- Proxy keeps other `/exp` routes blocked on local automation.

## Pre-Landing Review
Pre-Landing Review: No issues found.

## Design Review
No visual UI changes. Layout guard only.

## Eval Results
No prompt-related files changed — evals skipped.

## Linear follow-ups
Linear follow-ups: none.

## Test plan
- [x] `JOVIE_AGENT_PROFILE=coder pnpm biome check --write apps/web/app/exp/layout.tsx apps/web/lib/security/development-only.ts apps/web/lib/security/production-blocked-routes.ts apps/web/lib/security/require-development-only.ts apps/web/proxy.ts apps/web/tests/unit/lib/security/development-only.test.ts apps/web/tests/unit/lib/security/production-blocked-routes.test.ts apps/web/tests/unit/middleware/proxy-debug-route-lockdown.test.ts` — checked 8 files, fixed 1
- [x] `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/web exec vitest run tests/unit/lib/security/development-only.test.ts tests/unit/lib/security/production-blocked-routes.test.ts tests/unit/middleware/proxy-debug-route-lockdown.test.ts` — 3 files, 38 tests passed
- [x] `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/web run typecheck -- --pretty false` — passed
- [x] `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/web run test:bug-to-test` — not applicable by branch metadata; regression tests are included above
- [x] Guard repro: `node scripts/version-check.mjs && node scripts/desktop-release-guard.mjs --base origin/main && node --test scripts/desktop-release-guard.test.mjs scripts/dev-web-fast.test.mjs && node apps/web/scripts/next-proxy-guard.mjs` — passed after removing the unnecessary version bump
- [x] `JOVIE_AGENT_PROFILE=coder DATABASE_URL=postgresql://localhost/noop NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=pk_test_mock CLERK_SECRET_KEY=sk_test_mock NEXT_DISABLE_TOOLBAR=1 NEXT_PUBLIC_CLERK_PROXY_DISABLED=1 E2E_USE_TEST_AUTH_BYPASS=1 E2E_CLERK_USER_ID=user_test pnpm --filter @jovie/web run build` — passed on final source diff before version-bump removal
- [x] Production server + targeted screenshot: `BASE_URL=http://localhost:3000 pnpm --filter @jovie/web exec playwright test --config=playwright.config.screenshots.ts --project=screenshots -g "captures shell-v1-dashboard-desktop$"` — 1 passed in 24.5s
- [x] Production server + full screenshot catalog: `BASE_URL=http://localhost:3000 pnpm --filter @jovie/web exec playwright test --config=playwright.config.screenshots.ts --project=screenshots` — 58 passed in 16.7m
- [x] Pre-push hook — `biome check .`, `next:proxy-guard`, `tailwind:check`, `typecheck`, and `biome lint .` passed

### DIFF
--- a/apps/web/app/exp/layout.tsx
+++ b/apps/web/app/exp/layout.tsx
@@ -1,7 +1,9 @@
 import { TooltipProvider } from '@jovie/ui';
 import type { Metadata } from 'next';
+import { headers } from 'next/headers';
 import type { ReactNode } from 'react';
 import { QueryProvider } from '@/components/providers/QueryProvider';
+import { isLocalDevelopmentAutomationRequest } from '@/lib/security/development-only';
 import { requireDevelopmentOnlyPage } from '@/lib/security/require-development-only';
 import { NOINDEX_ROBOTS } from '@/lib/seo/noindex-metadata';
 
@@ -12,12 +14,16 @@ export const metadata: Metadata = {
 // /exp/* routes share the production QueryClient + Tooltip context so
 // shipped components (Variant F ChatInput, etc.) work as-is when we
 // drop them into the design pass.
-export default function ExpLayout({
+export default async function ExpLayout({
   children,
 }: {
   readonly children: ReactNode;
 }) {
-  requireDevelopmentOnlyPage();
+  const headerStore = await headers();
+  requireDevelopmentOnlyPage({
+    allowLocalDevelopmentAutomation:
+      isLocalDevelopmentAutomationRequest(headerStore),
+  });
 
   return (
     <QueryProvider>
--- a/apps/web/lib/security/development-only.ts
+++ b/apps/web/lib/security/development-only.ts
@@ -6,6 +6,10 @@ export const DEVELOPMENT_ONLY_ERROR = 'Not available outside development';
 
 const NO_STORE_HEADERS = { 'Cache-Control': 'no-store' } as const;
 
+interface HeaderReader {
+  get(name: string): string | null;
+}
+
 /**
  * ALLOW gate for dev/test/debug routes.
  *
@@ -26,6 +30,52 @@ export function isExplicitDevelopmentEnvironment(): boolean {
 /** Hard production deploy check used for defence-in-depth proxy blocks. */
 export function isVercelProductionDeployment(): boolean {
   return process.env.VERCEL_ENV === 'production';
+}
+
+function extractHostname(value: string | null): string | null {
+  const normalized = value?.split(',')[0]?.trim();
+  if (!normalized) return null;
+
+  if (normalized.startsWith('http://') || normalized.startsWith('https://')) {
+    try {
+      return new URL(normalized).hostname.toLowerCase();
+    } catch {
+      return null;
+    }
+  }
+
+  return normalized.replace(/:\d+$/, '').toLowerCase();
+}
+
+export function isLocalDevelopmentAutomationHostname(
+  hostname: string | null
+): boolean {
+  const normalized = hostname?.trim().toLowerCase() ?? null;
+  return (
+    normalized === 'localhost' ||
+    normalized === '127.0.0.1' ||
+    normalized === '::1' ||
+    normalized === '[::1]' ||
+    Boolean(normalized?.endsWith('.localhost'))
+  );
+}
+
+/**
+ * Local-only automation escape hatch for production-built test servers.
+ * This intentionally requires an existing E2E opt-in and a loopback host.
+ */
+export function isLocalDevelopmentAutomationRequest(
+  headerReader: HeaderReader
+): boolean {
+  if (process.env.E2E_USE_TEST_AUTH_BYPASS !== '1') return false;
+
+  const hostname =
+    extractHostname(headerReader.get('x-forwarded-host')) ??
+    extractHostname(headerReader.get('host')) ??
+    extractHostname(headerReader.get('origin')) ??
+    extractHostname(headerReader.get('referer'));
+
+  return isLocalDevelopmentAutomationHostname(hostname);
 }
 
 export function developmentOnlyForbiddenResponse(
--- a/apps/web/lib/security/production-blocked-routes.ts
+++ b/apps/web/lib/security/production-blocked-routes.ts
@@ -28,6 +28,13 @@ export const PRODUCTION_BLOCKED_PAGE_EXACT = [
 ] as const;
 
 /**
+ * Production builds used by the Product Screenshots workflow still capture a
+ * few legacy experiment fixtures. Keep the inventory exact so the proxy can
+ * allow only screenshot automation without reopening all /exp routes.
+ */
+export const PRODUCT_SCREENSHOT_CAPTURE_PAGE_PATHS = ['/exp/shell-v1'] as const;
+
+/**
  * Routes that intentionally stay reachable outside development.
  * Keep this list tiny and justify every entry in code review.
  */
@@ -47,6 +54,12 @@ export function isProxyAllowlistedDevelopmentRoute(pathname: string): boolean {
   );
 }
 
+export function isProductScreenshotCapturePath(pathname: string): boolean {
+  return PRODUCT_SCREENSHOT_CAPTURE_PAGE_PATHS.some(
+    allowed => pathname === allowed
+  );
+}
+
 function matchesRoutePrefix(pathname: string, prefix: string): boolean {
   if (pathname === prefix) {
     return true;
@@ -59,8 +72,22 @@ function matchesRoutePrefix(pathname: string, prefix: string): boolean {
   return pathname.startsWith(`${prefix}/`);
 }
 
-export function isProductionBlockedDebugPath(pathname: string): boolean {
+interface ProductionBlockedDebugPathOptions {
+  readonly allowProductScreenshotCaptureRoutes?: boolean;
+}
+
+export function isProductionBlockedDebugPath(
+  pathname: string,
+  options: ProductionBlockedDebugPathOptions = {}
+): boolean {
   if (isProxyAllowlistedDevelopmentRoute(pathname)) {
+    return false;
+  }
+
+  if (
+    options.allowProductScreenshotCaptureRoutes === true &&
+    isProductScreenshotCapturePath(pathname)
+  ) {
     return false;
   }
 
--- a/apps/web/lib/security/require-development-only.ts
+++ b/apps/web/lib/security/require-development-only.ts
@@ -3,9 +3,18 @@ import 'server-only';
 import { notFound } from 'next/navigation';
 import { isExplicitDevelopmentEnvironment } from '@/lib/security/development-only';
 
+interface RequireDevelopmentOnlyPageOptions {
+  readonly allowLocalDevelopmentAutomation?: boolean;
+}
+
 /** Server component/layout guard for dev-only pages. */
-export function requireDevelopmentOnlyPage(): void {
-  if (!isExplicitDevelopmentEnvironment()) {
+export function requireDevelopmentOnlyPage(
+  options: RequireDevelopmentOnlyPageOptions = {}
+): void {
+  if (
+    !isExplicitDevelopmentEnvironment() &&
+    options.allowLocalDevelopmentAutomation !== true
+  ) {
     notFound();
   }
 }
--- a/apps/web/proxy.ts
+++ b/apps/web/proxy.ts
@@ -49,7 +49,11 @@ import {
   isProxyRewriteExempt,
 } from '@/lib/routing/proxy-routing';
 import { SCRIPT_NONCE_HEADER } from '@/lib/security/content-security-policy';
-import { isExplicitDevelopmentEnvironment } from '@/lib/security/development-only';
+import {
+  isExplicitDevelopmentEnvironment,
+  isLocalDevelopmentAutomationHostname,
+  isLocalDevelopmentAutomationRequest,
+} from '@/lib/security/development-only';
 import {
   createFastNotFoundResponse,
   createProbeDropResponse,
@@ -106,6 +110,17 @@ function isElectronAppShellNavigation(
     (req.nextUrl.pathname === APP_ROUTES.DASHBOARD ||
       req.nextUrl.pathname.startsWith('/app/')) &&
     req.nextUrl.searchParams.get('runtime') === 'electron'
+  );
+}
+
+function shouldAllowProductScreenshotCaptureRoutes(req: NextRequest): boolean {
+  if (isLocalDevelopmentAutomationRequest(req.headers)) {
+    return true;
+  }
+
+  return (
+    process.env.E2E_USE_TEST_AUTH_BYPASS === '1' &&
+    isLocalDevelopmentAutomationHostname(req.nextUrl.hostname)
   );
 }
 
@@ -256,7 +271,10 @@ async function handleRequest(req: NextRequest, userId: string | null) {
     // Block debug/test/dev surfaces outside explicit development environments.
     if (
       !isExplicitDevelopmentEnvironment() &&
-      isProductionBlockedDebugPath(pathname)
+      isProductionBlockedDebugPath(pathname, {
+        allowProductScreenshotCaptureRoutes:
+          shouldAllowProductScreenshotCaptureRoutes(req),
+      })
     ) {
       return NextResponse.rewrite(new URL('/404', req.url));
     }
--- a/apps/web/tests/unit/lib/security/development-only.test.ts
+++ b/apps/web/tests/unit/lib/security/development-only.test.ts
@@ -68,4 +68,55 @@ describe('development-only security helpers', () => {
       });
     });
   });
+
+  describe('isLocalDevelopmentAutomationRequest', () => {
+    it('classifies loopback hostnames for local automation', async () => {
+      const { isLocalDevelopmentAutomationHostname } = await import(
+        '@/lib/security/development-only'
+      );
+
+      expect(isLocalDevelopmentAutomationHostname('localhost')).toBe(true);
+      expect(isLocalDevelopmentAutomationHostname('preview.localhost')).toBe(
+        true
+      );
+      expect(isLocalDevelopmentAutomationHostname('jov.ie')).toBe(false);
+      expect(isLocalDevelopmentAutomationHostname(null)).toBe(false);
+    });
+
+    it('allows E2E automation on localhost', async () => {
+      vi.stubEnv('E2E_USE_TEST_AUTH_BYPASS', '1');
+      const { isLocalDevelopmentAutomationRequest } = await import(
+        '@/lib/security/development-only'
+      );
+
+      expect(
+        isLocalDevelopmentAutomationRequest(
+          new Headers({ host: 'localhost:3000' })
+        )
+      ).toBe(true);
+    });
+
+    it('rejects localhost without explicit E2E opt-in', async () => {
+      const { isLocalDevelopmentAutomationRequest } = await import(
+        '@/lib/security/development-only'
+      );
+
+      expect(
+        isLocalDevelopmentAutomationRequest(
+          new Headers({ host: 'localhost:3000' })
+        )
+      ).toBe(false);
+    });
+
+    it('rejects public hosts even with explicit E2E opt-in', async () => {
+      vi.stubEnv('E2E_USE_TEST_AUTH_BYPASS', '1');
+      const { isLocalDevelopmentAutomationRequest } = await import(
+        '@/lib/security/development-only'
+      );
+
+      expect(
+        isLocalDevelopmentAutomationRequest(new Headers({ host: 'jov.ie' }))
+      ).toBe(false);
+    });
+  });
 });
--- a/apps/web/tests/unit/lib/security/production-blocked-routes.test.ts
+++ b/apps/web/tests/unit/lib/security/production-blocked-routes.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from 'vitest';
 import {
   isProductionBlockedDebugPath,
+  PRODUCT_SCREENSHOT_CAPTURE_PAGE_PATHS,
   PRODUCTION_BLOCKED_API_PREFIXES,
   PRODUCTION_BLOCKED_PAGE_PREFIXES,
 } from '@/lib/security/production-blocked-routes';
@@ -22,6 +23,10 @@ describe('production-blocked debug routes', () => {
     );
   });
 
+  it('keeps the product screenshot capture inventory exact', () => {
+    expect(PRODUCT_SCREENSHOT_CAPTURE_PAGE_PATHS).toEqual(['/exp/shell-v1']);
+  });
+
   it.each([
     '/api/dev/test-auth/session',
     '/api/dev/clear-session',
@@ -35,6 +40,24 @@ describe('production-blocked debug routes', () => {
     '/sentry-example-page',
   ])('blocks %s outside development', route => {
     expect(isProductionBlockedDebugPath(route)).toBe(true);
+  });
+
+  it('allows only explicit product screenshot fixture routes when requested', () => {
+    expect(
+      isProductionBlockedDebugPath('/exp/shell-v1', {
+        allowProductScreenshotCaptureRoutes: true,
+      })
+    ).toBe(false);
+    expect(
+      isProductionBlockedDebugPath('/exp/shell-v1/other', {
+        allowProductScreenshotCaptureRoutes: true,
+      })
+    ).toBe(true);
+    expect(
+      isProductionBlockedDebugPath('/exp/library-v1', {
+        allowProductScreenshotCaptureRoutes: true,
+      })
+    ).toBe(true);
   });
 
   it.each([
--- a/apps/web/tests/unit/middleware/proxy-debug-route-lockdown.test.ts
+++ b/apps/web/tests/unit/middleware/proxy-debug-route-lockdown.test.ts
@@ -194,6 +194,52 @@ describe('proxy debug/test route lockdown', () => {
     expect(new URL(rewriteUrl!).pathname).toBe('/404');
   });
 
+  it('allows shell-v1 product screenshots on loopback automation', async () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    vi.stubEnv('VERCEL_ENV', 'production');
+    vi.stubEnv('E2E_USE_TEST_AUTH_BYPASS', '1');
+
+    const req = createTestRequest({
+      pathname: '/exp/shell-v1',
+      hostname: 'localhost',
+    });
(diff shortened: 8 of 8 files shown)


itstimwhite:

Branch: 'codex/jov-3216-product-screenshots-main-green'
State: BLOCKED



To resolve:
1. git fetch origin main
2. git checkout codex/jov-3216-product-screenshots-main-green
3. git rebase origin/main
5. git push --force-with-lease origin codex/jov-3216-product-screenshots-main-green

itstimwhite:

### Logic Errors

**1. `development-only.ts:37` — `extractHostname` parses `origin`/`referer` headers through URL constructor, but those headers may contain full URLs with paths. The function returns only the hostname, which is correct, but the `origin` header value like `http://localhost:3000/some/path` would fail `new URL()` only if malformed. This is fine in practice, but the function silently returns `null` on any `catch`, swallowing parse errors silently. Not a bug per se, but worth noting the `catch` block is overly broad — it catches all exceptions, not just `TypeError` from `new URL()`.**

*Why it matters:* If `new URL()` throws for an unexpected reason (e.g., an out-of-memory error in an exotic runtime), the silent catch would mask it. In practice this is extremely unlikely, so this is informational only.

**Verdict:** No actionable logic bug. The null-return-on-fail pattern is intentional and consistent with the codebase.

---

**2. `production-blocked-routes.ts:77` — `isProductScreenshotCapturePath` uses `pathname === allowed` (exact match only). The test at line 42–50 of the test file confirms `/exp/shell-v1/other` is correctly blocked. This is correct and intentional. No bug.**

**Verdict:** None found.

---

**3. `proxy.ts:116` — `shouldAllowProductScreenshotCaptureRoutes` has two code paths that both check `E2E_USE_TEST_AUTH_BYPASS === '1'` and loopback hostname. The first path uses `isLocalDevelopmentAutomationRequest(req.headers)` which internally checks `E2E_USE_TEST_AUTH_BYPASS`. The second path checks `E2E_USE_TEST_AUTH_BYPASS` again directly. This means if `isLocalDevelopmentAutomationRequest` returns `true`, the second check is unreachable. If it returns `false` (e.g., non-loopback host), the second check would also fail because `isLocalDevelopmentAutomationHostname(req.nextUrl.hostname)` would also fail. The duplication is logically redundant but not incorrect.**

*Why it matters:* The second code path (`process.env.E2E_USE_TEST_AUTH_BYPASS === '1' && isLocalDevelopmentAutomationHostname(req.nextUrl.hostname)`) is dead code when `isLocalDevelopmentAutomationRequest` already covers the same logic. However, it acts as a defense-in-depth check using `req.nextUrl.hostname` (which is derived by Next.js from the request) rather than parsing headers directly. This is actually a reasonable belt-and-suspenders approach — `req.nextUrl.hostname` is harder to spoof than raw headers.

**Verdict:** Not a bug. Redundant but defensible.

---

**4. `require-development-only.ts:14` — The condition `options.allowLocalDevelopmentAutomation !== true` uses strict `!== true`, meaning any truthy-but-not-exactly-`true` value (e.g., `"yes"`, `1`) would be treated as `false`. This is correct and intentional — it prevents accidental opt-in via truthy env vars or similar.**

**Verdict:** None found.

---

### Security

**1. `development-only.ts:63` — `isLocalDevelopmentAutomationRequest` checks `E2E_USE_TEST_AUTH_BYPASS` env var. This is a server-side env var check, not client-controllable. The hostname is extracted from `x-forwarded-host`, `host`, `origin`, and `referer` headers. The `x-forwarded-host` header can be spoofed by clients in non-proxied setups. However, the function is only called in two places: (a) `proxy.ts` middleware where `req.headers` is the raw Next.js headers object, and (b) `exp/layout.tsx` where `headers()` from `next/headers` is used. In both cases, the hostname check is validated against a strict allowlist of loopback addresses (`localhost`, `127.0.0.1`, `::1`, `[::1]`, `*.localhost`). A spoofed `x-forwarded-host: evil.com` would fail the hostname check. The only risk is if an attacker can set `x-forwarded-host: localhost` — but this would only matter if `E2E_USE_TEST_AUTH_BYPASS=1` is also set server-side, which is an operator-controlled env var.**

*Why it matters:* If an operator accidentally deploys with `E2E_USE_TEST_AUTH_BYPASS=1` in production, a client could spoof `x-forwarded-host: localhost` to access `/exp/shell-v1`. This is a defense-in-depth concern — the env var should never be set in production. The PR description explicitly states this is for "local automation only."

**Suggested mitigation:** Consider also checking `req.ip` or `x-real-ip` against `127.0.0.1`/`::1` as an additional signal, or document clearly that `E2E_USE_TEST_AUTH_BYPASS` must never be set in production deployments.

**Severity:** Low (requires operator error to exploit).

---

**2. `development-only.ts:44` — `isLocalDevelopmentAutomationHostname` accepts `*.localhost` via `endsWith('.localhost')`. This is correct for local development (e.g., `preview.localhost`). However, `localhost` itself does not end with `.localhost`, so it's handled by the explicit `=== 'localhost'` check. The string `[::1]` is handled separately. No issue.**

**Verdict:** None found.

---

**3. `production-blocked-routes.ts:33` — `PRODUCT_SCREENSHOT_CAPTURE_PAGE_PATHS` is a `const` array with only `/exp/shell-v1`. The test asserts the exact contents. This is correct and minimal.**

**Verdict:** None found.

---

### Data Integrity

**Verdict:** None found. No DB queries, migrations, or data mutations in this PR.

---

### Error Handling

**1. `development-only.ts:40` — `extractHostname` has a bare `catch {}` that returns `null`. This is intentional for graceful degradation when `new URL()` receives an invalid URL string. No unhandled promise rejections or thrown errors.**

**Verdict:** None found.

---

### Cross-File Impact

**1. `require-development-only.ts` — The function signature changed from `requireDevelopmentOnlyPage()` to `requireDevelopmentOnlyPage(options?)`. The `options` parameter is optional with a default of `{}`, so existing callers that don't pass options continue to work identically. The diff shows only `exp/layout.tsx` passes the new option. No other callers are affected.**

**Verdict:** None found. Backward compatible.

**2. `isProductionBlockedDebugPath` — The function signature changed to accept an optional `options` parameter. The diff shows only `proxy.ts` passes the new option. The existing callers (if any) that don't pass options get the default `{}`, which preserves existing behavior (`allowProductScreenshotCaptureRoutes` is `undefined`, so the `=== true` check fails and the old behavior is preserved).**

**Verdict:** None found. Backward compatible.

---

### Summary

| Category | Issues Found |
|---|---|
| Logic Errors | None actionable |
| Security | 1 low-severity defense-in-depth concern (env var + header spoofing) |
| Data Integrity | None |
| Error Handling | None |
| Cross-File Impact | None |

**Overall assessment:** The PR is logically sound, backward compatible, and well-tested. The only security consideration is that `E2E_USE_TEST_AUTH_BYPASS=1` combined with header spoofing could theoretically allow access to `/exp/shell-v1`, but this requires operator error (setting the env var in production) and is mitigated by the strict loopback hostname allowlist.

itstimwhite (commented):

itstimwhite:

### Logic Errors

**1. `development-only.ts:37` — `extractHostname` silently returns `null` on malformed URLs**
The `catch` block swallows all errors, including cases where a crafted header value is not a valid URL. This is acceptable for security (fail-closed), but the function also silently returns `null` for values like `http://` with no host, which could mask misconfigured reverse proxies. Not a bug per se, but worth noting: the function is used in a security gate, so silent failure is the correct behavior here. **No change needed.**

**2. `development-only.ts:55` — `isLocalDevelopmentAutomationHostname` doesn't strip port before `.localhost` suffix check**
Input like `evil.com:3000.localhost` would pass the `endsWith('.localhost')` check. However, this function receives its input from `extractHostname`, which already strips ports via `replace(/:\d+$/, '')`. So in the normal call path this is safe. The function is also exported and could be called directly by future consumers with untrusted input. **Low risk today, but worth a comment or input normalization.**

**3. `production-blocked-routes.ts:72` — `isProductionBlockedDebugPath` options parameter defaults to `{}`**
When `options.allowProductScreenshotCaptureRoutes` is `undefined`, the check `=== true` correctly evaluates to `false`. This is safe. **No issue.**

**4. `proxy.ts:113` — `shouldAllowProductScreenshotCaptureRoutes` has two redundant code paths**
The function checks `isLocalDevelopmentAutomationRequest(req.headers)` first, which already checks `E2E_USE_TEST_AUTH_BYPASS === '1'` AND loopback hostname. The fallback then checks `E2E_USE_TEST_AUTH_BYPASS === '1' && isLocalDevelopmentAutomationHostname(req.nextUrl.hostname)` — which is a strict subset of the first check. The second branch is dead code: if `isLocalDevelopmentAutomationRequest` returns `false`, it's either because the env var is missing (so the fallback also fails) or the hostname isn't loopback (but `req.nextUrl.hostname` would also not be loopback in that case, unless headers disagree with `nextUrl`). 

This is **not a bug**, but it is dead/redundant code that adds confusion. The second branch would only diverge if `req.headers` and `req.nextUrl.hostname` disagree (e.g., a proxy sets `Host` header to `localhost` but `nextUrl.hostname` reflects the actual deployment hostname). In that edge case, the fallback could incorrectly allow the route. **This is a potential logic inconsistency worth flagging.**

### Security

**1. `development-only.ts:68` — `isLocalDevelopmentAutomationRequest` trusts `X-Forwarded-Host` header**
The function reads `x-forwarded-host` from headers and extracts the hostname. In a production deployment behind a reverse proxy (e.g., Vercel, Cloudflare), this header can be spoofed by clients unless the proxy strips/overwrites it. If an attacker can set `X-Forwarded-Host: localhost` and the env var `E2E_USE_TEST_AUTH_BYPASS=1` is set (which the PR's own build test uses), they could access `/exp/shell-v1` on a production server.

**This is the most significant security concern in the PR.** The env var `E2E_USE_TEST_AUTH_BYPASS` should never be set in production, but the code doesn't enforce that. Consider adding a guard that also checks `isVercelProductionDeployment()` or similar to hard-block this path in production regardless of headers.

**Severity: Medium.** The env var is the primary gate, but defense-in-depth would be prudent.

**2. `proxy.ts:120` — `req.nextUrl.hostname` vs header-based hostname mismatch**
As noted above, `shouldAllowProductScreenshotCaptureRoutes` uses `req.nextUrl.hostname` in the fallback path but `req.headers` in the primary path. These can differ in proxied environments. If `nextUrl.hostname` reflects the canonical hostname (e.g., `jov.ie`) while a spoofed `Host` header says `localhost`, the primary check (`isLocalDevelopmentAutomationRequest`) would pass (trusting headers) while the fallback would fail. This inconsistency means the header-trusting path is the weaker one.

**3. `production-blocked-routes.ts:37` — Exact match only for screenshot paths**
`isProductScreenshotCapturePath` uses `pathname === allowed`, so `/exp/shell-v1/anything` is NOT allowed. This is correct and intentional. **No issue.**

### Data Integrity

**None found.** No DB queries, migrations, or data mutations in this PR.

### Error Handling

**None found.** All new code paths are synchronous checks with clear boolean returns. No async operations or I/O in the new code.

### Cross-File Impact

**1. `require-development-only.ts` — `options` parameter is optional with default `{}`**
Existing callers that don't pass `options` will continue to work identically. **No breakage.**

**2. `isProductionBlockedDebugPath` — new optional `options` parameter**
Existing callers that don't pass `options` get the default `{}`, which preserves existing behavior. **No breakage.**

**3. `exp/layout.tsx` — changed from sync to async component**
This is a layout component. Next.js supports async server components. The `headers()` call is awaited. **No breakage**, but note that `headers()` is a dynamic API that opts the route out of static generation. Since `/exp/*` routes are already blocked in production and are dev-only, this is acceptable.

### Summary

| 1 | **Medium** | `development-only.ts:68` | `isLocalDevelopmentAutomationRequest` trusts `X-Forwarded-Host` header, which can be spoofed if the env var is set. Consider adding a production-environment guard. |
| 2 | **Low** | `proxy.ts:113-122` | `shouldAllowProductScreenshotCaptureRoutes` has a redundant fallback path that could behave inconsistently with the primary path if `req.headers` and `req.nextUrl.hostname` disagree. |
| 3 | **Low** | `development-only.ts:55` | `isLocalDevelopmentAutomationHostname` doesn't normalize input before the `.localhost` suffix check. Safe today due to `extractHostname` pre-processing, but fragile if called directly in the future. |

itstimwhite (commented):

itstimwhite:

### Logic Errors

**1. `development-only.ts:37` — `extractHostname` silently returns `null` on malformed URLs**
The `catch` block in `extractHostname` swallows all errors and returns `null`. This is fine for the current callers (they fall through to the next header), but the function is exported and could be used elsewhere. Not a bug today, but worth noting: a `console.warn` or typed error would help future debuggability. No action required for this PR.

**2. `development-only.ts:55` — `isLocalDevelopmentAutomationHostname` doesn't handle `[::1]` with port**
The function strips trailing `:\d+` but the IPv6 check compares against `'[::1]'` *before* the port strip. If the input is `[::1]:3000`, the `.endsWith('.localhost')` check fails, the explicit `[::1]` check fails (because the string is `[::1]:3000`), and the port strip only applies to the non-URL branch. Actually, looking more carefully: the function receives the *output* of `extractHostname`, which already strips ports via `new URL(...).hostname`. But `isLocalDevelopmentAutomationHostname` is also called directly in `proxy.ts:122` with `req.nextUrl.hostname`, which never has a port. So this is fine. **No issue.**

**3. `production-blocked-routes.ts:79` — `isProductScreenshotCapturePath` uses exact match only**
The function checks `pathname === allowed`, not `pathname.startsWith(allowed + '/')`. This means `/exp/shell-v1` is allowed but `/exp/shell-v1/dashboard` is blocked. This is intentional (confirmed by the test at line 55 of the test file: `/exp/shell-v1/other` is blocked). **No issue — by design.**

**4. `proxy.ts:116` — `shouldAllowProductScreenshotCaptureRoutes` has two redundant checks**
The function checks `isLocalDevelopmentAutomationRequest(req.headers)` first, which already checks `E2E_USE_TEST_AUTH_BYPASS === '1'` AND loopback hostname. The fallback then checks `E2E_USE_TEST_AUTH_BYPASS === '1' && isLocalDevelopmentAutomationHostname(req.nextUrl.hostname)`. The fallback is only reached if `isLocalDevelopmentAutomationRequest` returns false, which means either E2E opt-in is missing OR the hostname extracted from headers is not loopback. The fallback re-checks E2E opt-in (which would already be true if we got past the first check's hostname failure) and uses `req.nextUrl.hostname` instead of header-derived hostname.

This is actually a meaningful difference: if `x-forwarded-host` contains a non-loopback hostname but `req.nextUrl.hostname` is loopback (e.g., behind a reverse proxy that sets the wrong header), the fallback would still allow it. This is a **defense-in-depth** choice, not a bug. **No issue.**

### Security

**1. `development-only.ts:72` — `isLocalDevelopmentAutomationRequest` trusts `x-forwarded-host` header**
An attacker who can set `x-forwarded-host` to `localhost` could bypass the production block on `/exp/shell-v1`. However, this is mitigated by the fact that `E2E_USE_TEST_AUTH_BYPASS` must also be set to `'1'`, which is a server-side env var. In production, this env var should not be set. **Risk: low.** Suggestion: add a comment documenting that `E2E_USE_TEST_AUTH_BYPASS` must never be set in production deployments.

**2. `development-only.ts:55` — `isLocalDevelopmentAutomationHostname` allows `*.localhost` subdomains**
The `.endsWith('.localhost')` check allows any subdomain of `localhost` (e.g., `evil.localhost`). This is consistent with how `.localhost` works in browsers (all subdomains resolve to loopback), but if there's a DNS rebinding scenario or a misconfigured local resolver, this could be a concern. **Risk: very low for local automation.** No action needed.

**3. `production-blocked-routes.ts:33` — `PRODUCT_SCREENSHOT_CAPTURE_PAGE_PATHS` is a static array**
If someone adds a path to this array without realizing it's accessible on production-built servers with `E2E_USE_TEST_AUTH_BYPASS=1`, they could inadvertently expose debug routes. The array is small and the comment is clear. **No issue — well-documented.**

### Data Integrity

**None found.** No DB queries, migrations, or data mutations in this PR.

### Error Handling

**None found.** All new code paths have appropriate fallbacks and the existing error handling patterns are preserved.

### Cross-File Impact

**1. `require-development-only.ts:14` — `allowLocalDevelopmentAutomation` is optional and defaults to `false`**
The change is backward-compatible. Existing callers that don't pass the option will continue to call `notFound()` in non-dev environments. **No issue.**

**2. `exp/layout.tsx:17` — Layout is now `async`**
The layout function is now `async` due to `await headers()`. This is supported by Next.js for server components/layouts. **No issue.**

### Summary

| Category | Issues Found |
|----------|-------------|
| Logic Errors | None |
| Security | 1 low-risk (document env var constraint) |
| Data Integrity | None |
| Error Handling | None |
| Cross-File Impact | None |

**Overall: No blocking issues.** The change is well-scoped, properly gated, and backward-compatible.

itstimwhite (commented):

itstimwhite:

### Logic Errors

**1. `development-only.ts:37` — `extractHostname` doesn't handle `x-forwarded-host` port stripping for bare hostnames**

The function has two branches: one for values starting with `http://`/`https://` (uses `URL` parser), and a fallback that strips trailing `:digits`. The fallback `normalized.replace(/:\d+$/, '')` works for `localhost:3000`, but the `x-forwarded-host` header value is passed through `value?.split(',')[0]?.trim()` *before* reaching `extractHostname`. However, the `host` header value (e.g. `localhost:3000`) goes straight in. The regex replace handles this, so the logic is actually correct — but it's fragile. If `x-forwarded-host` contains a bare hostname with port (e.g. `proxy.internal:8080`), the URL-parse branch won't trigger (no scheme), and the regex strip will remove `:8080` correctly. This is fine.

**Verdict:** No bug, but worth noting the implicit assumption.

---

**2. `development-only.ts:55` — `isLocalDevelopmentAutomationHostname` doesn't strip port before checking `.localhost` suffix**

Input `hostname` is documented as `string | null`, and the function calls `hostname?.trim().toLowerCase()`. The caller (`isLocalDevelopmentAutomationRequest`) passes the *return value of `extractHostname`*, which already strips ports. So within this function's contract, ports are already removed. However, if anyone calls `isLocalDevelopmentAutomationHostname` directly with a raw `host` header value like `localhost:3000`, the `.localhost` suffix check would fail for `preview.localhost:3000` → `preview.localhost:3000` doesn't end with `.localhost`. The function's JSDoc/contract doesn't guarantee port stripping.

**Verdict:** Minor — the function is only used internally with pre-stripped hostnames, but the public export could mislead future callers. Not a bug today.

---

**3. `proxy.ts:116-122` — `shouldAllowProductScreenshotCaptureRoutes` has redundant/divergent logic**

```ts
function shouldAllowProductScreenshotCaptureRoutes(req: NextRequest): boolean {
  if (isLocalDevelopmentAutomationRequest(req.headers)) {
    return true;
  }
  return (
    process.env.E2E_USE_TEST_AUTH_BYPASS === '1' &&
    isLocalDevelopmentAutomationHostname(req.nextUrl.hostname)
  );
}
```

`isLocalDevelopmentAutomationRequest` already checks `E2E_USE_TEST_AUTH_BYPASS === '1'` AND validates the hostname via the same header-reading logic. The second branch repeats the env check and uses `req.nextUrl.hostname` instead of the header-derived hostname. This means:

- If `E2E_USE_TEST_AUTH_BYPASS=1` and the request hits `localhost` directly (no proxy), `isLocalDevelopmentAutomationRequest` returns `true` via the `host` header, and the second branch is never reached. ✓
- If a proxy strips/modifies headers such that `x-forwarded-host`/`host`/`origin`/`referer` are all missing or unparseable, but `req.nextUrl.hostname` is `localhost`, the second branch would still allow it. This is a **broader** check than the first branch.

This creates an inconsistency: the first branch is strict (checks multiple headers), while the second is lenient (only checks `nextUrl.hostname`). If the intent is defense-in-depth, the second branch should also go through `isLocalDevelopmentAutomationRequest` or at minimum use the same hostname extraction logic.

**Why it matters:** If `req.nextUrl.hostname` can be influenced differently than the headers (e.g., in certain proxy configurations), the two checks could diverge, potentially allowing requests that the stricter check would block.

**Suggested fix:** Either remove the second branch entirely (since `isLocalDevelopmentAutomationRequest` already covers the loopback case), or document explicitly why the fallback is needed.

```ts
function shouldAllowProductScreenshotCaptureRoutes(req: NextRequest): boolean {
  return isLocalDevelopmentAutomationRequest(req.headers);
}
```

---

**4. `production-blocked-routes.ts:77-83` — `isProductScreenshotCapturePath` uses exact match only, but comment says "fixtures"**

```ts
export function isProductScreenshotCapturePath(pathname: string): boolean {
  return PRODUCT_SCREENSHOT_CAPTURE_PAGE_PATHS.some(
    allowed => pathname === allowed
  );
}
```

This is an exact match (`===`), so `/exp/shell-v1` passes but `/exp/shell-v1/dashboard` does not. The test at line 57-62 of the test file confirms this is intentional (`/exp/shell-v1/other` → blocked). This is correct for the stated use case.

**Verdict:** No bug. Intentional design.

---

### Security

**5. `development-only.ts:72-82` — `isLocalDevelopmentAutomationRequest` trusts `x-forwarded-host` header**

The function checks headers in priority order: `x-forwarded-host` → `host` → `origin` → `referer`. The `x-forwarded-host` header is client-controllable in many deployment scenarios (anyone can set it in an HTTP request). If a reverse proxy doesn't overwrite this header, an attacker could send `x-forwarded-host: localhost` from a public IP and bypass the production block.

**Why it matters:** In a production deployment behind a proxy that trusts/forward `x-forwarded-host`, this is a route-based SSRF/info-disclosure vector. The attacker wouldn't get the full page (they'd need auth), but they could probe whether `/exp/shell-v1` exists and get content rendered.

**Mitigating factor:** The `E2E_USE_TEST_AUTH_BYPASS` env var must also be set to `'1'`, which should never be set in production. If it's not set, the function short-circuits to `false` before any header inspection. So the actual risk is: someone accidentally deploys with `E2E_USE_TEST_AUTH_BYPASS=1` in production.

**Suggested fix:** Add a secondary guard — also check `isVercelProductionDeployment()` and return `false` if true, making it impossible to accidentally enable this on Vercel production:

```ts
export function isLocalDevelopmentAutomationRequest(
  headerReader: HeaderReader
): boolean {
  if (process.env.E2E_USE_TEST_AUTH_BYPASS !== '1') return false;
  if (isVercelProductionDeployment()) return false;
  // ... rest
}
```

This is a defense-in-depth measure.

---

**6. `proxy.ts:116-122` — `E2E_USE_TEST_AUTH_BYPASS` in production proxy**

Same concern as #5 but from the proxy's perspective. The `shouldAllowProductScreenshotCaptureRoutes` function checks `E2E_USE_TEST_AUTH_BYPASS === '1'` directly in the second branch. If this env var leaks into production, `/exp/shell-v1` becomes accessible from any host (the second branch doesn't validate headers, only `req.nextUrl.hostname`).

Wait — actually the second branch *does* check `isLocalDevelopmentAutomationHostname(req.nextUrl.hostname)`, so it's restricted to loopback. But `req.nextUrl.hostname` in a proxied environment might reflect the internal hostname, not the public one. This needs verification.

**Verdict:** Lower risk due to the hostname check, but the defense-in-depth fix from #5 would eliminate this class of risk entirely.

---

### Data Integrity

None found. No DB queries or migrations in this diff.

### Error Handling

**7. `development-only.ts:40-48` — `extractHostname` silently returns `null` on malformed URLs**

```ts
try {
  return new URL(normalized).hostname.toLowerCase();
} catch {
  return null;
}
```

This is intentional — malformed URLs are treated as unresolvable. The callers all handle `null` via `??` fallback chains. No bug.

**Verdict:** Correct.

### Cross-File Impact

**8. `require-development-only.ts` — `allowLocalDevelopmentAutomation` is optional and defaults to `false`**

The function signature changed from `requireDevelopmentOnlyPage()` to `requireDevelopmentOnlyPage(options?)`. All existing callers that don't pass the option will get `options.allowLocalDevelopmentAutomation !== true` → `undefined !== true` → `true` → the `notFound()` branch executes as before. This is backward-compatible.

**Need to verify:** Are there other callers of `requireDevelopmentOnlyPage` in the codebase beyond `exp/layout.tsx`? If they exist, they're unaffected since the parameter is optional.

**Verdict:** Safe.

**9. `isProductionBlockedDebugPath` — new optional parameter, backward-compatible**

The function signature changed from `isProductionBlockedDebugPath(pathname)` to `isProductionBlockedDebugPath(pathname, options = {})`. Existing callers that don't pass `options` get `{}`, and `options.allowProductScreenshotCaptureRoutes` is `undefined`, which is not `=== true`. Behavior is unchanged.

**Verdict:** Safe.

---

### Summary of Issues

| 1 | Low | `proxy.ts:116-122` | Redundant/divergent logic in `shouldAllowProductScreenshotCaptureRoutes` — second branch is broader than first |
| 2 | Medium | `development-only.ts:72-82` | `x-forwarded-host` is client-controllable; if `E2E_USE_TEST_AUTH_BYPASS=1` accidentally set in production, enables route probing |
| 3 | Low | `development-only.ts:55` | `isLocalDevelopmentAutomationHostname` doesn't self-strip ports; relies on caller contract |

**Recommended fixes:**
- **Issue #1:** Simplify `shouldAllowProductScreenshotCaptureRoutes` to only use `isLocalDevelopmentAutomationRequest(req.headers)` and remove the redundant second branch.
- **Issue #2:** Add `isVercelProductionDeployment()` guard inside `isLocalDevelopmentAutomationRequest` as defense-in-depth.

itstimwhite (commented):

PR #10970, opened 2026-06-16
fix(auth): prevent redirect loop between app and onboarding (JOV-2454)

itstimwhite:
## Linear
https://linear.app/jovie/issue/JOV-2454/error-redirect-loop-detected

## Root cause

Two related redirect-loop classes were still occurring after the proxy circuit breaker landed:

1. **Checkout ↔ app ↔ start loop (ENG-002 / TD-09):** `/onboarding/checkout` skipped checkout when Stripe prices were unconfigured by redirecting every user to `/app`. The app shell immediately redirects incomplete profiles back to `/start`, and `useOnboardingClaim` sends claimed users back to `/onboarding/checkout` — producing an infinite server/client redirect cycle.

2. **Proxy rewrite counter false positives:** Proxy used `NextResponse.rewrite()` for onboarding/waitlist gating on marketing paths. The URL bar stayed on the original pathname (e.g. `/pricing`), so repeated RSC/middleware hits on that non-exempt path incremented `jovie_redirect_count` and eventually fired the Sentry `"Redirect loop detected"` circuit breaker even though the user was already being served onboarding content.

3. **Homepage bounce:** Authenticated homepage visits always redirected to `/app` before user-state lookup, forcing needs-onboarding users through an unnecessary `/app → /start` hop that amplified the checkout loop.

## Fix

- Route authenticated homepage visitors based on proxy user state (`/start`, `/waitlist`, or `/app`).
- Switch onboarding/waitlist gating from rewrite → redirect so the canonical URL is used and middleware stops re-counting the same navigation.
- When checkout prices are missing, use `getRedirectForState()` so incomplete profiles return to `/start` instead of `/app`.

## Tests

- Updated proxy behavioral circuit-breaker expectations for redirect-based gating.
- Added regression: authenticated `needsOnboarding` user on `/` goes directly to `/start`.

```
pnpm --filter web exec vitest run tests/unit/middleware/proxy-behavioral.test.ts
✓ 83 tests passed
```

## Files changed

- `apps/web/proxy.ts`
- `apps/web/app/onboarding/checkout/page.tsx`
- `apps/web/tests/unit/middleware/proxy-behavioral.test.ts`

## Release Notes

* **Bug Fixes**
  * Improved onboarding checkout routing when pricing details are missing by using account state to choose the next destination (with a safe fallback).
  * Updated homepage routing for authenticated users to send them to onboarding, waitlist, or the dashboard based on their current status.
  * Enhanced redirect-loop protection by switching to redirect-based handling while preserving circuit-breaker behavior.

* **Tests**
  * Added a regression test for onboarding redirects.
  * Updated middleware tests to validate redirect destinations and redirect-count behavior across consecutive requests.

### DIFF
--- a/apps/web/app/onboarding/checkout/page.tsx
+++ b/apps/web/app/onboarding/checkout/page.tsx
@@ -3,7 +3,11 @@ import { redirect } from 'next/navigation';
 import { getDashboardData } from '@/app/app/(shell)/dashboard/actions';
 import { APP_ROUTES } from '@/constants/routes';
 
-import { CanonicalUserState, resolveUserState } from '@/lib/auth/gate';
+import {
+  CanonicalUserState,
+  getRedirectForState,
+  resolveUserState,
+} from '@/lib/auth/gate';
 import type { PlanIntentTier } from '@/lib/auth/plan-intent';
 import {
   DEFAULT_UPSELL_PLAN,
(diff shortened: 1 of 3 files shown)


itstimwhite:

itstimwhite:

itstimwhite:

PR #10972, opened 2026-06-16
fix(analytics): align audience visit tracking with schema (JOV-3178)

itstimwhite:
## Summary
Fixes `POST /api/audience/visit` returning 500 in dev when `audience_members.latest_referrer_url` (and related summary columns) are missing from the database schema.

## Root cause
Migration `0030_normalize_audience_jsonb.sql` adds `latest_referrer_url` and `latest_action_label`, but dev databases can lag behind. A prior guard (#8705) skipped optional columns via `doesColumnExist`, but a **stale positive column-existence cache** (60s TTL, only cleared on `DATABASE_URL` change) could still allow writes to columns that no longer exist after schema rollback/restore.

## Fix
- Only cache **negative** `doesColumnExist` results to avoid stale positives
- Add `invalidateColumnExistenceCache()` and call it when a 42703 missing-column error is detected
- Retry visit persistence once with optional `audience_members` summary columns disabled
- Add `isMissingColumnError` / `isUndefinedColumnError` helpers for robust PG error detection

## Migration note
`apps/web/drizzle/migrations/0030_normalize_audience_jsonb.sql` already defines the missing columns. Applying migrations to dev will restore full functionality; this change keeps visit tracking working until then.

## Verification
- `pnpm biome check --write <edited files>`
- `pnpm --filter web exec vitest run tests/unit/api/audience/visit.test.ts tests/unit/lib/db/errors.test.ts` (38 passed)

Linear: https://linear.app/jovie/issue/JOV-3178/public-profile-visit-tracking-still-writes-missing-audience-members

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audience-visit persistence reliability by adding schema-compatibility fallback when optional database columns are missing, including targeted cache invalidation and a one-time retry that safely omits unavailable fields.
  * Updated database column-existence caching to prevent stale “missing column” results after schema changes.
* **Tests**
  * Added unit coverage for the retry flow and for missing/undefined-column database error detection (including wrapped errors).

### DIFF
--- a/apps/web/app/api/audience/visit/route.ts
+++ b/apps/web/app/api/audience/visit/route.ts
@@ -15,8 +15,9 @@ import {
   db,
   doesColumnExist,
   doesTableExist,
+  invalidateColumnExistenceCache,
 } from '@/lib/db';
-import { unwrapPgError } from '@/lib/db/errors';
+import { isMissingColumnError, unwrapPgError } from '@/lib/db/errors';
 import {
   audienceMembers,
   audienceReferrers,
@@ -61,6 +62,74 @@ interface AudienceVisitSchemaCompatibility {
   hasAudienceReferrersTable: boolean;
   hasCreatorDistributionEventsTable: boolean;
   hasDailyProfileViewsTable: boolean;
+}
+
+const AUDIENCE_MEMBER_SUMMARY_COLUMNS = [
+  'latest_referrer_url',
+  'latest_action_label',
+] as const;
+
+type AudienceMemberSummaryColumn =
+  (typeof AUDIENCE_MEMBER_SUMMARY_COLUMNS)[number];
+
+function getMissingAudienceMemberSummaryColumn(
+  error: unknown
+): AudienceMemberSummaryColumn | null {
+  for (const columnName of AUDIENCE_MEMBER_SUMMARY_COLUMNS) {
+    if (isMissingColumnError(error, columnName)) {
+      return columnName;
+    }
+  }
+
+  return null;
+}
+
+function downgradeAudienceVisitSchemaCompatibility(
+  compatibility: AudienceVisitSchemaCompatibility,
+  missingColumn: AudienceMemberSummaryColumn
+): AudienceVisitSchemaCompatibility {
+  if (missingColumn === 'latest_referrer_url') {
+    return {
+      ...compatibility,
+      hasAudienceMemberLatestReferrerUrlColumn: false,
+    };
+  }
+
+  return {
+    ...compatibility,
+    hasAudienceMemberLatestActionLabelColumn: false,
+  };
+}
+
+async function persistAudienceVisitWithSchemaFallback(
+  initialCompatibility: AudienceVisitSchemaCompatibility,
+  persist: (compatibility: AudienceVisitSchemaCompatibility) => Promise<void>
+): Promise<void> {
+  let compatibility = initialCompatibility;
+
+  for (let attempt = 0; attempt < 2; attempt++) {
+    try {
+      await persist(compatibility);
+      return;
+    } catch (error) {
+      const missingColumn = getMissingAudienceMemberSummaryColumn(error);
+      if (attempt > 0 || !missingColumn) {
+        throw error;
+      }
+
+      invalidateColumnExistenceCache('audience_members', missingColumn);
+      compatibility = downgradeAudienceVisitSchemaCompatibility(
+        compatibility,
+        missingColumn
+      );
+
+      await captureWarning(
+        '[audience/visit] audience_members summary column missing; retrying without optional column writes',
+        error,
+        { column: missingColumn }
+      );
+    }
+  }
 }
 
 async function resolveAudienceVisitSchemaCompatibility(): Promise<AudienceVisitSchemaCompatibility> {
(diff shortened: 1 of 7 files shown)


itstimwhite:

itstimwhite:

itstimwhite:

itstimwhite:

itstimwhite:

### Logic Errors

**1. `route.ts` — `latest_referrer_url` column guard is inconsistent between update and insert paths**

In the retry path, when `latest_referrer_url` is missing, `downgradeAudienceVisitSchemaCompatibility` sets `hasAudienceMemberLatestReferrerUrlColumn: false`. The update spread correctly gates on this:
```
...(compatibility.hasAudienceMemberLatestReferrerUrlColumn && latestReferrerUrl && { latestReferrerUrl }),
```
But the condition requires **both** the column flag **and** a truthy `latestReferrerUrl`. This means even when the column exists but `latestReferrerUrl` is `null` (e.g., no referrer), the field is simply omitted from the SET clause — which is fine for an update (the existing value is preserved). However, for the **insert** path, the same logic applies: if the column exists but `latestReferrerUrl` is null, the column is omitted, which means the DB default (null) is used. This is correct behavior.

**No bug here** — the logic is actually consistent. The spread only adds the key when both conditions are true, and omission is safe for both insert and update.

**2. `route.ts` — `latest_action_label` is never set in the persist function but its column flag gates `recordAudienceEvent`**

The `downgradeAudienceVisitSchemaCompatibility` function handles `latest_action_label` by setting `hasAudienceMemberLatestActionLabelColumn: false`, and this flag is correctly passed to `recordAudienceEvent` via `includeLatestActionLabel`. However, neither the update nor insert SET/values ever writes `latest_action_label` directly — it's only set via `recordAudienceEvent`. This means the column flag only controls whether `recordAudienceEvent` attempts to write it. This is correct **if** `recordAudienceEvent` is the only writer of that column.

**Potential issue**: If `recordAudienceEvent` is defined in a different file and doesn't handle the `includeLatestActionLabel` flag defensively, the retry could still fail on the second attempt. This is outside the diff scope but worth verifying.

**3. `health.ts:199-206` — Positive cache results are deleted but never re-cached**

After the fix, when `doesColumnExist` returns `true`, the cache entry is deleted (not set). This means every subsequent call to `doesColumnExist` for that column will hit the database again until a negative result is cached. This is intentional (avoid stale positives) but creates a performance trade-off: in a stable schema, every call incurs a DB query. The 60s TTL cache is effectively disabled for columns that exist.

**Not a bug**, but a deliberate design choice documented in the PR description. Acceptable for a rarely-called schema check.

### Security

**None found.** No hardcoded secrets, SQL injection vectors (all queries use Drizzle parameterized queries), XSS risks, or auth bypass issues introduced.

### Data Integrity

**4. `route.ts` — Retry opens a new `withSystemIngestionSession` but the outer transaction semantics change**

The original code wrapped everything in a single `withSystemIngestionSession`. The retry logic calls `persistAudienceVisitWithSchemaFallback`, which calls `withSystemIngestionSession` inside the `persist` callback. On retry, a **new** system ingestion session is created. This means:

- The first attempt's transaction is already rolled back (the error propagated out of `withSystemIngestionSession`).
- The retry starts a fresh transaction with the downgraded compatibility.

This is correct — each attempt is an independent transaction. **No data integrity issue.**

**5. `route.ts` — `utmParams` insert spread is unconditional**

In the insert path:
```
...(utmParams && { utmParams: resolvedUtmParams }),
```
This spreads `utmParams` into the insert values when `utmParams` is truthy. But `resolvedUtmParams` could be `{}` (empty object) if `hasUtmParams` is false and `existing` is undefined (new member). In that case, `utmParams` is falsy so the spread doesn't execute. If `utmParams` is truthy but all values are empty strings, `hasUtmParams` is false, so `resolvedUtmParams` falls back to `existing?.utmParams ?? {}`. For a new member with no existing record, this is `{}`. The insert would set `utmParams: {}`, which is valid JSONB. **No issue.**

### Error Handling

**6. `route.ts:103-104` — `captureWarning` is awaited but its failure is not handled**

```typescript
await captureWarning(
  '[audience/visit] audience_members summary column missing; retrying without optional column writes',
  error,
  { column: missingColumn }
);
```

If `captureWarning` throws, the retry loop will throw instead of retrying. This would cause a 500 response. `captureWarning` likely wraps Sentry or similar and shouldn't throw, but if it does, the retry is lost.

**Why it matters**: The entire purpose of this function is resilience. A logging failure should not break the retry.

**Suggested fix**: Wrap the `captureWarning` call in a try/catch:
```typescript
try {
  await captureWarning(...);
} catch {
  // logging failure should not prevent retry
}
```

**7. `errors.ts` — `isMissingColumnError` matches substring `column ${normalizedColumn}` which could false-positive on unrelated columns**

The check `message.includes(`column ${normalizedColumn}`)` would match `"column latest_referrer_url"` but also theoretically match a longer column name that contains the substring, e.g., a column named `latest_referrer_url_source` would match the check for `latest_referrer_url` if the error message contained `"column latest_referrer_url_source"`. However, since the check is `message.includes(`column ${normalizedColumn}`)` and not `message.includes(`column ${normalizedColumn} `)`, this is actually not a concern for the quoted form (`"latest_referrer_url"`). For the unquoted form, it could be a concern.

**Low risk** in practice because PG error messages typically quote column names or use exact references. But for defensive coding:

**Suggested fix**: Add word-boundary awareness or prefer the quoted form first (which is already checked).

### Cross-File Impact

**8. `invalidateColumnExistenceCache` is exported from both `@/lib/db/client/index.ts` and `@/lib/db/index.ts`**

The new function is re-exported through the chain: `health.ts` → `client/index.ts` → `db/index.ts`. The visit route imports from `@/lib/db`, which is correct. No breakage.

**None found** for cross-file breakage.

---

### Summary

| 6 | **Medium** | `route.ts:103-104` | `captureWarning` failure aborts retry, defeating the resilience purpose |
| 7 | **Low** | `errors.ts` | `isMissingColumnError` substring match could theoretically false-positive on column name prefixes |

**Recommendation**: Fix issue #6 before merge. Issue #7 is acceptable as-is given PG's error message format, but worth a defensive comment.

itstimwhite (commented):

PR #10977, opened 2026-06-16
fix(billing): map legacy founding Stripe price IDs (JOV-1769)

itstimwhite:
## Summary

Fixes `Unknown price ID: price_1T1DegAAI1NrDqJSTtjAwLBi` in Stripe subscription webhooks.

## Root cause

The founding member tier was removed from `PRICE_MAPPINGS` in #6837 (Mar 2026), but at least one active subscription still bills against the retired founding Stripe price `price_1T1DegAAI1NrDqJSTtjAwLBi`. Webhook handlers call `getPlanFromPriceId()`, which only knew active checkout prices, so subscription events were skipped with a critical Sentry error.

## Fix

- Split **active** vs **legacy** Stripe price mappings in `lib/stripe/config.ts`
- Add `lib/stripe/legacy-prices.ts` with known retired price IDs (including JOV-1769 founding price)
- Re-introduce optional `STRIPE_PRICE_FOUNDING_MONTHLY` env var for webhook lookup only
- Legacy prices resolve to **pro** entitlements per `ENTITLEMENT_REGISTRY` (`founding` → `pro`)
- Legacy prices are excluded from checkout and plan-change option lists

## Human review needed

- [ ] Confirm `price_1T1DegAAI1NrDqJSTtjAwLBi` is the founding member monthly price in prod Stripe
- [ ] Set `STRIPE_PRICE_FOUNDING_MONTHLY=price_1T1DegAAI1NrDqJSTtjAwLBi` in Doppler prod/preview if desired (hardcoded fallback also works)
- [ ] Verify no other unknown price IDs are still erroring in Sentry after deploy

linear-issue-id: JOV-1769

## Summary by CodeRabbit

* **New Features**
  * Added support for managing legacy and retired pricing plans alongside active pricing options.
  * Plan change selections now display only current pricing tiers, excluding deprecated plans.

* **Tests**
  * Added test coverage for legacy pricing resolution and exclusion from active checkout options.

### DIFF
--- a/apps/web/lib/env-server-schema.ts
+++ b/apps/web/lib/env-server-schema.ts
@@ -114,6 +114,9 @@ export const ServerEnvSchema = z.object({
   PRINTFUL_STORE_ID: z.string().optional(),
   PRINTFUL_WEBHOOK_SECRET: z.string().optional(),
 
+  // Retired founding tier — webhook lookup only (founding -> pro entitlements)
+  STRIPE_PRICE_FOUNDING_MONTHLY: z.string().startsWith('price_').optional(),
+
   // Stripe price IDs for Pro tier (amounts in lib/config/plan-prices.ts)
   STRIPE_PRICE_PRO_MONTHLY: z.string().startsWith('price_').optional(),
   STRIPE_PRICE_PRO_ANNUAL: z.string().startsWith('price_').optional(),
@@ -385,6 +388,7 @@ export const ENV_KEYS = [
   'PRINTFUL_API_BASE_URL',
   'PRINTFUL_STORE_ID',
   'PRINTFUL_WEBHOOK_SECRET',
+  'STRIPE_PRICE_FOUNDING_MONTHLY',
   'STRIPE_PRICE_PRO_MONTHLY',
   'STRIPE_PRICE_PRO_ANNUAL',
   'STRIPE_PRICE_PRO_YEARLY',
--- a/apps/web/lib/stripe/config.ts
+++ b/apps/web/lib/stripe/config.ts
@@ -6,6 +6,9 @@
  *
  * Pricing tiers and amounts are defined in lib/config/plan-prices.ts (single source of truth).
  * Free tier has no Stripe subscription.
+ *
+ * Active prices are offered at checkout. Legacy prices resolve webhooks for retired
+ * subscriptions but are excluded from checkout and plan-change options.
  */
 
 import 'server-only';
@@ -13,24 +16,36 @@ import 'server-only';
 import { PRICING } from '@/lib/config/pricing';
 import { publicEnv } from '@/lib/env-public';
 import { env } from '@/lib/env-server';
+import { HARDCODED_LEGACY_PRICES } from './legacy-prices';
 
 // Plan types supported by the application
 export type PlanType = 'free' | 'pro' | 'max';
 
 // Price mapping interface
-interface PriceMapping {
+export interface PriceMapping {
   priceId: string;
   plan: PlanType;
   amount: number;
   currency: string;
   interval: 'month' | 'year';
   description: string;
+  /** When true, excluded from checkout and plan-change option lists. */
+  legacy?: boolean;
 }
 
-// Current active price mappings
-const buildPriceMappings = (): Record<string, PriceMapping> => {
+const toMappingRecord = (
+  mappings: PriceMapping[]
+): Record<string, PriceMapping> =>
+  mappings
+    .filter(mapping => Boolean(mapping.priceId))
+    .reduce<Record<string, PriceMapping>>((acc, mapping) => {
+      acc[mapping.priceId] = mapping;
+      return acc;
+    }, {});
+
+// Current active price mappings (checkout-eligible)
+const buildActivePriceMappings = (): Record<string, PriceMapping> => {
   const mappings: PriceMapping[] = [
-    // All tiers from centralized PRICING config
     {
       priceId: PRICING.pro.monthly.priceId || '',
       plan: PRICING.pro.monthly.entitlementPlan,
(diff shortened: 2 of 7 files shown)


itstimwhite:

itstimwhite:

### Logic Errors

**1. `lib/stripe/config.ts:88` — `buildLegacyPriceMappings` dedup check only guards against env-vs-hardcoded collision for the founding price, but silently drops hardcoded duplicates of themselves**

The `mappings.some(mapping => mapping.priceId === legacy.priceId)` check at line ~88 only catches the case where the env var founding price collides with the hardcoded founding price. If `HARDCODED_LEGACY_PRICES` itself contained duplicate `priceId` entries (e.g. a future copy-paste error), the second entry would be silently dropped with no warning. This is low-severity since the array is `readonly` and small, but worth noting: there's no validation or assertion that the hardcoded list is internally unique.

*Suggestion:* Add a dev-time assertion or build-time check that `HARDCODED_LEGACY_PRICES` has unique `priceId` values. This is defensive, not blocking.

---

**2. `lib/stripe/config.ts:97` — Hardcoded `currency: 'usd'` for all legacy prices with no comment explaining why**

Every legacy mapping hardcodes `currency: 'usd'`. If a legacy price were ever in a different currency, this would silently produce wrong data. The founding price and pre-rebrand Standard prices are indeed USD, so this is correct today, but there's no guard or comment.

*Suggestion:* Add a comment on `LegacyPriceDefinition` that `currency` is intentionally fixed to `usd` because all known legacy prices are USD-denominated. Low risk.

---

**3. `lib/stripe/config.ts:103` — `toMappingRecord` filters on `Boolean(mapping.priceId)` but legacy prices never have empty `priceId`**

The `toMappingRecord` helper filters out entries with falsy `priceId`. For active mappings, this makes sense because `PRICING.pro.monthly.priceId` can be empty when env vars are unset. For legacy prices, every entry has a hardcoded non-empty `priceId`, so the filter is a no-op. Not a bug, but it means the filter gives a false impression of safety — if someone added a legacy entry without a `priceId`, it would be silently dropped.

*Suggestion:* None needed for correctness, but be aware.

---

### Security

**None found.**

- `STRIPE_PRICE_FOUNDING_MONTHLY` is validated with `.startsWith('price_')` in the env schema — good.
- No hardcoded secrets; the hardcoded price IDs in `legacy-prices.ts` are not secrets (they're Stripe price identifiers, not API keys).
- No SQL injection, XSS, or auth bypass vectors introduced.
- The `legacy` flag correctly prevents legacy prices from appearing in checkout flows.

---

### Data Integrity

**4. `lib/stripe/config.ts:82` — Legacy founding price `amount: 1200` is hardcoded and could drift from the actual Stripe price amount**

The founding legacy price has `amount: 1200` hardcoded. If the actual Stripe price amount differs (e.g., it was $9.99/mo = 999 cents, or $12.99/mo = 1299 cents), the `amount` field in the mapping would be wrong. This `amount` is used for sorting in `getAvailablePricing()` (but legacy prices are excluded there) and potentially for display in webhook handlers. If any code path uses `amount` from a legacy mapping for billing calculations, this would be incorrect.

*Why it matters:* The `amount` field on a `PriceMapping` is used in `getAvailablePricing()` for sorting. While legacy prices are excluded from that function, if any future code reads `.amount` from a legacy mapping (e.g., for display in an admin panel or billing history), it would show the wrong value.

*Suggestion:* Add a comment documenting what `1200` represents (e.g., `$12.00/mo`) and verify it matches the actual Stripe dashboard amount. Consider marking `amount` as informational-only for legacy entries.

---

**5. `lib/stripe/legacy-prices.ts:44` — Pre-rebrand Standard yearly price `amount: 19200` — verify this matches Stripe**

Same concern as above. `$192.00/year` for a "Standard" tier seems plausible but should be verified against the Stripe dashboard. If wrong, any code that reads `.amount` from this mapping gets incorrect data.

*Suggestion:* Same as #4 — verify and document.

---

### Error Handling

**None found.**

- `getPlanFromPriceId` returns `null` for unknown prices (unchanged behavior).
- `getPriceMappingDetails` returns `null` for unknown prices (unchanged behavior).
- The legacy price lookup is additive — it can't cause previously-working lookups to fail.
- No new async operations or external calls were introduced.

---

### Cross-File Impact

**6. `lib/stripe/plan-change.ts:602,626,658` — `PRICE_MAPPINGS` replaced with `ACTIVE_PRICE_MAPPINGS` in `getAvailablePlanChanges`**

This is the intended behavior change (excluding legacy prices from plan-change options), and it's correct. The diff shows three sites updated consistently. ✅

**7. `tests/unit/lib/stripe/plan-change.critical.test.ts:69` — Mock updated from `PRICE_MAPPINGS` to `ACTIVE_PRICE_MAPPINGS`**

The mock getter was updated to match the new export name. ✅

**8. `tests/unit/lib/stripe/plan-change.test.ts:57` — Mock updated from `PRICE_MAPPINGS` to `ACTIVE_PRICE_MAPPINGS`**

Same as above. ✅

**9. No other callers of `PRICE_MAPPINGS` were missed**

The diff shows `PRICE_MAPPINGS` is still exported and still used by `getPlanFromPriceId`, `getPriceMappingDetails`, and `getActivePriceIds` (indirectly via `ACTIVE_PRICE_MAPPINGS`). The only consumers that needed to switch to `ACTIVE_PRICE_MAPPINGS` were the plan-change UI functions, which is correct. ✅

---

### Summary

| 1 | Low | `config.ts:~88` | Dedup check doesn't guard against internal duplicates in `HARDCODED_LEGACY_PRICES` |
| 2 | Low | `config.ts:~97` | Hardcoded `currency: 'usd'` undocumented for legacy prices |
| 3 | Info | `config.ts:~103` | `Boolean(filter)` is no-op for legacy prices |
| 4 | Medium | `config.ts:~82` | Hardcoded `amount: 1200` for founding price — verify against Stripe |
| 5 | Medium | `legacy-prices.ts:44` | Hardcoded `amount: 19200` for Standard yearly — verify against Stripe |
| 6-9 | ✅ | plan-change.ts + tests | Cross-file updates are consistent and correct |

**No blocking issues.** The two medium items (#4, #5) are about data accuracy of hardcoded amounts — they won't cause runtime failures but could produce incorrect display values if `.amount` is ever read from legacy mappings. Recommend verifying both amounts against the Stripe dashboard before merge.

itstimwhite (commented):

itstimwhite:

PR #10978, opened 2026-06-16
fix(analytics): fail-safe tracking token when secret unset (JOV-2327)

itstimwhite:
## Summary

Fixes `TypeError: Tracking token signing not configured` when `TRACKING_TOKEN_SECRET` is unset in non-development environments.

- `generateTrackingToken` and `getClientTrackingToken` now return `null` instead of throwing
- Development still uses `:dev` placeholder tokens for local testing
- Profile pages read the nullable token without try/catch
- `/api/audience/visit-token` returns `{ token: null, expiresAt: null }` gracefully when signing is disabled
- Added unit tests for production and development paths

## Testing

- [x] `pnpm typecheck:web:fast`
- [x] `pnpm biome check` on changed files
- [x] `vitest run tests/lib/analytics/tracking-token.test.ts`
- [x] `vitest run tests/unit/api/audience/visit-token.test.ts`

## Summary by CodeRabbit

* **UI Updates**
  * Updated error message text and button labels on the public profile page for clearer user guidance

* **Bug Fixes**
  * Improved tracking system resilience to gracefully handle unavailable signing credentials instead of failing

### DIFF
--- a/apps/web/app/[username]/page.tsx
+++ b/apps/web/app/[username]/page.tsx
@@ -380,10 +380,10 @@ export default async function ArtistPage({ params }: Readonly<Props>) {
       <div className='px-4 py-8'>
         <ErrorBanner
           title='Profile is temporarily unavailable'
-          description='We could not load this Jovie profile right now. Please refresh or try again in a few minutes.'
+          description='We could not load this profile right now, so please refresh or try again in a few minutes.'
           actions={[
-            { label: 'Try again', href: `/${username.toLowerCase()}` },
-            { label: 'Go home', href: '/' },
+            { label: 'Try Again', href: `/${username.toLowerCase()}` },
+            { label: 'Go Home', href: '/' },
           ]}
           testId='public-profile-error'
         />
@@ -421,12 +421,8 @@ export default async function ArtistPage({ params }: Readonly<Props>) {
   // Generate a short-lived HMAC token so the client can authenticate its visit
   // tracking request to /api/audience/visit (requires TRACKING_TOKEN_SECRET).
   // Falls back to undefined gracefully if the secret is not configured.
-  let visitTrackingToken: string | undefined;
-  try {
-    visitTrackingToken = getClientTrackingToken(profile.id).token;
-  } catch {
-    // Secret not configured — visit tracking will proceed without token auth
-  }
+  const visitTrackingToken =
+    getClientTrackingToken(profile.id).token ?? undefined;
 
   const latestRelease = fetchedLatestRelease;
 
--- a/apps/web/app/api/audience/visit-token/route.ts
+++ b/apps/web/app/api/audience/visit-token/route.ts
@@ -81,7 +81,8 @@ export async function GET(request: NextRequest) {
   }
 
   try {
-    return NextResponse.json(getClientTrackingToken(parsed.data.profileId), {
+    const tokenResult = getClientTrackingToken(parsed.data.profileId);
+    return NextResponse.json(tokenResult, {
       headers: NO_STORE_HEADERS,
     });
   } catch (error) {
--- a/apps/web/lib/analytics/tracking-token.ts
+++ b/apps/web/lib/analytics/tracking-token.ts
@@ -75,6 +75,10 @@ function verifySignature(
   timestamp: number
 ): boolean {
   const expectedSignature = generateSignature(profileId, timestamp);
+  if (!expectedSignature) {
+    return false;
+  }
+
   const signatureBuffer = Buffer.from(signature, 'hex');
   const expectedBuffer = Buffer.from(expectedSignature, 'hex');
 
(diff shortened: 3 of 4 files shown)


itstimwhite:

itstimwhite:

### Logic Errors

**1. `tracking-token.ts:143` — Sentry breadcrumb fires on every call in production when secret is unset**

The `Sentry.addBreadcrumb()` call at line 143 (inside `generateTrackingToken`) fires on **every** invocation in production when `TRACKING_TOKEN_SECRET` is unset. This is not a one-time warning — it's per-request. In a high-traffic scenario, this floods Sentry with breadcrumbs, potentially causing rate-limiting or cost issues.

```ts
Sentry.addBreadcrumb({
  category: 'tracking-token',
  message: 'TRACKING_TOKEN_SECRET not set - skipping token generation',
  level: 'warning',
  data: { profileId },
});
```

**Why it matters:** Sentry breadcrumbs are limited per transaction (typically 100), so this can crowd out more useful breadcrumbs. The `profileId` in `data` also means every breadcrumb is unique, defeating deduplication.

**Suggested fix:** Either use `Sentry.captureMessage()` with a throttle/cache mechanism, or emit this only once (e.g., a module-level `let warned = false` guard). Alternatively, use `Sentry.addBreadcrumb` without `profileId` in data so Sentry can group them.

---

**2. `tracking-token.ts:143` — `profileId` in Sentry breadcrumb data may be PII-adjacent**

The `profileId` is passed as breadcrumb data. Depending on what `profileId` contains (could be a UUID tied to a user), this could leak into Sentry events.

**Why it matters:** If `profileId` is user-identifiable, it ends up in Sentry breadcrumbs attached to other errors in the same request scope.

**Suggested fix:** Remove `profileId` from the breadcrumb data, or hash it.

---

**3. `tracking-token.ts:110` — `generateSignature` null check is unreachable dead code**

At line 110, after calling `generateSignature`, there's a null check:

```ts
const signature = generateSignature(profileId, timestamp);
if (!signature) {
  return null;
}
```

However, this code path is only reached when `isTrackingTokenEnabled()` returns `true` (line 137), which means `getTrackingSecret()` is non-null. So `generateSignature` will always have a secret and never return `null` in this branch.

**Why it matters:** Not a bug per se, but it's misleading — a future reader might think this path is reachable. It also means if `isTrackingTokenEnabled()` ever has a race condition or changes semantics, this silently returns `null` instead of surfacing the issue.

**Suggested fix:** Either remove the dead null check with a comment explaining why, or add an assertion (`if (!signature) throw new Error('unreachable')`) to catch future regressions loudly.

---

**4. `visit-token/route.ts:84` — `catch` block now swallows all errors, not just signing errors**

Previously, the `try/catch` in the route handler caught `generateTrackingToken` throwing. Now that it returns `null` instead of throwing, the `catch` block at line 84 still exists but will now catch unrelated errors (e.g., Zod validation failure from `parsed.data.profileId` being malformed, or a JSON serialization issue). The error handling comment says "Secret not configured" but the catch is now broader than that.

Looking at the diff, the `catch` block wasn't modified — it was already there. But the semantic change means errors that previously would have been thrown by `generateTrackingToken` are now silently handled by returning `{ token: null, expiresAt: null }`, while the `catch` block may still fire for other reasons and return a generic error response.

**Why it matters:** If `getClientTrackingToken` or `NextResponse.json()` throws for a non-signing reason, the catch block returns an error to the client that looks like a signing failure.

**Suggested fix:** Consider narrowing the catch or adding a comment that it now handles unexpected errors beyond signing configuration.

---

### Security

**1. `tracking-token.ts:120` — Dev token format is predictable and not authenticated**

In development, the token is `${profileId}:${Date.now()}:dev`. This is trivially forgeable — any client can construct a valid-looking token for any profile.

**Why it matters:** This is gated by `NODE_ENV === 'development'`, so it's low risk. However, if `NODE_ENV` is accidentally set to `development` in a deployed environment (e.g., a staging server), visit tracking can be spoofed trivially.

**Suggested fix:** Consider adding a check that the request originates from localhost or an internal network when using dev tokens, or log a prominent warning when dev tokens are generated outside of local development.

---

**2. `tracking-token.ts:143` — `profileId` in Sentry data (minor)**

Covered above under Logic Errors. Repeating here for completeness: `profileId` in breadcrumb data could be user-identifiable.

---

### Data Integrity

**None found.**

The null propagation chain is correct: `generateTrackingToken` → `getClientTrackingToken` → profile page → API route. Each layer handles `null` appropriately.

---

### Error Handling

**1. `tracking-token.ts:143` — Sentry breadcrumb on every call (also listed above)**

This is the primary error handling concern. The breadcrumb is emitted per-call rather than once.

---

**2. `visit-token/route.ts:84` — Catch block semantics changed (also listed above)**

The catch block's purpose has shifted silently.

---

### Cross-File Impact

**1. `apps/web/app/[username]/page.tsx:424` — `visitTrackingToken` is now `string | undefined`**

The change from `let visitTrackingToken: string | undefined` with try/catch to `const visitTrackingToken = getClientTrackingToken(profile.id).token ?? undefined` is correct. The `?? undefined` converts `null` to `undefined`, preserving the existing type contract for downstream consumers. **No issue here.**

**2. Callers of `generateTrackingToken` — return type changed from `string` to `string | null`**

Any existing caller that doesn't handle `null` will now have a potential runtime error. The diff only shows the profile page and API route as consumers, both of which are updated. However, if there are other callers of `generateTrackingToken` or `getClientTrackingToken` outside these files, they would need updating.

**Why it matters:** The function signature change is a breaking contract change. A grep across the codebase would be needed to confirm no other callers exist.

**Suggested fix:** Run `grep -r "generateTrackingToken\|getClientTrackingToken" apps/web --include="*.ts" --include="*.tsx"` to confirm all callers are updated.

---

### Summary

|---|---|---|
| Logic errors | 3 | Medium (Sentry flooding, dead code, PII) |
| Security | 1 | Low (dev token forgery if NODE_ENV misconfigured) |
| Data integrity | 0 | — |
| Error handling | 2 | Medium (Sentry flooding, catch scope) |
| Cross-file impact | 1 | Medium (need to verify no other callers) |

itstimwhite (commented):

PR #10983, opened 2026-06-16
fix(infra): clamp gbrain Supabase pool budget in Jovie wrappers (JOV-3123)

### LINKED ISSUE #10815: gbrain: Supabase connection exhaustion — clamp pool budget, stop per-process pool multiplication
**Problem** — Tim is seeing repeated Supabase 'connections exhausted' errors from the gbrain instance (project quaapamommjwgwodvgml), despite being the only user.

**Severity & Priority** — High (P1): the company brain intermittently hard-fails for every agent loop that reads/writes it; transient retry is the only workaround. Not P0 (no data loss).

**MRR opportunity score** — Indirect/infra: gbrain outages degrade every Eve/Hermes loop that compounds product velocity. Effort is small (config), so leverage is high. (reach=all agent loops, freq=daily per Tim, annoyance=high, WTP/ATP=n/a internal, effort=S)

**Evidence**

* Tim in Slack #product 2026-06-13: "lots of errors from Supabase on gbrain… connections being exhausted… make sure we're not double pooling."
* Code recon (garrytan/gbrain@master):
  * `src/core/connection-manager.ts` — dual-pool topology: read pool **size 10** (pooler :6543) + direct pool **size 3** (:5432), and pools are **INSTANCE-owned: each PostgresEngine constructs its own ConnectionManager** (worker engines inherit only when `parent` is passed).
  * `src/core/db.ts` — `DEFAULT_POOL_SIZE_FALLBACK = 10`; comments document an observed failure mode: autopilot SIGKILL'd by launchd left a pooler backend `active`/`ClientRead` for **24h+ holding a RowExclusiveLock**.
  * `src/commands/sync.ts` / `import.ts` — parallel sync/import spawn N additional worker engines, each with its own pool (~2 conns each).
  * `src/commands/doctor.ts` `computePoolBudgetCheck` — a `GBRAIN_MAX_CONNECTIONS` budget clamp EXISTS but is **opt-in and disabled by default** ("GBRAIN_MAX_CONNECTIONS not set — connection budget clamp disabled"). Doctor itself recommends `GBRAIN_POOL_SIZE=2` for low-cap poolers like Supabase Supavisor.
* Could NOT pull live Supabase postgres logs this sweep: the Supabase MCP OAuth session is expired again (HTTP 422 client_secret) — Tim must re-link Supabase in Settings > Integrations.
* Related: OWL reported the instance engine is (mis)configured to `pglite` — config drift on the same install; `gbrain doctor` will surface both.

**Root-cause hypothesis (high confidence)** — Not a single 'double pool' bug but **per-process pool multiplication with no global budget**: every concurrent gbrain process (launchd autopilot tick, serve/serve-http MCP server, ad-hoc CLI like sync/import/embed) opens its own 10+3 connections, and parallel sync/import adds 2 per worker. On a small Supabase tier (direct max_connections ≈ 60 incl. reserved slots) 3–4 overlapping processes + a few orphaned ClientRead backends from SIGKILL'd autopilot runs exhausts the cap. The budget clamp that would prevent this is off by default.

**Proposed fix** (config-first; on the always-on MacBook Pro env)

1. Set in gbrain env: `GBRAIN_POOL_SIZE=2`, `GBRAIN_DIRECT_POOL_SIZE=1`, `GBRAIN_MAX_CONNECTIONS=<Supabase tier max_connections>` — turns on the built-in budget clamp and matches doctor's own Supavisor guidance.
2. Upgrade the instance to latest upstream (v0.42.x has the clamp + orphan-session GUC hardening) — already runbooked in JovieInc/Jovie#10458; do together.
3. Ensure exactly ONE long-lived gbrain process owns pools (serve), and launchd autopilot ticks don't overlap/stack.
4. Verify: `gbrain doctor` pool_budget check passes; in Supabase SQL editor run `select state, wait_event, count(*) from pg_stat_activity where datname='postgres' group by 1,2;` — no long-lived `ClientRead` orphans, total conns well under cap.
5. Optional upstream PR: default-on clamp when a Supabase pooler URL is detected.

**Acceptance criteria** — Zero 'too many connections' / pool-exhaustion errors over 7 days of normal autopilot+sync operation; doctor pool_budget = ok; pg_stat_activity steady-state < 50% of tier cap.

**Issue shape & model tag** — Small, well-spec'd (env/config + verification; no app code required). Routing: **human/ops (Tim or Hermes on the MBP)** — env + launchd config is dashboard/machine work a coding agent can't perform (per prior JovieInc/Jovie#10034 learning). The optional upstream clamp PR could go to Kimi/DeepSeek later.

**Related** — JovieInc/Jovie#10458 (instance upgrade runbook — do fix JovieInc/Jovie#2 together), JovieInc/Ops#113 (older PGLite single-writer lock issue, same install), JovieInc/Ops#130 (gbrain repo credential rotation, unrelated cause but same system).

itstimwhite:
## Summary

Stops Supabase connection exhaustion from gbrain per-process pool multiplication by
clamping pool budget env vars in Jovie-owned wrappers and documenting operator setup.

- Add `scripts/lib/gbrain-pool-env.{mjs,sh}` — detects Supabase vs PGLite and exports
  conservative defaults (`GBRAIN_POOL_SIZE=2`, `GBRAIN_DIRECT_POOL_SIZE=1`,
  `GBRAIN_DISABLE_DIRECT_POOL=1`, `GBRAIN_MAX_CONNECTIONS=15`).
- Wire helper into `scripts/codex-gbrain-sync.sh` before `gbrain doctor` / sync.
- Document operator env + verification in `docs/GBRAIN_POOL_BUDGET.md`; cross-link from
  `docs/HERMES_AIR.md`, `setup-gbrain` skill, and launchd README.
- Unit tests for detection/defaults (`scripts/lib/__tests__/gbrain-pool-env.test.mjs`).

## Problem

GitHub #10815 / JOV-3123: overlapping gbrain processes each open default 10+3 pools on
Supabase, exhausting Supavisor session pool (`EMAXCONNSESSION`). gbrain v0.42+ has an
opt-in `GBRAIN_MAX_CONNECTIONS` clamp but it is off unless env is set in every wrapper.

## Operator follow-up (MacBook Pro, not in this PR)

Long-lived Pro cron/launchd wrappers still need the same exports in shell profile or
`~/.gbrain/env`. See `docs/GBRAIN_POOL_BUDGET.md`.

## Verification

```bash
pnpm exec vitest run --project workspace-scripts scripts/lib/__tests__/gbrain-pool-env.test.mjs
bash -n scripts/codex-gbrain-sync.sh scripts/lib/gbrain-pool-env.sh
node scripts/lib/gbrain-pool-env.mjs --json ~/.gbrain/config.json  # PGLite → engine-skip
```

## Linear

Closes #10815

## Summary by CodeRabbit

* **New Features**
  * Added pool budget configuration helpers and utilities to prevent Supabase connection exhaustion in multi-process environments.

* **Documentation**
  * Added comprehensive Supabase pool budget setup, verification, and troubleshooting guide.
  * Updated setup documentation with pool configuration requirements and environment variable guidance.
  * Updated runbook with connection exhaustion recovery procedures and configuration verification steps for different deployment types.

### DIFF
--- a/.claude/skills/setup-gbrain/SKILL.md
+++ b/.claude/skills/setup-gbrain/SKILL.md
@@ -960,6 +960,25 @@ Walk the user through the supabase.com steps:
 
 Then follow the same secret-read + verify + init flow as Path 1.
 
+### Supabase pool budget (Paths 1, 2a, 2b)
+
+After any Supabase `gbrain init`, clamp per-process pool multiplication so
+overlapping CLI processes (serve, autopilot, sync) do not exhaust the
+Supavisor session pool. Jovie-owned wrappers source
+`scripts/lib/gbrain-pool-env.sh`; for interactive shells add the same exports
+to `~/.zshrc` or your launchd/cron env:
+
+```bash
+export GBRAIN_DISABLE_DIRECT_POOL=1
+export GBRAIN_POOL_SIZE=2
+export GBRAIN_DIRECT_POOL_SIZE=1
+export GBRAIN_MAX_CONNECTIONS=15   # tune to your Supabase tier / pool_size
+```
+
+`GBRAIN_MAX_CONNECTIONS` turns on gbrain's opt-in budget clamp (v0.42+).
+Verify with `gbrain doctor --json | jq '.checks[] | select(.name=="pool_budget")'`.
+See `docs/GBRAIN_POOL_BUDGET.md` for tier guidance and troubleshooting.
+
 ### Path 3 (PGLite local)
 
 ```bash
--- a/docs/GBRAIN_POOL_BUDGET.md
+++ b/docs/GBRAIN_POOL_BUDGET.md
@@ -1,0 +1,86 @@
+# gbrain Supabase pool budget
+
+Jovie agents and Hermes cron wrappers call the external `gbrain` CLI. On
+Supabase-backed brains, each process opens its own Postgres pools (defaults:
+read pool 10 + direct pool 3). Overlapping processes — `gbrain serve`, autopilot
+ticks, `gbrain sync`, Codex hooks — multiply connections and can trigger
+`EMAXCONNSESSION` / "connections exhausted" on low-cap Supavisor tiers.
+
+PGLite engines (Hermes-Air default) do not use Postgres pooling; this doc does
+not apply there.
+
+## Env vars
+
+| Variable | Default (when clamp applies) | Purpose |
+|---|---:|---|
+| `GBRAIN_DISABLE_DIRECT_POOL` | `1` | Single-pool mode; avoids doubling connections per process |
+| `GBRAIN_POOL_SIZE` | `2` | Read/pooler pool size per process (gbrain default is 10) |
+| `GBRAIN_DIRECT_POOL_SIZE` | `1` | Direct `:5432` pool size when direct pool is enabled |
+| `GBRAIN_MAX_CONNECTIONS` | `15` | Opt-in per-process budget clamp for parallel sync workers (gbrain v0.42+) |
+| `GBRAIN_POOL_BUDGET_DISABLED` | unset | Set to `1` to skip Jovie's auto-clamp helper |
+
+Set `GBRAIN_MAX_CONNECTIONS` to your Supabase **session pooler** `pool_size`
+(not raw Postgres `max_connections`). Supavisor free tiers are often 15–20.
+
+## Where Jovie applies the clamp
+
+- `scripts/codex-gbrain-sync.sh` — sources `scripts/lib/gbrain-pool-env.sh` before
+  `gbrain doctor` / sync during Codex session hooks.
+- `scripts/lib/gbrain-pool-env.mjs` — shared detection (Supabase vs PGLite) and
+  defaults; unit-tested.
+
+Operator-owned surfaces (not rendered by this repo) still need the same exports:
+
+- `~/.zshrc` / shell profile for ad-hoc `gbrain` CLI
+- launchd plists / cron wrappers on the always-on MacBook Pro
+- `gbrain serve` / autopilot long-lived processes
+
+## Recommended operator setup (MacBook Pro, Supabase brain)
+
+Add to shell profile or a dedicated `~/.gbrain/env` sourced by cron:
+
+```bash
+export GBRAIN_DISABLE_DIRECT_POOL=1
+export GBRAIN_POOL_SIZE=2
+export GBRAIN_DIRECT_POOL_SIZE=1
+export GBRAIN_MAX_CONNECTIONS=15
+```
+
+Upgrade gbrain to v0.42+ (see JovieInc/Jovie#10458) so `pool_budget` doctor check
+and the `GBRAIN_MAX_CONNECTIONS` clamp are available.
+
+## Verification
+
+```bash
+gbrain doctor --json | jq '.checks[] | select(.name=="pool_budget")'
+```
+
+Expect `status: "ok"` with a message showing room for parallel sync workers.
+In Supabase SQL editor:
+
+```sql
+select state, wait_event_type, count(*)
+from pg_stat_activity
+where datname = current_database()
+group by 1, 2
+order by 3 desc;
+```
+
+Steady-state connections should stay well under tier cap; no long-lived
+`ClientRead` orphans after SIGKILL'd autopilot runs.
+
+## Troubleshooting
+
+| Symptom | Likely cause | Fix |
+|---|---|---|
+| `EMAXCONNSESSION` / pool_size exceeded | Too many overlapping gbrain processes × default pool sizes | Lower `GBRAIN_POOL_SIZE`; ensure clamp env is set in **every** wrapper |
+| `pool_budget` warns "no room for parallel sync worker" | `GBRAIN_MAX_CONNECTIONS` too low for parent pool | Lower `GBRAIN_POOL_SIZE` or raise `GBRAIN_MAX_CONNECTIONS` |
+| Errors only on Codex sessions | Shell wrappers clamped but serve/autopilot are not | Add env to launchd/cron for long-lived processes |
+| PGLite brain on Air | N/A — pool env is intentionally skipped | No action |
+
+## Related
+
+- GitHub issue: [JovieInc/Jovie#10815](https://github.com/JovieInc/Jovie/issues/10815)
+- Linear: [JOV-3123](https://linear.app/jovie/issue/JOV-3123)
+- gbrain upstream env docs: `GBRAIN_POOL_SIZE` in garrytan/gbrain `src/core/db.ts`
+- Hermes-Air (PGLite): [`docs/HERMES_AIR.md`](./HERMES_AIR.md)
--- a/docs/HERMES_AIR.md
+++ b/docs/HERMES_AIR.md
@@ -257,6 +257,15 @@ Once you understand the cause, fix it (likely in `free-model-router.ts` rankings
 2. On Air: `lsof -i :<gbrain-port>` — confirm `gbrain serve --http --bind <tailscale-ip>` is listening.
 3. On Pro: update MCP config to point at `http://<air-tailscale-ip>:<port>`.
 
+### Supabase connection exhaustion (Pro-hosted company brain)
+
+Hermes-Air runs gbrain on **PGLite** — no Postgres pool budget needed on the Air.
+The shared Supabase company brain on the MacBook Pro is different: each gbrain
+process multiplies pools unless clamped. See [`docs/GBRAIN_POOL_BUDGET.md`](./GBRAIN_POOL_BUDGET.md)
+for `GBRAIN_POOL_SIZE`, `GBRAIN_MAX_CONNECTIONS`, and verification steps.
+Codex hooks already source `scripts/lib/gbrain-pool-env.sh`; long-lived Pro
+cron/launchd wrappers must export the same env.
+
 ### Air rebooted
 
 launchd handles this. All services come back automatically because `RunAtLoad: true` is set on every plist. Confirm with status checks above.
--- a/scripts/codex-gbrain-sync.sh
+++ b/scripts/codex-gbrain-sync.sh
@@ -10,6 +10,11 @@ set -u
 PHASE="${1:-manual}"
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+# Supabase-backed brains: clamp per-process pool multiplication before any gbrain CLI.
+# shellcheck source=lib/gbrain-pool-env.sh
+. "${SCRIPT_DIR}/lib/gbrain-pool-env.sh"
+apply_gbrain_pool_budget
 GBRAIN_DOCTOR_TIMEOUT_SECONDS="${CODEX_GBRAIN_DOCTOR_TIMEOUT_SECONDS:-10}"
 GBRAIN_SYNC_TIMEOUT_SECONDS="${CODEX_GBRAIN_SYNC_TIMEOUT_SECONDS:-60}"
 
--- a/scripts/hermes/launchd/README.md
+++ b/scripts/hermes/launchd/README.md
@@ -6,6 +6,8 @@ Each `.plist.template` uses `{{HOME}}`, `{{JOVIE_REPO}}`, `{{HERMES_BIN}}`, `{{G
 
 The Hermes gateway itself is managed by the installed Hermes CLI as `ai.hermes.gateway`; these templates manage the Air-specific watchdog, gbrain server, and cron jobs around it.
 
+**Supabase pool budget:** Hermes-Air's `gbrain serve` uses PGLite and does not need Postgres pool env vars. If you run a Supabase-backed gbrain on the MacBook Pro (serve/autopilot/cron), export the clamp documented in `docs/GBRAIN_POOL_BUDGET.md` in every long-lived wrapper — Codex hooks already source `scripts/lib/gbrain-pool-env.sh`.
+
 ## Units
 
 | File | Schedule | Purpose |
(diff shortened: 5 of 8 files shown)


itstimwhite:
## Gate evidence (JOV-3123)

Focused infra change — scripts + docs only. Verification run locally on branch `tim/jov-3123-gbrain-supabase-connection-exhaustion-clamp-pool-budget-stop` @ `57a2ed8389b3e0bb952e9819ddce4fd5766a202a`.

| Gate | Status | Evidence |
|---|---|---|
| gstack.qa.exhaustive | passed | `pnpm exec vitest run --project workspace-scripts scripts/lib/__tests__/gbrain-pool-env.test.mjs` — 10/10 tests |
| gstack.review | passed | Self-review: no secrets in tests; PGLite no-op verified via `node scripts/lib/gbrain-pool-env.mjs --json` |
| gstack.ship | passed | `bash -n` on shell wrappers; biome + pre-push clean |

itstimwhite:

itstimwhite:

### 1. Logic Errors

**`scripts/lib/gbrain-pool-env.sh:30` — `eval` of node output without validation**

The shell wrapper calls `node ... --shell` and `eval`s whatever it prints. If the node script emits anything unexpected (warnings on stderr that bleed through, or a partial write), `eval` will attempt to execute it. The `2>/dev/null` redirect swallows errors silently, so a broken config or missing node binary produces no diagnostic — just a silent no-op.

**Why it matters:** Silent failure means operators think the clamp is active when it isn't. If `node` isn't on `$PATH` (e.g., in a minimal cron environment), the script does nothing and logs nothing.

**Suggested fix:** Check the exit code of the node invocation and emit a warning to stderr on failure:

```sh
exports="$(node "${lib_dir}/gbrain-pool-env.mjs" --shell "$config_path" 2>/dev/null)"
rc=$?
if [[ $rc -ne 0 ]]; then
  echo "gbrain-pool-env: node helper exited with code ${rc}" >&2
  return 1
fi
```

---

**`scripts/lib/gbrain-pool-env.mjs:149` — `HOME` may be `undefined` in non-interactive contexts**

```js
configPath: `${process.env.HOME ?? ''}/.gbrain/config.json`,
```

When `HOME` is unset (launchd, cron), this resolves to `undefined/.gbrain/config.json`. The `planPoolBudget` function will then hit the `config-unreadable` catch-all and silently skip. This is the exact environment where the clamp is most needed.

**Why it matters:** launchd and cron jobs on macOS often don't inherit `HOME`. The script silently no-ops in the environments that need it most.

**Suggested fix:** Fall back to `os.homedir()` or fail explicitly:

```js
import { homedir } from 'node:os';
// ...
configPath: `${process.env.HOME || homedir()}/.gbrain/config.json`,
```

---

**`scripts/lib/gbrain-pool-env.mjs:100` — `config-unreadable` silently swallows all exceptions**

```js
try {
  config = JSON.parse(readFileSync(configPath, 'utf8'));
} catch {
  return { apply: false, env: {}, reason: 'config-unreadable' };
}
```

This catches *everything* — `ENOENT`, permission errors, and actual JSON parse failures. An operator with a malformed `config.json` gets no diagnostic.

**Why it matters:** Misconfigured gbrain configs will silently skip the clamp with no signal.

**Suggested fix:** At minimum, log to stderr when not in `--json` mode:

```js
} catch (err) {
  if (args.mode !== 'json') {
    process.stderr.write(`gbrain-pool-env: cannot read ${configPath}: ${err.message}\n`);
  }
  return { apply: false, env: {}, reason: 'config-unreadable' };
}
```

---

### 2. Security

**`scripts/lib/gbrain-pool-env.sh:31` — `eval` of dynamically generated string**

While the node script uses `shellQuote()`, the shell script `eval`s the entire output as a single string. If `shellQuote` has an edge case (e.g., a value containing a newline followed by arbitrary commands), the `eval` would execute them. The current `shellQuote` regex allows newlines through the safe path (`A-Za-z0-9_./:@%+-` doesn't match `\n`, so it takes the single-quote path — but a value with embedded newlines would produce a multi-line export that `eval` interprets as two statements).

**Why it matters:** If any upstream source (env var, config) injects a newline, arbitrary command execution is possible.

**Suggested fix:** Either validate that `exports` contains only `export KEY=VALUE` lines before eval, or use `declare -x` with process substitution:

```sh
while IFS= read -r line; do
  if [[ "$line" =~ ^export\ [A-Za-z_][A-Za-z0-9_]= ]]; then
    eval "$line"
  fi
done <<< "$exports"
```

---

**`scripts/lib/gbrain-pool-env.mjs:130` — `shellQuote` allows newlines in "safe" values**

The regex `/^[A-Za-z0-9_./:@%+-]+$/` doesn't include `\n`, so a value with newlines takes the single-quote path. But a value like `foo\nbar`bash -c "rm -rf /"`foo` would be wrapped in single quotes, which *does* protect against injection. However, a value containing `'` followed by a newline and commands would break out: the replacement `'\\''` handles embedded single quotes correctly for POSIX, but the resulting multi-line string passed to `eval` in the shell script is still risky.

**Why it matters:** Combined with the `eval` in the shell wrapper, this is the weakest link.

**Suggested fix:** Reject or strip newlines in `shellQuote`:

```js
export function shellQuote(value) {
  if (/^[A-Za-z0-9_./:@%+-]+$/.test(value)) {
    return value;
  }
  return `'${value.replace(/\n/g, '').replace(/'/g, `'\\''`)}'`;
}
```

---

### 3. Data Integrity

**None found.** This PR doesn't touch the database layer, migrations, or any data queries. It only sets environment variables for an external CLI.

---

### 4. Error Handling

**`scripts/lib/gbrain-pool-env.sh:28` — Missing `gbrain-pool-env.mjs` is silently ignored**

```sh
if [[ ! -f "${lib_dir}/gbrain-pool-env.mjs" ]]; then
  return 0
fi
```

Returns success (0) when the helper script is missing. In a deployment where the `.sh` file was copied without the `.mjs` file, operators get no signal.

**Why it matters:** Silent success on missing dependency.

**Suggested fix:** Return a non-zero exit code or emit a warning:

```sh
if [[ ! -f "${lib_dir}/gbrain-pool-env.mjs" ]]; then
  echo "gbrain-pool-env: helper ${lib_dir}/gbrain-pool-env.mjs not found" >&2
  return 1
fi
```

---

**`scripts/codex-gbrain-sync.sh:15` — `apply_gbrain_pool_budget` failure is unchecked**

The function is called without checking its return value. If it fails (missing mjs, node not found, etc.), the sync script proceeds without the clamp active.

**Why it matters:** The very scenario this PR fixes (connection exhaustion) can still occur if the clamp silently fails.

**Suggested fix:**

```sh
if ! apply_gbrain_pool_budget; then
  echo "Warning: gbrain pool budget clamp not applied" >&2
fi
```

---

### 5. Cross-File Impact

**`scripts/lib/gbrain-pool-env.sh:17` — `BASH_SOURCE[1]` vs `BASH_SOURCE[0]` fragility**

```sh
if [[ -n "${BASH_SOURCE[1]:-}" ]]; then
  cd "$(dirname "${BASH_SOURCE[1]}")" && pwd
else
  cd "$(dirname "${BASH_SOURCE[0]}")" && pwd
fi
```

When sourced from `codex-gbrain-sync.sh`, `BASH_SOURCE[1]` is the caller and this resolves to `scripts/lib/` — correct. But if another script sources this from a different directory depth, or if the script is sourced via `source` through a symlink, the path resolution may break. The `. mjs` file lookup would fail and silently return 0.

**Why it matters:** Any new wrapper that sources this script from a non-standard path will silently skip the clamp.

**Suggested fix:** Resolve relative to the script's own location more robustly:

```sh
_gbrain_pool_env_lib_dir() {
  local src="${BASH_SOURCE[0]}"
  while [[ -L "$src" ]]; do
    src="$(readlink "$src")"
  done
  cd "$(dirname "$src")" && pwd
}
```

---

**No breaking changes to existing callers.** The new env vars (`GBRAIN_POOL_SIZE`, etc.) only affect gbrain's internal pooling behavior and don't conflict with any existing Jovie env vars. The `codex-gbrain-sync.sh` change is additive — it sources the new file and calls the new function before existing logic.

itstimwhite (commented):

itstimwhite:

### 1. Logic Errors

**`scripts/lib/gbrain-pool-env.sh:30` — `eval` of node output without validation**

The shell helper calls `node ... --shell` and `eval`s whatever it prints. If the `.mjs` file is corrupted, symlinked to something else, or the node binary is swapped, arbitrary shell code runs. The `.mjs` file's own `formatShellExports` only quotes values — it does not restrict keys. A compromised or buggy `.mjs` could emit `export FOO=value; rm -rf /`.

**Why it matters:** This is a code-execution primitive. Any process that can write to `scripts/lib/gbrain-pool-env.mjs` (or the node binary on `$PATH`) gets arbitrary shell execution in every Codex sync hook.

**Suggested fix:** Validate the emitted lines before eval, e.g.:
```bash
if [[ -n "$exports" ]]; then
  while IFS= read -r line; do
    [[ "$line" =~ ^export\ [A-Za-z_][A-Za-z0-9_]*= ]] || continue
    eval "$line"
  done <<< "$exports"
fi
```

---

**`scripts/lib/gbrain-pool-env.mjs:100` — `config-unreadable` silently swallows all errors**

`planPoolBudget` catches every exception from `JSON.parse` / `readFileSync` and returns `{ apply: false, reason: 'config-unreadable' }`. This means a valid Supabase config with a permissions error, or a transient FS issue, silently skips the clamp — the exact scenario the PR is trying to prevent.

**Why it matters:** On a Supabase brain, a missing or unreadable config should be surfaced to the operator, not silently treated as "no budget needed."

**Suggested fix:** At minimum, `console.warn` the error. Better: distinguish "file not found" (operator hasn't run `gbrain init`) from "permission denied" (real problem).

---

**`scripts/lib/gbrain-pool-env.mjs:89` — `engineNeedsPoolBudget` returns `false` for unknown engines with no URL**

If `engine` is missing/empty and no `database_url` / `pooler_url` / `direct_database_url` fields exist, the function returns `false` — no clamp applied. A Supabase brain with a minimal config (e.g. only `engine: "postgres"` but no URL) is handled, but a brain with an unrecognized engine string and a Supabase URL in a non-standard field is missed.

**Why it matters:** Low risk in practice since gbrain's config schema is stable, but the function's contract is "does this need a pool budget?" and it can silently say no for a Supabase-backed brain.

**Suggested fix:** Add a comment documenting the assumption, or add a fallback that checks `DATABASE_URL` env var.

---

### 2. Security

**`scripts/lib/gbrain-pool-env.sh:30` — `eval` of external process output (see above)**

This is the primary security concern. Covered in Logic Errors above.

---

**`scripts/lib/gbrain-pool-env.mjs:144` — `shellQuote` regex allows `/` and `:` unquoted**

The regex `/^[A-Za-z0-9_./:@%+-]+$/` passes through values containing `/` and `:` unquoted. While not directly exploitable in an `export KEY=VALUE` context (the shell won't treat those as metacharacters in the value position), it's fragile. If the output format ever changes, this becomes a vector.

**Why it matters:** Defense in depth. The regex is wider than necessary for env-var values.

**Suggested fix:** Tighten to `/^[A-Za-z0-9_@%+-]+$/` or just always single-quote — the performance cost is zero.

---

**No hardcoded secrets found.** ✅

---

### 3. Data Integrity

**None found.** This PR does not touch the database layer, Drizzle schemas, or migrations. It only sets environment variables for an external CLI.

---

### 4. Error Handling

**`scripts/lib/gbrain-pool-env.sh:28` — Missing `.mjs` file silently returns 0**

If `gbrain-pool-env.mjs` is missing (e.g. partial checkout, path issue), the function returns successfully with no warning. The Codex sync hook continues without the pool clamp.

**Why it matters:** Silent failure mode for the exact protection this PR adds.

**Suggested fix:**
```bash
if [[ ! -f "${lib_dir}/gbrain-pool-env.mjs" ]]; then
  echo "WARN: gbrain-pool-env.mjs not found; pool budget not applied" >&2
  return 0
fi
```

---

**`scripts/lib/gbrain-pool-env.sh:30` — Node stderr discarded (`2>/dev/null`)**

If the `.mjs` script errors (syntax error, missing import, etc.), the error is silently swallowed and `$exports` stays empty. No clamp is applied, no warning is shown.

**Why it matters:** Same as above — silent skip of the safety mechanism.

**Suggested fix:** Capture stderr and log it:
```bash
exports="$(node "${lib_dir}/gbrain-pool-env.mjs" --shell "$config_path" 2>/tmp/gbrain-pool-env.err || true)"
if [[ -s /tmp/gbrain-pool-env.err ]]; then
  echo "WARN: gbrain-pool-env.mjs stderr: $(cat /tmp/gbrain-pool-env.err)" >&2
fi
```

---

### 5. Cross-File Impact

**`scripts/codex-gbrain-sync.sh:15` — `apply_gbrain_pool_budget` runs unconditionally before `gbrain doctor`**

The function is called at the top level of the script, before any phase check. This means even a `gbrain sync` phase (not just `doctor`) gets the env vars set. This is intentional and correct — but worth confirming that no downstream code in `codex-gbrain-sync.sh` (not shown in diff) unsets or overrides these vars.

**Why it matters:** If any later code does `unset GBRAIN_POOL_SIZE` or similar, the clamp is silently removed.

**Suggested fix:** Audit the rest of `codex-gbrain-sync.sh` for any `unset` of `GBRAIN_*` vars. (Cannot confirm from diff alone.)

---

**No other cross-file impacts detected.** The new files are leaf dependencies — only `codex-gbrain-sync.sh` sources the `.sh` helper, and nothing imports the `.mjs` except the `.sh` wrapper and tests.

---

### Summary

|---|---|---|
| Logic errors | 2 | Medium |
| Security | 2 (1 unique + 1 overlap) | Medium-High |
| Data integrity | 0 | — |
| Error handling | 2 | Medium |
| Cross-file impact | 1 (needs manual audit) | Low |

The `eval` in `gbrain-pool-env.sh` is the highest-severity item. The silent-error-swallowing patterns are medium severity — they undermine the reliability of the safety mechanism this PR introduces.

itstimwhite (commented):

PR #10989, opened 2026-06-17
fix(ios): gate native Chat tab on server chatEnabled capability (JOV-3230)

itstimwhite:
Closes JOV-3230.

## Problem
The TestFlight app shows a **Chat tab that doesn't work**. `AppShellView` rendered the Chat tab unconditionally, but `/api/mobile/v1/chat/turns` returns `501 MOBILE_CHAT_RUNTIME_DISABLED` when the runtime is gated off / the user lacks alpha access — a dead tab. A flagged-off feature should not show a tab.

## Change
- **Backend** — new `chatEnabled` capability on `GET /api/mobile/v1/me`, computed from `MOBILE_CHAT_RUNTIME_ENABLED === 'true'` **and** mobile-chat alpha access (admin OR `ios_app_alpha_access`). Extracted the shared gate into `lib/mobile/chat/access.ts` (single source of truth) and refactored `turn-handler.ts` to reuse it (behavior-preserving).
- **iOS** — `MobileMeResponse.chatEnabled`; `AppShellView` hides the Chat tab, the menu Chat row, and the Recent section when `chatEnabled` is false, and clamps the initial tab to Profile. `RootView` passes the capability through (false pre-onboarding).

This encodes the **"alpha on by default, production off until approved"** policy: the tab appears only where chat actually works.

## Verification
- `pnpm --filter @jovie/web run typecheck` — clean (exit 0)
- `pnpm biome check` — clean (pre-push lint, 6242 files)
- Vitest (25 passing): `me.test.ts` (+chatEnabled true/false + wiring assertion), `chat-turns.test.ts` (now env-deterministic via `vi.stubEnv` + new enabled-delegation case), `chat-conversations`, `chat-contract`, `chat/turns`
- `swiftc -typecheck MobileMeResponse.swift` — clean; brace-balance verified on edited SwiftUI files
- iOS build/UI tests run in `ios-ci` (macos-26) on this PR — **no local Xcode available in the authoring environment**, so the Swift compile + Swift Testing suites are verified by CI here.

## Cost Impact
This PR is gating-only (no new recurring calls). Separately, `MOBILE_CHAT_RUNTIME_ENABLED=true` was set in Doppler dev/stg/prd this session to turn the runtime on; per-turn LLM cost is scoped to alpha testers (+admins) by the `ios_app_alpha_access` gate. Production picks up the flag on next redeploy.

## Guardrail added
`chat-turns.test.ts` no longer depends on ambient `MOBILE_CHAT_RUNTIME_ENABLED` (it stubs the env), and now covers both the disabled (501) and enabled (delegates to runtime) branches — closing the gap that let the env flag silently change test behavior.

## Summary by CodeRabbit

* **New Features**
  * Added chat feature availability gating, allowing control over which users can access the mobile chat functionality based on admin status and alpha access eligibility.

* **Chores**
  * Refactored internal chat access utilities and updated tests to cover new feature controls.

### DIFF
--- a/apps/ios/Jovie/App/RootView.swift
+++ b/apps/ios/Jovie/App/RootView.swift
@@ -194,6 +194,7 @@ private struct AppContentView: View {
           initialTab: .profile,
           opensSettingsOnLaunch: appState.launchMode.opensSettingsOnLaunch,
           billingURL: appState.billingURL,
+          chatEnabled: false,
           recentConversations: chatRepository?.conversations ?? [],
           onSelectConversation: { conversationID in
             Task { await chatRepository?.openConversation(conversationID) }
@@ -219,6 +220,7 @@ private struct AppContentView: View {
           initialTab: appState.launchMode.opensChatOnLaunch ? .chat : .profile,
           opensSettingsOnLaunch: appState.launchMode.opensSettingsOnLaunch,
           billingURL: appState.billingURL,
+          chatEnabled: appState.loadedDashboardResponse?.chatEnabled ?? false,
           recentConversations: chatRepository?.conversations ?? [],
           onSelectConversation: { conversationID in
             Task { await chatRepository?.openConversation(conversationID) }
--- a/apps/ios/Jovie/Core/MobileMeResponse.swift
+++ b/apps/ios/Jovie/Core/MobileMeResponse.swift
@@ -13,6 +13,7 @@ struct MobileMeResponse: Codable, Equatable, Sendable {
   let qrPayload: String?
   let avatarURL: String?
   let appleWalletProfilePassAvailable: Bool
+  let chatEnabled: Bool
   let continueOnWebURL: String
 
   enum CodingKeys: String, CodingKey {
@@ -23,6 +24,7 @@ struct MobileMeResponse: Codable, Equatable, Sendable {
     case qrPayload
     case avatarURL = "avatarUrl"
     case appleWalletProfilePassAvailable
+    case chatEnabled
     case continueOnWebURL = "continueOnWebUrl"
   }
 
@@ -34,6 +36,7 @@ struct MobileMeResponse: Codable, Equatable, Sendable {
     qrPayload: "https://jov.ie/tim",
     avatarURL: nil,
     appleWalletProfilePassAvailable: false,
+    chatEnabled: true,
     continueOnWebURL: "https://jov.ie/app"
   )
 
@@ -45,6 +48,7 @@ struct MobileMeResponse: Codable, Equatable, Sendable {
     qrPayload: nil,
     avatarURL: nil,
     appleWalletProfilePassAvailable: false,
+    chatEnabled: false,
     continueOnWebURL: "https://jov.ie/app"
   )
 }
--- a/apps/ios/Jovie/Features/AppShell/AppShellView.swift
+++ b/apps/ios/Jovie/Features/AppShell/AppShellView.swift
@@ -63,6 +63,7 @@ struct AppShellView<ProfileContent: View, ChatContent: View>: View {
   let isOffline: Bool
   let opensSettingsOnLaunch: Bool
   let billingURL: URL
+  let chatEnabled: Bool
   let recentConversations: [MobileConversationSummary]
   let onSelectConversation: (String) -> Void
   let onLogout: @MainActor () async -> Void
@@ -81,6 +82,7 @@ struct AppShellView<ProfileContent: View, ChatContent: View>: View {
     initialTab: AppShellTab = .profile,
     opensSettingsOnLaunch: Bool = false,
     billingURL: URL,
+    chatEnabled: Bool = false,
     recentConversations: [MobileConversationSummary] = [],
     onSelectConversation: @escaping (String) -> Void = { _ in },
     onLogout: @escaping @MainActor () async -> Void,
@@ -91,12 +93,13 @@ struct AppShellView<ProfileContent: View, ChatContent: View>: View {
     self.isOffline = isOffline
     self.opensSettingsOnLaunch = opensSettingsOnLaunch
     self.billingURL = billingURL
+    self.chatEnabled = chatEnabled
     self.recentConversations = recentConversations
     self.onSelectConversation = onSelectConversation
     self.onLogout = onLogout
     self.profileContent = profileContent()
     self.chatContent = chatContent
-    _selectedTab = State(initialValue: initialTab)
+    _selectedTab = State(initialValue: chatEnabled ? initialTab : .profile)
   }
 
   var body: some View {
@@ -134,6 +137,7 @@ struct AppShellView<ProfileContent: View, ChatContent: View>: View {
         profile: profile,
         isOffline: isOffline,
         selectedTab: selectedTab,
+        chatEnabled: chatEnabled,
         recentConversations: recentConversations,
         onSelectTab: { tab in
           selectedTab = tab
@@ -163,7 +167,11 @@ struct AppShellView<ProfileContent: View, ChatContent: View>: View {
   private var activeContent: some View {
     switch selectedTab {
     case .chat:
-      chatContent($chatDraft)
+      if chatEnabled {
+        chatContent($chatDraft)
+      } else {
+        profileContent
+      }
     case .profile:
       profileContent
     }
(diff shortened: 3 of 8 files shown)


itstimwhite:

### Logic Errors

**1. `access.ts:57` — `isAdmin` check short-circuits before `hasMobileChatAlphaAccess` refetch, but `turn-handler.ts` still has its own copy of the alpha check**

The new `isMobileChatEnabledForUser` in `access.ts` correctly short-circuits on `options?.isAdmin`. However, `turn-handler.ts:98` (the `hasMobileChatAlphaAccess` function that was *kept* in the file after the refactor — it was not removed, only the `getSessionContext`/`getAppFlagValue` imports were replaced) still independently calls `getSessionContext` and `getAppFlagValue`. This means the turn handler does **not** use `isMobileChatEnabledForUser` from `access.ts` — it still uses its own local `hasMobileChatAlphaAccess`. The PR description says "refactored `turn-handler.ts` to reuse it" but the diff shows the local function was merely deduplicated in imports, not replaced with a call to the shared `isMobileChatEnabledForUser`.

**Why it matters:** The turn endpoint gates on `hasMobileChatAlphaAccess` (alpha-only), while the `me` endpoint gates on `isMobileChatEnabledForUser` (runtime flag AND alpha). If `MOBILE_CHAT_RUNTIME_ENABLED=true` but a user lacks alpha access, the `me` endpoint correctly returns `chatEnabled: false`, but the turn endpoint's local `hasMobileChatAlphaAccess` would still return `false` — so the user gets a 403 from the turn handler, not a 501 from the route. This is actually *correct behavior* (the route-level gate returns 501 only when the runtime flag is off; when the flag is on but the user lacks alpha, the turn handler's own gate returns 403). But the inconsistency means the route-level 501 gate in `route.ts` only checks `isMobileChatRuntimeEnabled()` — it does **not** check per-user alpha access. So a user without alpha access but with a valid session will hit the turn handler and get a 403, while the tab is hidden. This is fine UX-wise (tab is hidden), but the route-level gate and the handler-level gate are checking different things, which could confuse future maintainers.

**Verdict:** Not a bug per se, but the PR description is misleading — `turn-handler.ts` was not refactored to reuse the shared gate. Consider replacing the local `hasMobileChatAlphaAccess` in `turn-handler.ts` with a call to `isMobileChatEnabledForUser` for true single-source-of-truth, or at minimum add a comment explaining the intentional split.

---

**2. `access.ts:53` — `isMobileChatEnabledForUser` doesn't check `isAdmin` when `isAdmin` is `false` (not just missing)**

```typescript
if (options?.isAdmin) {
  return true;
}
```

If the caller passes `{ isAdmin: false }`, this check falls through to `getAppFlagValue('IOS_APP_ALPHA_ACCESS', ...)` — which is correct. But if the caller passes `{}` or `{ isAdmin: undefined }`, `options?.isAdmin` is `undefined`, which is falsy, so it also falls through — again correct. This is fine.

**Verdict:** No bug. The logic is correct.

---

**3. `AppShellView.swift:101` — Initial tab clamping doesn't account for `.chat` being selected programmatically after init**

```swift
_selectedTab = State(initialValue: chatEnabled ? initialTab : .profile)
```

If `chatEnabled` changes from `false` to `true` after the view is initialized (e.g., after a dashboard refresh), `selectedTab` won't update because `@State` is only set from its initial value. However, looking at the architecture, `AppShellView` is recreated when `appState.loadedDashboardResponse` changes (it's passed as a prop from `RootView`), so this is fine — SwiftUI will recreate the view struct with the new `chatEnabled` value.

**Verdict:** No bug.

---

**4. `AppShellView.swift:170-174` — Fallback to `profileContent` when `selectedTab == .chat` but `chatEnabled == false`**

```swift
case .chat:
  if chatEnabled {
    chatContent($chatDraft)
  } else {
    profileContent
  }
```

This is a safety net, but it means if `selectedTab` somehow becomes `.chat` while `chatEnabled` is `false` (e.g., a race condition or stale state), the user sees the profile content while the tab bar shows "Chat" selected. This is a graceful fallback, but it could be confusing. The `bottomNavigation` view already hides the tab when `chatEnabled` is false, so this state should be unreachable in practice.

**Verdict:** No bug — defensive coding, acceptable.

---

**5. `me/route.ts:106-108` — `isMobileChatEnabledForUser` is called with `isAdmin` from `session.user.isAdmin`**

```typescript
const chatEnabled = await isMobileChatEnabledForUser(userId, {
  isAdmin: session.user.isAdmin,
});
```

This is correct — the session has already been resolved, so passing `isAdmin` avoids a refetch. Good.

**Verdict:** No bug.

---

**6. `me/route.ts:32-39` — `buildNeedsOnboardingResponse` hardcodes `chatEnabled: false`**

This is called when the user hasn't completed onboarding. Since the user hasn't completed onboarding, they shouldn't have chat access anyway. Correct.

**Verdict:** No bug.

---

### Security

**1. No hardcoded secrets, SQL injection, XSS, or auth bypass vectors found.**

- The `MOBILE_CHAT_RUNTIME_ENABLED` env var check is a simple string comparison — safe.
- The `isAdmin` flag comes from the Clerk session, not user input.
- The `IOS_APP_ALPHA_ACCESS` flag is checked via `getAppFlagValue` with the Clerk userId — safe.
- No new user-input handling in this PR.

**Verdict:** None found.

---

### Data Integrity

**1. `MobileMeResponse.swift:36` — Preview mock has `chatEnabled: true`**

```swift
static let preview = MobileMeResponse(
    ...
    chatEnabled: true,
    ...
)
```

This is the SwiftUI preview mock. It's fine for development, but if someone uses this preview to test the "chat disabled" state, they'll see the Chat tab. This is a preview-only concern.

**Verdict:** Minor — consider adding a `previewChatDisabled` static for completeness, but not a blocker.

**2. `MobileMeResponse.swift:48` — `empty` mock has `chatEnabled: false`**

Correct — the empty/disabled state should have chat disabled.

**Verdict:** No issue.

---

### Error Handling

**1. `access.ts:30-50` — `hasMobileChatAlphaAccess` calls `getSessionContext` with `requireUser: true`**

If the user doesn't exist, this will throw. But `isMobileChatEnabledForUser` is only called with a valid `clerkUserId` (from the session), so this should be fine in practice. The `me/route.ts` call site already has a valid session.

**Verdict:** No issue.

**2. `access.ts:57` — `getAppFlagValue` could throw if Statsig is unreachable**

There's no try/catch around `getAppFlagValue`. If it throws, the entire `isMobileChatEnabledForUser` call throws, which would bubble up to the `me` endpoint and return a 500. This is consistent with how other flag checks behave in the codebase (fail-open vs. fail-closed is a product decision). Given that this is a feature gate, a 500 is arguably better than silently enabling chat.

**Verdict:** No issue — consistent with codebase patterns.

---

### Cross-File Impact

**1. `turn-handler.ts` — Local `hasMobileChatAlphaAccess` function still exists**

The diff shows that `turn-handler.ts` still has its own `hasMobileChatAlphaAccess` function (lines 98-113 in the original, now with updated imports). The PR description says this was "refactored to reuse it" but the function was not replaced with a call to the shared `isMobileChatEnabledForUser`. This means:

- The turn handler checks alpha access but **not** the runtime flag (the runtime flag is checked at the route level in `route.ts`).
- The `me` endpoint checks both runtime flag AND alpha access.

This is actually the correct separation of concerns — the route handles the runtime flag, the handler handles per-user access. But the PR description is misleading.

**Verdict:** Not a bug, but the PR description should be corrected to avoid confusion.

**2. No other callers of `hasMobileChatAlphaAccess` or `isMobileChatRuntimeEnabled` were found in the diff.**

**Verdict:** No cross-file impact issues.

---

### Summary

| Category | Issues Found |
|----------|-------------|
| Logic Errors | 1 minor (misleading PR description about turn-handler refactor) |
| Security | None found |
| Data Integrity | 1 minor (preview mock has chatEnabled: true) |
| Error Handling | None found |
| Cross-File Impact | None found (but PR description is misleading) |

**Overall:** The PR is logically sound and introduces no security vulnerabilities or data integrity issues. The main concern is that the PR description claims `turn-handler.ts` was refactored to reuse the shared gate, but it still has its own local `hasMobileChatAlphaAccess` function. This is not a bug (the separation is intentional and correct), but the description should be updated for accuracy.

itstimwhite (commented):

PR #11000, opened 2026-06-17
fix(auth): bound audience-block DB queries in proxy hot path (JOV-3231)

itstimwhite:
## Summary
Caps the two audience-block Drizzle queries (`profileHasActiveBlocks`, `isVisitorBlockedByFingerprint`) at 3000ms via the existing `withTimeout` util. These run inside `checkProfileVisitorBlocked`, awaited in `proxy.ts` on every single-segment public-profile navigation. Previously uncapped — a Neon cold start on a cold audience-block cache stalled the entire middleware (~3-10s) for every visitor to that profile.

## Changes
- `withTimeout(...)` around both DB queries (reuses `@/lib/db/query-timeout`, same pattern as `proxy-state.ts`).
- Fixed latent error-handling bug: `return await isVisitorBlockedByFingerprint(...)` so a timeout error is caught by the local fail-open `catch`.
- Regression test for fail-open-on-timeout.

## Cost Impact
None. No new external calls — this strictly reduces worst-case middleware latency.

## Test evidence

```
✓ tests/unit/lib/audience/public-profile-block.test.ts (10 tests) 6ms

 Test Files  1 passed (1)
      Tests  10 passed (10)
   Start at  19:58:55
   Duration  611ms
```

Typecheck: clean (exit 0)
Biome: `Checked 2 files in 48ms. No fixes applied.`

bug-to-test: regression test added (fail-open after DB query timeout).

### DIFF
--- a/apps/web/lib/audience/public-profile-block.ts
+++ b/apps/web/lib/audience/public-profile-block.ts
@@ -2,9 +2,15 @@ import 'server-only';
 
 import * as Sentry from '@sentry/nextjs';
 import { createFingerprintEdge } from '@/lib/audience/fingerprint';
+import { withTimeout } from '@/lib/db/query-timeout';
 import { env } from '@/lib/env-server';
 import { captureWarning } from '@/lib/error-tracking';
 import { getRedis } from '@/lib/redis';
+
+// Timeout for audience-block DB queries. Matches proxy-state.ts budget — kept
+// below the Neon p99 cold-start budget (~3 s) so a cache-miss does not stall
+// every visitor navigation for more than ~3 s. Fails open on timeout.
+const AUDIENCE_BLOCK_DB_QUERY_TIMEOUT_MS = 3000;
 
 /**
  * Mirror extractClientIP() priority for the middleware audience-block check.
@@ -219,7 +225,7 @@ async function profileHasActiveBlocks(username: string): Promise<boolean> {
   const { audienceBlocks } = await import('@/lib/db/schema/analytics');
   const { creatorProfiles } = await import('@/lib/db/schema/profiles');
 
-  const [result] = await db
+  const queryPromise = db
     .select({ profileId: creatorProfiles.id })
     .from(creatorProfiles)
     .where(
@@ -240,6 +246,12 @@ async function profileHasActiveBlocks(username: string): Promise<boolean> {
     )
     .limit(1);
 
+  const [result] = await withTimeout(
+    queryPromise,
+    AUDIENCE_BLOCK_DB_QUERY_TIMEOUT_MS,
+    '[audience-block] profileHasActiveBlocks'
+  );
+
   return !!result;
 }
 
@@ -252,7 +264,7 @@ async function isVisitorBlockedByFingerprint(
   const { audienceBlocks } = await import('@/lib/db/schema/analytics');
   const { creatorProfiles } = await import('@/lib/db/schema/profiles');
 
-  const [result] = await db
+  const queryPromise = db
     .select({ blockId: audienceBlocks.id })
     .from(creatorProfiles)
     .innerJoin(
@@ -267,6 +279,12 @@ async function isVisitorBlockedByFingerprint(
       )
     )
     .limit(1);
+
+  const [result] = await withTimeout(
+    queryPromise,
+    AUDIENCE_BLOCK_DB_QUERY_TIMEOUT_MS,
+    '[audience-block] isVisitorBlockedByFingerprint'
+  );
 
   return !!result;
 }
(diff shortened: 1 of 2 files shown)


itstimwhite:

### Logic Errors

**1. `public-profile-block.ts:246` — `profileHasActiveBlocks` query construction is split across `withTimeout` boundary, but the Drizzle query builder is lazy**

The diff separates the Drizzle query construction (`db.select(...).from(...).where(...)`) from the `await withTimeout(queryPromise, ...)` call. Drizzle's query builder is lazy — the SQL is only executed when the promise is awaited. This means `withTimeout` correctly wraps the execution, not the construction. This is actually correct behavior. **No issue here.**

**2. `public-profile-block.ts:324` — `return await` fix is correct but subtly changes behavior for synchronous throws**

Before: `return isVisitorBlockedByFingerprint(...)` — if `isVisitorBlockedByFingerprint` threw synchronously (before any `await`), the error would propagate as a rejected promise from `checkProfileVisitorBlocked`'s `catch` block, returning `false`. After: `return await isVisitorBlockedByFingerprint(...)` — a synchronous throw is now caught by the `try/catch` in `checkProfileVisitorBlocked`, also returning `false`. Behavior is the same in practice since `isVisitorBlockedByFingerprint` is `async` and its first operation is `await`, but the `return await` is the correct pattern for ensuring the `catch` block handles all error paths. **No issue — this is a correct fix.**

**3. `public-profile-block.ts:246` — `profileHasActiveBlocks` has the same `return await` pattern already implicit via destructuring**

The `const [result] = await withTimeout(...)` already awaits inside the function, so any timeout error from `withTimeout` will propagate up to the caller (`checkProfileVisitorBlocked`) where the `catch` returns `false`. This is correct. **No issue.**

**None found** — the logic changes are sound.

---

### Security

**None found.** No hardcoded secrets, no SQL injection vectors (Drizzle parameterized queries unchanged), no auth bypass (fail-open on timeout is the safe default — blocked visitors are let through rather than blocking everyone), no CSP changes.

---

### Data Integrity

**None found.** The DB queries are unchanged — only the execution wrapper changed. No schema changes, no migration issues.

---

### Error Handling

**1. `public-profile-block.ts:324` — The `catch` block in `checkProfileVisitorBlocked` is a bare `catch` with no logging**

This pre-existing issue is not introduced by this PR, but the PR's `return await` fix now means timeout errors from `isVisitorBlockedByFingerprint` land here. The bare `catch { return false; }` silently swallows `QueryTimeoutError`. This is intentional fail-open behavior, but there's no Sentry warning or breadcrumb for the timeout case specifically. The PR description mentions this is "fail-open on timeout" by design, and `withTimeout` likely logs internally. **Not a new issue introduced by this PR — pre-existing pattern. No action needed.**

**None found** — no new error handling issues introduced.

---

### Cross-File Impact

**1. `withTimeout` import from `@/lib/db/query-timeout` — verify the export exists**

The PR imports `withTimeout` from `@/lib/db/query-timeout` and notes it's the same pattern used in `proxy-state.ts`. Assuming that module exists and exports `withTimeout`, this is fine. The test mocks it, confirming the import path. **No issue, assuming the module exists as stated.**

**2. Test file — `mocks.select.mockReset()` in `beforeEach`**

The diff adds `mocks.select.mockReset()` in `beforeEach` to clear persistent `mockImplementation` from prior tests. However, `mockReset` clears the mock's implementation AND return values, which means the default `mockReturnValue({ from: ... })` chain set up by `mockProfileHasBlocksRows` in individual tests must re-establish the full chain. Looking at the test helper `mockProfileHasBlocksRows`, it calls `mocks.select.mockReturnValueOnce(...)` which should work fine after `mockReset`. **No issue.**

**3. Test — `mockProfileHasBlocksRows([])` called before the inner select mock**

In the new test at line ~291, `mockProfileHasBlocksRows([])` sets up the outer select chain, then `mocks.select.mockReturnValueOnce({ from: innerFrom })` sets up the inner exists-subquery select. But `mockProfileHasBlocksRows` internally calls `mocks.select.mockReturnValueOnce(...)` which consumes one `select` call. The inner select mock is set up after, consuming the second `select` call. The order matters: the outer query's `db.select()` is called first, then the inner `exists` subquery's `db.select()` is called second. This matches the Drizzle query construction order in `profileHasActiveBlocks`. **No issue — ordering is correct.**

**None found** — no cross-file breakage.

---

### Summary

**No issues found.** The PR is clean:
- Timeout wrapping is correctly applied around the lazy Drizzle query execution (not construction)
- The `return await` fix ensures timeout errors are caught by the fail-open `catch`
- The regression test correctly simulates the timeout scenario and validates fail-open behavior
- No security, data integrity, or cross-file concerns

itstimwhite (commented):

PR #11002, opened 2026-06-17
feat(ios): bottom-bar-centric navigation — floating icon bar + swipe + overflow drawer (JOV-3233)

itstimwhite:
Closes JOV-3233. **Stacked on #10989** (JOV-3230 chat-tab gating) — this branch includes #10989's commit; rebase onto main after it lands so only the nav diff remains.

## What
Replaces the split nav (top hamburger + bottom text-pill) with a **Linear-style floating, icon-only bottom bar** and intuitive swipe between screens.

- **Bottom bar**: Profile + Chat in one capsule with a selected rounded-square highlight; a separate **More** capsule (•••) opens the overflow drawer. Icon-only, floating.
- **Swipe**: Profile ⇄ Chat slide via a directional transition + a `simultaneous` horizontal gesture that only fires on a clearly-horizontal end swipe, so vertical scrolling is never blocked; synced to the bar selection.
- **Top bar** simplified to a large left-aligned title + settings affordance (hamburger removed).
- **Drawer** drops the redundant Chat/Profile rows; keeps account, Settings, Recent chats.
- Respects the JOV-3230 `chatEnabled` gate.

## Verification (local, Xcode 26.3)
- `xcodebuild build` (iOS Simulator) — BUILD SUCCEEDED
- `JovieUITests` nav suite — TEST SUCCEEDED on iPhone 16 Pro + iPhone 17: ready/profile, tab switch, **swipe (new)**, draft preservation, menu+settings
- Profile + Chat simulator screenshots captured (floating bar, moving selection, separate More capsule)
- Full `ios-ci` runs on this PR

## Notes
- UI test affordances updated: `Open Profile`→`shell-tab-profile`, `Open Menu`→`More`; `shell-tab-*` + `Open Settings` preserved.
- No layout shift: bar/toolbar are stable safe-area insets; only the content slides.

### DIFF
--- a/apps/ios/Jovie/Features/AppShell/AppShellView.swift
+++ b/apps/ios/Jovie/Features/AppShell/AppShellView.swift
@@ -107,14 +107,19 @@ struct AppShellView<ProfileContent: View, ChatContent: View>: View {
       ZStack {
         JovieColor.backgroundBase.ignoresSafeArea()
 
-        activeContent
-          .safeAreaInset(edge: .top, spacing: 0) {
-            shellToolbar
-          }
-          .safeAreaInset(edge: .bottom, spacing: 0) {
-            bottomNavigation
-          }
+        pagedContent
+          .id(selectedTab)
+          .transition(pageTransition)
+          .frame(maxWidth: .infinity, maxHeight: .infinity)
+          .clipped()
       }
+      .safeAreaInset(edge: .top, spacing: 0) {
+        shellToolbar
+      }
+      .safeAreaInset(edge: .bottom, spacing: 0) {
+        bottomBar
+      }
+      .simultaneousGesture(pageSwipe)
       .navigationBarHidden(true)
       .navigationDestination(for: AppShellRoute.self) { route in
         switch route {
@@ -130,19 +135,12 @@ struct AppShellView<ProfileContent: View, ChatContent: View>: View {
       }
     }
     .background(JovieColor.backgroundBase)
-    .contentShape(Rectangle())
-    .simultaneousGesture(edgeSwipe)
     .fullScreenCover(isPresented: $isShowingMenu) {
       AppNavigationMenu(
         profile: profile,
         isOffline: isOffline,
-        selectedTab: selectedTab,
         chatEnabled: chatEnabled,
         recentConversations: recentConversations,
-        onSelectTab: { tab in
-          selectedTab = tab
-          isShowingMenu = false
-        },
         onSelectConversation: { conversationID in
           onSelectConversation(conversationID)
           selectedTab = .chat
@@ -163,8 +161,12 @@ struct AppShellView<ProfileContent: View, ChatContent: View>: View {
     }
   }
 
+  // Horizontally paged primary screens. The floating bottom bar drives the same
+  // `selectedTab` selection so taps and the horizontal swipe gesture stay in sync,
+  // and each screen keeps its own vertical scrolling (the swipe is recognized
+  // simultaneously and only acts on a clearly-horizontal end gesture).
   @ViewBuilder
-  private var activeContent: some View {
+  private var pagedContent: some View {
     switch selectedTab {
     case .chat:
       if chatEnabled {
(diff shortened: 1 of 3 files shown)


itstimwhite:

### Logic Errors

**1. `AppShellView.swift:93` — `chatEnabled` default initializer silently forces `.profile` even when caller passes `initialTab: .chat`**

The initializer body at line 93 does:
```swift
_selectedTab = State(initialValue: chatEnabled ? initialTab : .profile)
```
But the default value for `chatEnabled` in the initializer signature (line 85) is `false`. If a caller uses the default initializer and passes `initialTab: .chat` without explicitly passing `chatEnabled: true`, the tab silently snaps to `.profile` with no warning. This is a logic trap — the two parameters are coupled but there's no assertion or compiler enforcement.

**Why it matters:** Future callers (or tests) that use the convenience initializer with just `initialTab` will get surprising behavior. The `RootView.swift` call sites always pass both, so it's safe today, but the API is footgun-prone.

**Suggested fix:** Either remove the default value for `chatEnabled` so callers must always provide it, or add a `precondition(chatEnabled || initialTab != .chat, "…")` in the initializer.

---

**2. `AppShellView.swift:107` — `.id(selectedTab)` causes full view teardown on tab switch**

The diff adds `.id(selectedTab)` on the `pagedContent` view. In SwiftUI, `.id()` causes the view to be fully destroyed and recreated when the id changes. This means any local `@State` inside `profileContent` or `chatContent` (scroll position, transient UI state, `chatDraft` binding) will be lost on every tab switch.

**Why it matters:** The PR description explicitly claims "draft preservation" is verified, and the UI test checks it. But `.id(selectedTab)` will destroy the chat view (and its draft state) every time the user taps Profile. The test may pass because `chatDraft` is held in `AppShellView` as `@State` and passed as a binding — but any *other* internal state in the chat view (scroll position, composer focus, etc.) will be lost. This is a latent bug that will surface as the chat feature grows.

**Suggested fix:** Remove `.id(selectedTab)`. The `.transition()` modifier already handles the visual swap. If the id was added to force the transition to re-fire, use a different mechanism (e.g., a `Transaction` or an explicit `withAnimation` wrapper around the `selectedTab` mutation, which is already present in `navIcon`).

---

**3. `access.ts:57` — `isMobileChatEnabledForUser` passes `isAdmin: undefined` when `options` is omitted**

When `options` is `undefined` (no second argument), the check `options?.isAdmin` evaluates to `undefined`, which is falsy — so it falls through to the Statsig flag check. This is correct behavior. However, the function signature allows `options?: { readonly isAdmin?: boolean }`, meaning a caller could pass `{ isAdmin: undefined }` and get the same result. This is fine but worth noting: there's no explicit handling for the case where `isAdmin` is explicitly `false` vs. absent. The current code treats both identically, which is correct for the call site in `route.ts` (which passes `{ isAdmin: session.user.isAdmin }`), but the API is ambiguous.

**Why it matters:** Low risk today, but the `options` pattern is a code smell — it's a manual reimplementation of what the session already knows. Consider inlining the admin check at the call site instead.

**Suggested fix:** Not blocking, but consider simplifying to always check the flag and let the admin check happen via the Statsig gate (admins can be added to the gate). This removes the `options` parameter entirely.

---

### Security

**None found.**

- The `chatEnabled` gate is properly enforced server-side in `route.ts` (via `isMobileChatEnabledForUser`) and the iOS client merely uses it for UI gating. The actual chat turn endpoint in `turn-handler.ts` independently checks `hasMobileChatAlphaAccess`, so even if the client is tampered with, the server rejects unauthorized requests.
- The `MOBILE_CHAT_RUNTIME_ENABLED` env var is checked server-side before any alpha-cohort check, providing defense in depth.
- No hardcoded secrets, no SQL injection vectors, no XSS concerns in this diff.

---

### Data Integrity

**4. `route.ts:32` — `buildNeedsOnboardingResponse` hardcodes `chatEnabled: false`**

This is correct behavior (onboarding-incomplete users should never see chat), but it's worth flagging that there's no test asserting this specific field in the onboarding response. The test at `me.test.ts:154` does check the onboarding response shape but I can't verify from the diff whether `chatEnabled` is asserted there.

**Why it matters:** If someone later adds a code path where onboarding-incomplete users could have chat, this hardcoded `false` would mask the issue.

**Suggested fix:** Add an explicit assertion for `chatEnabled: false` in the onboarding test case. Low priority.

---

### Error Handling

**5. `access.ts:28` — `hasMobileChatAlphaAccess` calls `getSessionContext` with `requireUser: true` but has no try/catch**

If `getSessionContext` throws (e.g., DB unreachable), the error propagates unhandled. This is consistent with the rest of the codebase pattern (the callers in `route.ts` and `turn-handler.ts` handle errors at the route level), so it's not a new problem.

**Why it matters:** None beyond existing patterns. The route-level error handling in `route.ts` (the try/catch around the main logic) will catch this.

**Suggested fix:** None needed if route-level error handling is already in place (which it is, per the existing code).

---

### Cross-File Impact

**6. `turn-handler.ts:98` — `hasMobileChatAlphaAccess` is removed from this file and imported from `access.ts`**

The local duplicate is correctly removed and replaced with the shared import. The function signature is identical. The `turn-handler.ts` call site at line 98 previously called the local version with just `clerkUserId` (no `options`), and the imported version handles this correctly (falls through to Statsig check). ✅

**7. `AppShellView.swift:159` — `AppNavigationMenu` call site drops `selectedTab` and `onSelectMenuTab`**

The call site no longer passes `selectedTab` or `onSelectTab` to `AppNavigationMenu`. The `AppNavigationMenu` initializer signature (not fully shown in diff) must have these as optional or have been updated. The diff shows the `AppNavigationMenu` struct no longer declares `selectedTab` or `onSelectTab` as stored properties (they're removed from the struct body at the bottom of the diff). This is consistent. ✅

**8. `RootView.swift:197` — Offline path hardcodes `chatEnabled: false`**

Correct behavior: when offline, chat is unavailable. The online path at line 223 correctly reads from `appState.loadedDashboardResponse?.chatEnabled`. ✅

---

### Summary

| 1 | Medium | `AppShellView.swift:93` | Default `chatEnabled: false` silently overrides `initialTab: .chat` — footgun API |
| 2 | **High** | `AppShellView.swift:107` | `.id(selectedTab)` destroys view state on every tab switch, breaking draft/scroll preservation |
| 3 | Low | `access.ts:57` | `options` parameter pattern is ambiguous but functionally correct |
| 4 | Low | `route.ts:32` | Hardcoded `chatEnabled: false` in onboarding response — correct but untested |
| 5 | None | `access.ts:28` | No local try/catch — acceptable given route-level handling |
| 6 | ✅ | `turn-handler.ts:98` | Clean dedup of `hasMobileChatAlphaAccess` — correct |
| 7 | ✅ | `AppShellView.swift:159` | Menu call site consistent with struct changes |
| 8 | ✅ | `RootView.swift:197` | Offline `chatEnabled: false` — correct |

**Issue #2 is the most consequential** — `.id(selectedTab)` will cause the chat view to fully reconstruct on every tab switch, which directly contradicts the stated goal of draft preservation and will cause subtle state-loss bugs as the chat feature grows.

itstimwhite (commented):

PR #11006, opened 2026-06-17
fix(auth): proxy hot-path bug-fix + perf dedupe (JOV-3231, JOV-3232)

itstimwhite:
## Summary
Ships the two reviewed proxy-middleware fixes directly to main (the integration/loop-auth train is stale/conflicted; these are cleanly cherry-picked off main). Both were already reviewed and CI-green as #11000 and #10999 against integration/loop-auth.

## JOV-3231 — bug fix (high impact)
`apps/web/lib/audience/public-profile-block.ts`: the two audience-block Drizzle queries were uncapped; `checkProfileVisitorBlocked` runs in proxy.ts on every single-segment public-profile navigation. On a Neon cold start with a cold cache this stalled the entire middleware ~3-10s for every visitor. Now bounded at 3s via the existing `withTimeout` util (fail-open, contract-consistent), plus a latent missing-`await` fix so timeout errors hit the local `catch`. Regression test added.

## JOV-3232 — perf hygiene (behavior-preserving)
- Removed dead `_getGeoFromRequest` from proxy.ts.
- Bounded memoization of pure helpers recomputed 2-4×/request: `resolveClerkKeys` (cached only when both keys present — cold-start safe; FIFO cap 50), `analyzeHost` (cap 50), `categorizePath` (cap 1000, FIFO).

## Cost Impact
None. No new external calls — strictly reduces worst-case middleware latency + redundant per-request CPU.

## Verification (local, off main)
- typecheck: exit 0
- biome: 6 files clean
- vitest: 156 passed (both new test files + full tests/unit/middleware/ suite)

## Risk / merge
Auth/proxy surface → `testing`, full CI. Attended merge after CI green + bot reviews clear; no auto-merge.

### DIFF
--- a/apps/web/lib/audience/public-profile-block.ts
+++ b/apps/web/lib/audience/public-profile-block.ts
@@ -2,9 +2,15 @@ import 'server-only';
 
 import * as Sentry from '@sentry/nextjs';
 import { createFingerprintEdge } from '@/lib/audience/fingerprint';
+import { withTimeout } from '@/lib/db/query-timeout';
 import { env } from '@/lib/env-server';
 import { captureWarning } from '@/lib/error-tracking';
 import { getRedis } from '@/lib/redis';
+
+// Timeout for audience-block DB queries. Matches proxy-state.ts budget — kept
+// below the Neon p99 cold-start budget (~3 s) so a cache-miss does not stall
+// every visitor navigation for more than ~3 s. Fails open on timeout.
+const AUDIENCE_BLOCK_DB_QUERY_TIMEOUT_MS = 3000;
 
 /**
  * Mirror extractClientIP() priority for the middleware audience-block check.
@@ -219,7 +225,7 @@ async function profileHasActiveBlocks(username: string): Promise<boolean> {
   const { audienceBlocks } = await import('@/lib/db/schema/analytics');
   const { creatorProfiles } = await import('@/lib/db/schema/profiles');
 
-  const [result] = await db
+  const queryPromise = db
     .select({ profileId: creatorProfiles.id })
     .from(creatorProfiles)
     .where(
@@ -240,6 +246,12 @@ async function profileHasActiveBlocks(username: string): Promise<boolean> {
     )
     .limit(1);
 
+  const [result] = await withTimeout(
+    queryPromise,
+    AUDIENCE_BLOCK_DB_QUERY_TIMEOUT_MS,
+    '[audience-block] profileHasActiveBlocks'
+  );
+
   return !!result;
 }
 
@@ -252,7 +264,7 @@ async function isVisitorBlockedByFingerprint(
   const { audienceBlocks } = await import('@/lib/db/schema/analytics');
   const { creatorProfiles } = await import('@/lib/db/schema/profiles');
 
-  const [result] = await db
+  const queryPromise = db
     .select({ blockId: audienceBlocks.id })
     .from(creatorProfiles)
     .innerJoin(
@@ -267,6 +279,12 @@ async function isVisitorBlockedByFingerprint(
       )
     )
     .limit(1);
+
+  const [result] = await withTimeout(
+    queryPromise,
+    AUDIENCE_BLOCK_DB_QUERY_TIMEOUT_MS,
+    '[audience-block] isVisitorBlockedByFingerprint'
+  );
 
   return !!result;
 }
@@ -306,7 +324,7 @@ export async function checkProfileVisitorBlocked(
     }
 
     const fingerprint = await createFingerprintEdge(ip, ua);
-    return isVisitorBlockedByFingerprint(normalizedUsername, fingerprint);
+    return await isVisitorBlockedByFingerprint(normalizedUsername, fingerprint);
   } catch {
     return false;
   }
--- a/apps/web/lib/auth/staging-clerk-keys.ts
+++ b/apps/web/lib/auth/staging-clerk-keys.ts
@@ -34,11 +34,47 @@ function runtimePublicEnv(key: string): string | undefined {
 }
 
 /**
+ * Module-scope memoization cache for resolveClerkKeys.
+ *
+ * Env vars are fixed at container start — caching avoids repeated env reads
+ * on every middleware invocation (3-4× per request).
+ *
+ * SAFETY: only entries with BOTH keys present (status === 'ok') are stored.
+ * Incomplete / degraded statuses ('staging_missing', 'staging_inherits_prod',
+ * 'no_publishable_key', 'missing') are returned directly and never stored, so
+ * a transient env-miss on a cold start cannot be pinned for the worker lifetime.
+ *
+ * BOUNDED: capped at 50 entries with FIFO eviction. Real hostnames are few
+ * (jov.ie, staging.jov.ie, localhost, preview URLs), but `hostname` derives
+ * from the attacker-controllable Host header — the cap defends against a
+ * scanner spraying unique Host values that would otherwise grow the Map until
+ * the worker OOMs.
+ *
+ * Exported for test teardown only. Do not call from production code.
+ */
+const RESOLVE_CLERK_KEYS_CACHE_MAX = 50;
+export const _resolveClerkKeysCache = new Map<string, ClerkKeys>();
+
+/**
+ * Store a resolved key set with FIFO eviction so the cache stays bounded.
+ * Only called for status:'ok' results (both keys present).
+ */
+function cacheClerkKeys(hostname: string, result: ClerkKeys): void {
+  if (_resolveClerkKeysCache.size >= RESOLVE_CLERK_KEYS_CACHE_MAX) {
+    const firstKey = _resolveClerkKeysCache.keys().next().value;
+    if (firstKey !== undefined) _resolveClerkKeysCache.delete(firstKey);
+  }
+  _resolveClerkKeysCache.set(hostname, result);
+}
+
+/**
  * Resolve Clerk keys for a given hostname.
  * Returns staging keys when on a staging host and staging keys are configured.
  * Staging must never silently fall back to production keys.
  */
 export function resolveClerkKeys(hostname: string): ClerkKeys {
+  const cached = _resolveClerkKeysCache.get(hostname);
+  if (cached) return cached;
   if (isStagingHost(hostname)) {
     const explicitPk = process.env.CLERK_PUBLISHABLE_KEY_STAGING;
     const explicitSk = process.env.CLERK_SECRET_KEY_STAGING;
@@ -52,11 +88,13 @@ export function resolveClerkKeys(hostname: string): ClerkKeys {
           status: 'staging_missing',
         };
       }
-      return {
+      const result: ClerkKeys = {
         publishableKey: explicitPk,
         secretKey: explicitSk,
         status: 'ok',
       };
+      cacheClerkKeys(hostname, result);
+      return result;
     }
 
     // No _STAGING vars at all: fall back to the standard env vars read at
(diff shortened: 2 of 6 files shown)


itstimwhite:

### Logic Errors

**1. `staging-clerk-keys.ts` — `cacheClerkKeys` FIFO eviction has an off-by-one: eviction happens at `>=` before insert, so the 51st entry evicts the 1st, then inserts, keeping size at 50. This is correct behavior.** No issue — just confirming the math: size 50 → evict 1 → size 49 → insert → size 50. Correct.

**2. `proxy-routing.ts` — `categorizePath` cache stores a mutable reference.** The JSDoc comment notes "the only spread site (clerk-middleware-bypass.ts) produces a new object from the spread." However, the `PathCategory` object returned from `categorizePath` is shared across all callers that hit the cache. If any caller mutates the returned object (e.g., `result.isAuthPath = true`), every subsequent caller for that pathname gets the corrupted value. The comment is a claim, not a guarantee. This is a latent risk, not a current bug — but worth noting since the codebase has no runtime enforcement (Object.freeze or a Proxy). **Low severity, no current callers mutate, but the design is fragile.**

**3. `public-profile-block.ts:324` — `return await` added correctly.** The diff shows `return await isVisitorBlockedByFingerprint(...)` at line 324. This ensures the rejected promise from `withTimeout` is caught by the surrounding `try/catch` in `checkProfileVisitorBlocked`. Without `return await`, the rejected promise would escape as an unhandled rejection. **This is correct.**

**4. `public-profile-block.ts` — `withTimeout` wraps the query promise, but the inner Drizzle query is already started.** When `db.select(...)` is called, the query execution begins. `withTimeout` wraps the *result* of the chain (the `.limit(1)` promise), not the query construction. If the query itself hangs before the promise is returned, `withTimeout` still correctly aborts the await. **No issue — this is the intended pattern.**

### Security

**5. `staging-clerk-keys.ts` — `_resolveClerkKeysCache` is exported for test teardown.** The export `_resolveClerkKeysCache` is accessible to any code that imports from this module. In production, nothing should call this. The JSDExported for test teardown only. Do not call from production code." is a comment, not a runtime guard. **Low severity** — this is a common pattern and the export name is prefixed with `_` to signal private. No current production callers. Acceptable.

**6. `proxy-routing.ts` — `analyzeHostCache` keyed on attacker-controlled `Host` header.** The JSDoc correctly notes this and caps at 50 entries with FIFO eviction. **This is properly mitigated.** No issue.

**7. `proxy-routing.ts` — `categorizePathCache` keyed on pathname.** Pathnames are attacker-controllable (URL path). Capped at 1000 with FIFO eviction. **Properly mitigated.** No issue.

### Data Integrity

**8. `staging-clerk-keys.ts` — Cache stores `status:'ok'` results only.** The code correctly avoids caching degraded statuses (`staging_missing`, `staging_inherits_prod`, `no_publishable_key`). A cold-start env-miss won't be pinned. **Correct.**

**9. `staging-clerk-keys.ts` — `no_publishable_key` status path now always returns `no_publishable_key` even when `publishableKey` is truthy but `secretKey` is falsy.** The diff shows the final return block:
```ts
if (publishableKey && secretKey) {
  const prodResult = { publishableKey, secretKey, status: 'ok' };
  cacheClerkKeys(hostname, prodResult);
  return prodResult;
}
return { publishableKey, secretKey, status: 'no_publishable_key' };
```
This correctly handles the case where `publishableKey` exists but `secretKey` is missing — returns `no_publishable_key` without caching. **Correct.**

### Error Handling

**10. `public-profile-block.ts` — `withTimeout` fail-open contract.** The `catch {}` block in `checkProfileVisitorBlocked` returns `false` on any error, including `QueryTimeoutError`. This is the documented fail-open behavior. **Correct.**

**11. `public-profile-block.ts` — `profileHasActiveBlocks` and `isVisitorBlockedByFingerprint` both wrap their queries in `withTimeout`.** If `withTimeout` rejects, the error propagates up to `checkProfileVisitorBlocked`'s `catch {}`. **Correct.**

### Cross-File Impact

**12. `_getGeoFromRequest` removed from `proxy.ts`.** Need to verify no other file imports it. The diff shows it was a private function (prefixed with `_`) defined in `proxy.ts`. If any other module imports it, this is a breaking change. **Cannot verify from diff alone — recommend a codebase-wide grep for `_getGeoFromRequest` before merge.**

**13. `resolveClerkKeys` caching changes the return type contract subtly.** Previously, each call created a new `ClerkKeys` object. Now, cached calls return the same object reference. If any caller does identity comparison (`===`) between results from different hostnames, behavior is unchanged (different keys). If a caller mutates the returned object, it would corrupt the cache. **Low severity — no evidence of either pattern in the diff, but worth a grep for mutation of the return value.**

### Summary

|---|---|---|
| Logic errors | 0 confirmed | — |
| Security | 0 confirmed | — |
| Data integrity | 0 confirmed | — |
| Error handling | 0 confirmed | — |
| Cross-file impact | 1 unverified | Medium |

**Action required before merge:**
- Grep the codebase for `_getGeoFromRequest` to confirm no external imports before deleting it from `proxy.ts`.

**Latent risk (non-blocking):**
- `categorizePath` cache returns shared mutable references. Consider `Object.freeze()` on the cached `PathCategory` objects to prevent future mutation bugs.

itstimwhite:

### Logic Errors

**1. `apps/web/lib/audience/public-profile-block.ts:280` — `isVisitorBlockedByFingerprint` timeout wrapper applied to wrong promise shape**

The diff shows the `withTimeout` call wraps `queryPromise` which is the `db.select(...)` chain. However, looking at the pattern from `profileHasActiveBlocks` (lines 246–251), the same structure is used. The concern here is that `isVisitorBlockedByFingerprint` has an `innerJoin` chain — the query builder's `.limit(1)` call at line 279 returns the promise. The `withTimeout` wrapping at line 281 correctly wraps the full chain. **This is fine** — no issue here on closer inspection.

**2. `apps/web/lib/routing/proxy-routing.ts:223` — FIFO eviction race: `size` check and `set` are not atomic**

```ts
if (categorizePathCache.size >= CATEGORIZE_PATH_CACHE_MAX) {
    const firstKey = categorizePathCache.keys().next().value;
    if (firstKey !== undefined) categorizePathCache.delete(firstKey);
}
categorizePathCache.set(pathname, result);
```

This is a classic TOCTOU pattern. If two concurrent requests both see `size === 1000` and both proceed to delete-and-set, the second `set` brings size to 1001 before any eviction happens. Over time under load, the map can drift above the cap. In a Next.js serverless environment, each worker is single-threaded per request, so true concurrency within one worker is unlikely. However, across multiple concurrent requests on the same worker (e.g., during `Promise.all` or streaming), interleaved awaits could cause this.

**Why it matters:** Under sustained load the cache could grow beyond 1000 entries, partially defeating the DoS protection the cap is meant to provide.

**Suggested fix:** Move the eviction *before* the `set` in a single atomic block, or use `Map` size as a post-set check and evict until under cap. The simplest fix:

```ts
if (categorizePathCache.size >= CATEGORIZE_PATH_CACHE_MAX) {
    const firstKey = categorizePathCache.keys().next().value;
    if (firstKey !== undefined) categorizePathCache.delete(firstKey);
}
categorizePathCache.set(pathname, result);
```

This same pattern repeats for `analyzeHostCache` at line 269 and `_resolveClerkKeysCache` at line 50 in `staging-clerk-keys.ts`. All three sites have the identical TOCTOU window.

**Actual severity:** Low in practice (Node.js is single-threaded per event loop, and the `await` points in these functions are not in the cache section). The cache set/delete happens synchronously without intervening `await`, so interleaving is not possible. **Not a real bug** — withdrawing this finding.

**3. `apps/web/lib/auth/staging-clerk-keys.ts:130` — `no_publishable_key` status now returned even when `publishableKey` is present but `secretKey` is missing**

```ts
if (publishableKey && secretKey) {
    const prodResult: ClerkKeys = { publishableKey, secretKey, status: 'ok' };
    cacheClerkKeys(hostname, prodResult);
    return prodResult;
}
return {
    publishableKey,
    secretKey,
    status: 'no_publishable_key',
};
```

Previously the status was `publishableKey && secretKey ? 'ok' : 'no_publishable_key'`. Now when `publishableKey` exists but `secretKey` is `undefined`, the function returns `{ publishableKey: 'pk_live_...', secretKey: undefined, status: 'no_publishable_key' }`. The old code would have returned the same shape and status. **No behavioral change** — the refactor is equivalent. No issue.

### Security

**4. `apps/web/lib/routing/proxy-routing.ts:217` — `categorizePath` cache key is `pathname` which includes usernames**

The cache stores `PathCategory` objects keyed by pathname (e.g., `/tim`, `/some-username`). The comment says "usernames are unbounded so an unguarded Map would grow without limit" and the 1000-entry FIFO cap is the mitigation. This is adequate — FIFO eviction means the oldest entry is always removed, and 1000 entries is a reasonable memory budget. **No issue.**

**5. `apps/web/lib/auth/staging-clerk-keys.ts:50` — `_resolveClerkKeysCache` keyed by `hostname` from `Host` header**

The comment correctly identifies that `hostname` derives from the attacker-controllable `Host` header and the 50-entry FIFO cap is the defense. This is sound. **No issue.**

**6. `apps/web/lib/audience/public-profile-block.ts:327` — `return await` inside `try/catch`**

```ts
return await isVisitorBlockedByFingerprint(normalizedUsername, fingerprint);
```

The `return await` ensures that if `isVisitorBlockedByFingerprint` rejects (e.g., from `withTimeout` throwing `QueryTimeoutError`), the rejection is caught by the surrounding `catch { return false; }`. Without `await`, the rejected promise would escape the `catch` block. This is **correctly fixed** — this is the latent bug mentioned in the PR description. **No issue — this is the fix.**

### Data Integrity

**7. `apps/web/lib/audience/public-profile-block.ts:246` — `withTimeout` wrapping a Drizzle query builder promise**

```ts
const queryPromise = db
  .select(...)
  .from(...)
  .where(...)
  .limit(1);

const [result] = await withTimeout(queryPromise, ...);
```

Drizzle's query builder returns a thenable that only executes on `await`/`then`. By wrapping the promise *before* `await`, the timeout correctly bounds the actual DB execution. If instead `withTimeout` were applied to the already-awaited result, it would be a no-op. **This is correct.**

### Error Handling

**8. `apps/web/lib/audience/public-profile-block.ts:324–329` — `catch {}` swallows all errors including non-timeout errors**

```ts
try {
    ...
    const fingerprint = await createFingerprintEdge(ip, ua);
    return await isVisitorBlockedByFingerprint(normalizedUsername, fingerprint);
} catch {
    return false;
}
```

This is intentional fail-open behavior (audience block should not break navigation). The `withTimeout` wrapper ensures timeout errors are caught here. Non-timeout errors (e.g., programming bugs in the query) are also swallowed. This is a design choice documented in the PR ("fail-open, contract-consistent"). **Acceptable for this use case.**

### Cross-File Impact

**9. `apps/web/lib/auth/staging-clerk-keys.ts:40` — `_resolveClerkKeysCache` exported for test teardown**

The export `_resolveClerkKeysCache` is used in the test file for `.clear()` in `beforeEach`. The Jovie convention is that test-only exports should be clearly marked. The comment says "Exported for test teardown only. Do not call from production code." This is acceptable, though a cleaner pattern would be to add a `beforeEach` hook in the module itself or use a `reset()` function. **Minor — no blocker.**

**10. `apps/web/proxy.ts:637` — `_getGeoFromRequest` removed**

The dead function `_getGeoFromRequest` is deleted. Need to verify no other file imports it. The diff shows it was only defined, not called. If any other file imports it, this would be a breaking change. **Cannot verify from diff alone — recommend `grep -r '_getGeoFromRequest' apps/web/` to confirm no callers exist.**

### Summary

| Category | Finding | Severity |
|----------|---------|----------|
| Logic | None found (TOCTOU window is not reachable — synchronous code) | — |
| Security | None found | — |
| Data integrity | None found | — |
| Error handling | None found (fail-open is intentional) | — |
| Cross-file | Verify `_getGeoFromRequest` has no external callers | Low — blocker if callers exist |

**One actionable item:** Run `grep -r '_getGeoFromRequest' apps/web/` to confirm no external imports before merge. If any file imports this function, the removal at `proxy.ts:637` will cause a build failure.

itstimwhite (commented):

itstimwhite:
⚠️ Merge Conflict Detected — This PR has state 'BEHIND' and cannot be auto-merged.

Branch: 'tim/jov-3231-3232-proxy-hotpath-to-main'
State: BEHIND



To resolve:
1. git fetch origin main
2. git checkout tim/jov-3231-3232-proxy-hotpath-to-main
3. git rebase origin/main
5. git push --force-with-lease origin tim/jov-3231-3232-proxy-hotpath-to-main

PR #11319, opened 2026-06-20
test(ci): cover merge queue guard decisions

itstimwhite:
## Summary
- Add pure merge-queue guard decision helpers for freshness, required status evaluation, autonomous overlap keys, ratchet count comparison, fast-track policy, and telemetry parsing.
- Add deterministic regression coverage for stale-head, conflict, hot-file overlap, ratchet decrease/regression, emergency fast-track, and telemetry metrics.

## Stack
1. This PR: pure guard library and tests.
2. Next: guarded Graphite enrollment wiring.
3. Next: autonomous dispatch serialization.
4. Top: conflict-free ratchets and queue telemetry report.

## Verification
- `pnpm exec vitest --root scripts --config vitest.config.mts run lib/__tests__/merge-queue-guard.test.mjs lib/__tests__/codex-issue-shipper.test.mjs`
- `node --check scripts/lib/merge-queue-guard.mjs`

## Summary by CodeRabbit

* **New Features**
  * Added merge queue system with automated eligibility guards and conflict detection
  * Implemented file overlap analysis to prevent concurrent incompatible merges
  * Added fast-track merge policies for priority changes
  * Integrated queue telemetry tracking and timeline metrics

* **Tests**
  * Added comprehensive test coverage for merge queue guard policies

### DIFF
--- a/scripts/lib/__tests__/merge-queue-guard.test.mjs
+++ b/scripts/lib/__tests__/merge-queue-guard.test.mjs
@@ -1,0 +1,211 @@
+import { describe, expect, it } from 'vitest';
+import {
+  compareRatchetCounts,
+  detectChangedFileOverlap,
+  detectIssueOverlap,
+  fastTrackPolicy,
+  isAutonomousBranch,
+  parseMergeQueueTimeline,
+  preQueueFreshnessDecision,
+  requiredStatusDecision,
+} from '../merge-queue-guard.mjs';
+
+const greenStatuses = [
+  { name: 'PR Ready', conclusion: 'SUCCESS' },
+  { name: 'Migration Guard', conclusion: 'SUCCESS' },
+  { name: 'Fork PR Gate', state: 'SUCCESS' },
+];
+
+describe('merge queue pre-queue freshness', () => {
+  it('waits for CI after a stale head is rebased and pushed', () => {
+    const decision = preQueueFreshnessDecision({
+      behindBy: 4,
+      rebaseAttempted: true,
+      rebaseOk: true,
+      pushedRebasedHead: true,
+      requiredStatuses: greenStatuses,
+    });
+
+    expect(decision.action).toBe('wait_for_ci');
+    expect(decision.reason).toContain('required checks must rerun');
+  });
+
+  it('blocks known conflicts before the merge label can be applied', () => {
+    const decision = preQueueFreshnessDecision({
+      behindBy: 2,
+      rebaseAttempted: true,
+      rebaseOk: false,
+      requiredStatuses: greenStatuses,
+    });
+
+    expect(decision.action).toBe('block_conflict');
+  });
+
+  it('enqueues only when the head is fresh and required statuses are green', () => {
+    expect(requiredStatusDecision(greenStatuses).ok).toBe(true);
+    expect(
+      preQueueFreshnessDecision({
+        behindBy: 0,
+        requiredStatuses: greenStatuses,
+      }).action
+    ).toBe('enqueue');
+
+    const waiting = preQueueFreshnessDecision({
+      behindBy: 0,
+      requiredStatuses: [
+        ...greenStatuses.slice(0, 2),
+        { name: 'Fork PR Gate', state: 'PENDING' },
+      ],
+    });
+    expect(waiting.action).toBe('wait_for_ci');
+    expect(waiting.reason).toContain('Fork PR Gate');
+  });
+});
+
+describe('autonomous dispatch overlap', () => {
+  it('does not classify generic human branches as autonomous blockers', () => {
+    expect(isAutonomousBranch('tim/manual-copy-fix')).toBe(false);
+    expect(isAutonomousBranch('tim/jov-123-generated-fix')).toBe(true);
+  });
+
+  it('blocks two autonomous PRs touching the same hot file', () => {
+    const result = detectChangedFileOverlap(
+      ['apps/web/tests/unit/design-system/arbitrary-values.baseline.json'],
+      [
+        {
+          number: 11263,
+          title: 'converge artists page arbitraries',
+          headRefName: 'tim/jov-ds-drift-artists-page',
+          changedFiles: [
+            'apps/web/tests/unit/design-system/arbitrary-values.baseline.json',
+          ],
+        },
+      ]
+    );
+
+    expect(result.blocked).toBe(true);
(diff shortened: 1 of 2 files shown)


itstimwhite:
## Review: PR #11319 — merge-queue-guard library + tests

---

**1. Logic errors**

**`scripts/lib/merge-queue-guard.mjs:370` — `requiredStatusDecision` uses loose `endsWith` matching that can match unintended status names**

```js
const suffixed = [...byName.entries()].find(([name]) =>
  name.endsWith(` / ${required}`)
)?.[1];
```

This picks the *first* entry whose name ends with ` / ${required}`, but `find` on a Map iterates insertion order. If two status checks share the same suffix (e.g., `PR Ready` and `PR Ready (re-run)`), the wrong one could be selected. More critically, `endsWith` is overly broad — a status named `Not PR Ready` would also match. This should use an exact match or a stricter prefix pattern like `startsWith`.

**Fix:** Prefer exact match first, then check for `startsWith(`${required} /`)` or similar structured patterns, and ensure only one match is considered.

---

**`scripts/lib/merge-queue-guard.mjs:370-372` — `suffixed` lookup silently falls through to `byName.get(required)` even when a suffixed match exists**

```js
const suffixed = [...byName.entries()].find(([name]) =>
  name.endsWith(` / ${required}`)
)?.[1];
const status =
  byName.get(required) ?? byName.get(`CI / ${required}`) ?? suffixed;
```

The intent seems to be: try exact name, then `CI / Name`, then suffixed. But if `byName.get(required)` returns a *different* status that happens to have the exact same name (unlikely but possible in malformed input), the suffixed result is never used. The logic is correct for the happy path, but the variable name `suffixed` is misleading — it's actually a "fallback suffix match." Not a bug per se, but fragile.

---

**`scripts/lib/merge-queue-guard.mjs:409` — `preQueueFreshnessDecision` returns inconsistent `reason` when `behindBy > 0` and all statuses are green**

```js
return {
  action: 'enqueue',
  reason:
    behindBy === 0
      ? 'head is fresh and required statuses are green'
      : 'head was already current after freshness validation',
  statuses,
};
```

When `behindBy > 0` (branch is behind but no rebase was attempted), the action is `enqueue` with reason "head was already current" — which is contradictory. A branch with `behindBy > 0` is *not* current. This looks like a logic gap: the function doesn't handle the case where `behindBy > 0` but no rebase was attempted. Should this be `wait_for_ci` or `block` instead?

**This is the most significant logic issue in the PR.**

---

**`scripts/lib/merge-queue-guard.mjs:409-415` — Missing handling for `behindBy > 0` without rebase**

If `behindBy` is 4, `rebaseAttempted` is false, `rebaseOk` is true (default), `pushedRebasedHead` is false, and statuses are green, the function falls through to the `enqueue` branch. A branch 4 commits behind main should not be enqueued without rebasing. There should be a check like:

```js
if (behindBy > 0 && !pushedRebasedHead) {
  return { action: 'wait_for_ci', reason: `branch is ${behindBy} commits behind main` };
}
```

---

**`scripts/lib/merge-queue-guard.mjs:449` — `compareRatchetCounts` treats missing keys as 0**

```js
const current = Number(currentCounts[key] ?? 0);
const base = Number(baseCounts[key] ?? 0);
```

If a key exists in `baseCounts` but not `currentCounts`, it's treated as 0 and counted as an "improvement." This may be intentional (the counter was removed), but it could mask data loss. If a ratchet counter disappears entirely, that's arguably a regression, not an improvement. Document this behavior or add a `missing` category.

---

**`scripts/lib/merge-queue-guard.mjs:473` — `parseTelemetryMarker` uses a non-greedy match that can fail on nested JSON**

```js
const match = body.match(/<!--\s*merge-queue-telemetry\s+({[\s\S]*?})\s*-->/);
```

The `[\s\S]*?` is non-greedy and will stop at the first `}`, which is correct for flat JSON but will break if the JSON contains nested objects (e.g., `{"event":"enqueue","meta":{"foo":"bar"}}`). The match would capture only `{"event":"enqueue","meta":{"foo":"bar"` — truncated.

**Fix:** Use a greedy match or a proper JSON extraction approach:
```js
const match = body.match(/<!--\s*merge-queue-telemetry\s+({[\s\S]*})\s*-->/);
```

---

**`scripts/lib/merge-queue-guard.mjs:500` — `conflictEvictions` and `ciComments` use comment text matching, but a single comment could match both patterns**

If a comment contains both "Merge Conflict Detected" and "CI failed", it would be counted in both `conflictComments` and `ciComments`, inflating both `conflictEvictions` and `ciEvictions`. The test case avoids this, but real-world data could trigger double-counting.

---

**2. Security**

**None found.** This is a pure utility library with no I/O, no auth, no external calls, no user-facing rendering. The regex patterns are well-scoped and don't exhibit ReDoS vulnerabilities.

---

**3. Data integrity**

**`scripts/lib/merge-queue-guard.mjs:473` — `parseTelemetryMarker` silently returns `null` on malformed JSON**

```js
try {
  return JSON.parse(match[1]);
} catch {
  return null;
}
```

This is intentional and acceptable for telemetry parsing — malformed markers should not crash the parser. No issue here, but worth noting that no warning is logged. Acceptable for a guard library.

---

**4. Error handling**

**`scripts/lib/merge-queue-guard.mjs:396` — `requiredStatusDecision` doesn't handle `statuses` being `null`/`undefined` gracefully if called without the optional chaining default**

The function signature uses `statuses` without a default, but callers in `preQueueFreshnessDecision` pass `requiredStatuses = []`. If called directly with `null`, `for (const status of statuses ?? [])` handles it. This is fine.

---

**5. Cross-file impact**

**None found.** This is a new, self-contained library with no existing callers. The exports are well-defined and the test file imports correctly.

---

### Summary of issues

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **High** | `:409-415` | `behindBy > 0` without rebase falls through to `enqueue` — branch behind main should not be enqueued |
| 2 | **Medium** | `:473` | Non-greedy JSON regex breaks on nested objects in telemetry markers |
| 3 | **Medium** | `:500` | Single comment matching both conflict and CI patterns causes double-counting |
| 4 | **Low** | `:370` | `endsWith` status matching is overly broad |
| 5 | **Low** | `:449` | Missing ratchet keys treated as 0/improvement — may mask data loss |

**Issue #1 is a blocking logic bug.** The function will incorrectly enqueue a stale branch.

itstimwhite (commented):

PR #11341, opened 2026-06-21
chore: candidate follow-up: Per-card impression + CTR analytics for catalog car (#11118)

### LINKED ISSUE #11118: Candidate follow-up: Per-card impression + CTR analytics for catalog carousel (JOV-11085)
## Source

* Current issue: JOV-11085 ([https://github.com/JovieInc/Jovie/issues/11085](<https://github.com/JovieInc/Jovie/issues/11085>))
* Source branch: codex/t_6a516173/20260619083218

## Follow-up

The catalog carousel (`ReleaseCatalogCarousel`) was shipped without per-card impression tracking and CTR analytics. The JOV-11085 acceptance criteria required:

* Per-card impression event when a card becomes visible
* Listen CTA click event on each catalog card (with `release_id`, card `index`, `artist_handle`)

## Why it matters

The analytics feed the album-art A/B optimizer. Without them, we can't measure whether surfacing more catalog increases streams/conversions per visit, and we can't validate the carousel didn't hurt primary-card CTR.

## Classification

Candidate

Should we instrument this with existing profile analytics helpers (`track()`) or wait until the A/B optimizer infra defines the event schema? Pickup agent must first judge whether to implement, close, or split this.

itstimwhite:
Autonomously implemented by the Hermes shipping loop (codex). Closes #11118.

Card: t_1ed51a49
Review + merge are gated by the Hermes auto-merge rails (strict CI green + not taste-touching + cross-model 2nd-opinion + spec-check + no hard-gate label).

## Operational validation (for /land-and-deploy canary)
**Monitor after deploy:**
- client console errors + render of affected pages
**Rollback trigger:** any new 5xx/console-error spike or p95 regression vs. pre-merge baseline on the surfaces above → revert this PR.

## Summary by CodeRabbit

* **New Features**
  * Added a release catalog carousel to artist profile home pages, displaying a curated collection of releases.
  * Implemented analytics tracking for carousel interactions (impressions and clicks).

* **Improvements**
  * Enhanced carousel card interactivity with click handlers.
  * Improved dark mode styling for profile sections.

### DIFF
--- a/apps/web/components/features/home/HomeProfileShowcase.tsx
+++ b/apps/web/components/features/home/HomeProfileShowcase.tsx
@@ -18,6 +18,7 @@ import { HomePhoneFrame } from './HomePhoneFrame';
 import {
   HOMEPAGE_PROFILE_PREVIEW_ARTIST,
   HOMEPAGE_PROFILE_PREVIEW_CONTACTS,
+  HOMEPAGE_PROFILE_PREVIEW_DRAWER_RELEASES,
   HOMEPAGE_PROFILE_PREVIEW_MOCK_HOME_ARTIST,
   HOMEPAGE_PROFILE_PREVIEW_MOCK_HOME_RELEASE,
   HOMEPAGE_PROFILE_PREVIEW_MOCK_HOME_SOCIAL_LINKS,
@@ -173,6 +174,14 @@ function getLatestRelease(stateId: ProfileShowcaseStateId) {
   }
 }
 
+function getPreviewReleases(stateId: ProfileShowcaseStateId) {
+  if (stateId === 'mock-home' || getLatestRelease(stateId) !== null) {
+    return [...HOMEPAGE_PROFILE_PREVIEW_DRAWER_RELEASES];
+  }
+
+  return [];
+}
+
 function getTourDates(stateId: ProfileShowcaseStateId) {
   switch (stateId) {
     case 'mock-home':
@@ -269,6 +278,7 @@ function ShowcaseSurface({
         pressPhotos={[]}
         allowPhotoDownloads={false}
         tourDates={getTourDates(state.id)}
+        releases={getPreviewReleases(state.id)}
         viewerLocation={getViewerLocation(state.id)}
         resolveNearbyTour={shouldResolveNearbyTour(state.id)}
         showSubscriptionConfirmedBanner={state.showSubscriptionConfirmedBanner}
--- a/apps/web/components/features/profile/ProfileHomeRail.tsx
+++ b/apps/web/components/features/profile/ProfileHomeRail.tsx
@@ -4,7 +4,6 @@ import { Bell } from 'lucide-react';
 import { useMemo } from 'react';
 import {
   type EntityCardModel,
-  EntityCarousel,
   merchToEntityCard,
   releaseToEntityCard,
   showToEntityCard,
@@ -26,6 +25,8 @@ import { getProfileReleaseVisibility } from '@/lib/profile/release-visibility';
 import type { TourDateViewModel } from '@/lib/tour-dates/types';
 import { cn } from '@/lib/utils';
 import type { Artist } from '@/types/db';
+import { ReleaseCatalogCarousel } from './ReleaseCatalogCarousel';
+import type { PublicRelease } from './releases/types';
 
 interface ProfileHomeRailProps {
   readonly artist: Artist;
@@ -44,6 +45,7 @@ interface ProfileHomeRailProps {
   readonly viewerLocation?: UserLocation | null;
   readonly resolveNearbyTour?: boolean;
   readonly merchCards?: readonly PublicMerchCard[];
+  readonly releases?: readonly PublicRelease[];
 }
 
 function getUpcomingTourDates(
@@ -128,7 +130,7 @@ function HomeAlertsCard({
         aria-hidden='true'
         data-testid='profile-home-alerts-switch'
       >
-        <span className='block h-6 w-6 rounded-full bg-white shadow-[0_3px_10px_rgba(0,0,0,0.32)]' />
+        <span className='block h-6 w-6 rounded-full bg-white dark:bg-white shadow-[0_3px_10px_rgba(0,0,0,0.32)]' />
       </span>
     </>
   );
@@ -167,6 +169,7 @@ export function ProfileHomeRail({
   viewerLocation,
   resolveNearbyTour = true,
   merchCards = [],
+  releases = [],
 }: Readonly<ProfileHomeRailProps>) {
   // Re-evaluate visibility at the release boundary so the rail's "Drops in"
   // chrome transitions to "Out Now" when the release drops, even if the
@@ -202,10 +205,19 @@ export function ProfileHomeRail({
     const hasFeaturedRelease = Boolean(
       releaseVisibility?.show && latestRelease
     );
+    const featuredReleaseSlug =
+      hasFeaturedRelease && latestRelease ? latestRelease.slug : null;
+    const featuredReleaseId =
+      featuredReleaseSlug === null
+        ? null
+        : (releases.find(release => release.slug === featuredReleaseSlug)?.id ??
+          null);
+
     if (hasFeaturedRelease && latestRelease) {
       items.push(
         releaseToEntityCard(
           {
+            id: featuredReleaseId ?? undefined,
             title: latestRelease.title,
             slug: latestRelease.slug,
             artworkUrl: latestRelease.artworkUrl,
@@ -233,6 +245,14 @@ export function ProfileHomeRail({
           external: true,
         },
       });
+    }
+
+    // Back catalog: include non-featured releases in the same carousel.
+    for (const release of releases) {
+      if (release.slug === '' || release.slug === featuredReleaseSlug) {
+        continue;
+      }
+      items.push(releaseToEntityCard(release, { handle: artist.handle, now }));
     }
 
     // Shoppable merch (revenue) next.
@@ -265,6 +285,7 @@ export function ProfileHomeRail({
     merchCards,
     nearbyTourDateId,
     now,
+    releases,
     releaseVisibility?.show,
     upcomingTourDates,
   ]);
@@ -292,10 +313,11 @@ export function ProfileHomeRail({
       data-testid='profile-home-rail'
     >
       {alertsCard}
-      <EntityCarousel
+      <ReleaseCatalogCarousel
         items={carouselItems}
-        surface='pearl'
-        dataTestId='profile-home-carousel'
+        artistHandle={artist.handle}
+        artistId={artist.id}
+        analyticsEnabled={renderMode !== 'preview'}
       />
     </div>
   );
--- a/apps/web/components/features/profile/ReleaseCatalogCarousel.tsx
+++ b/apps/web/components/features/profile/ReleaseCatalogCarousel.tsx
@@ -1,0 +1,83 @@
+'use client';
+
+import { useCallback } from 'react';
+import { EntityCarousel } from '@/components/organisms/entity-card';
+import type { EntityCardModel } from '@/components/organisms/entity-card/types';
+import { track } from '@/lib/analytics';
+
+interface ReleaseCatalogCarouselProps {
+  readonly items: readonly EntityCardModel[];
+  readonly artistHandle: string;
+  readonly artistId: string;
+  readonly analyticsEnabled?: boolean;
+  readonly dataTestId?: string;
+}
+
+function isCatalogMusicCard(model: EntityCardModel): boolean {
+  return model.kind === 'music' || model.kind === 'video';
+}
+
+function resolveReleaseId(model: EntityCardModel): string {
+  return model.releaseId ?? model.id;
+}
+
+export function ReleaseCatalogCarousel({
+  items,
+  artistHandle,
+  artistId,
+  analyticsEnabled = true,
+  dataTestId = 'profile-home-carousel',
+}: Readonly<ReleaseCatalogCarouselProps>) {
+  const handleCardImpression = useCallback(
+    (index: number, model: EntityCardModel) => {
+      if (!analyticsEnabled || !isCatalogMusicCard(model)) {
+        return;
+      }
+
+      track('catalog_carousel_card_impression', {
+        release_id: resolveReleaseId(model),
+        index,
+        artist_handle: artistHandle,
+        artist_id: artistId,
+        profile_id: artistId,
+        profile_slug: artistHandle,
+        release_slug: model.id,
+        is_featured: index === 0,
+        current_route_tab: 'home',
+      });
+    },
+    [analyticsEnabled, artistHandle, artistId]
+  );
+
+  const handleCardClick = useCallback(
+    (index: number, model: EntityCardModel) => {
+      if (!analyticsEnabled || !isCatalogMusicCard(model)) {
+        return;
+      }
+
+      track('catalog_carousel_listen_click', {
+        release_id: resolveReleaseId(model),
+        index,
+        artist_handle: artistHandle,
+        artist_id: artistId,
+        profile_id: artistId,
+        profile_slug: artistHandle,
+        release_slug: model.id,
+        is_featured: index === 0,
+        current_route_tab: 'home',
+        cta_location: 'catalog_carousel',
+      });
+    },
+    [analyticsEnabled, artistHandle, artistId]
+  );
+
+  return (
+    <EntityCarousel
+      items={items}
+      surface='pearl'
+      dataTestId={dataTestId}
+      onCardImpression={handleCardImpression}
+      onCardClick={handleCardClick}
+    />
+  );
+}
--- a/apps/web/components/features/profile/ReleaseCatalogCarousel.tsx
+++ b/apps/web/components/features/profile/ReleaseCatalogCarousel.tsx
@@ -13,12 +13,14 @@ interface ReleaseCatalogCarouselProps {
   readonly dataTestId?: string;
 }
 
-function isCatalogMusicCard(model: EntityCardModel): boolean {
-  return model.kind === 'music' || model.kind === 'video';
-}
-
-function resolveReleaseId(model: EntityCardModel): string {
-  return model.releaseId ?? model.id;
+function isCatalogReleaseCard(
+  model: EntityCardModel
+): model is EntityCardModel & { readonly releaseId: string } {
+  return (
+    (model.kind === 'music' || model.kind === 'video') &&
+    typeof model.releaseId === 'string' &&
+    model.releaseId.length > 0
+  );
 }
 
 export function ReleaseCatalogCarousel({
@@ -30,12 +32,12 @@ export function ReleaseCatalogCarousel({
 }: Readonly<ReleaseCatalogCarouselProps>) {
   const handleCardImpression = useCallback(
     (index: number, model: EntityCardModel) => {
-      if (!analyticsEnabled || !isCatalogMusicCard(model)) {
+      if (!analyticsEnabled || !isCatalogReleaseCard(model)) {
         return;
       }
 
       track('catalog_carousel_card_impression', {
-        release_id: resolveReleaseId(model),
+        release_id: model.releaseId,
         index,
         artist_handle: artistHandle,
         artist_id: artistId,
@@ -51,12 +53,12 @@ export function ReleaseCatalogCarousel({
 
   const handleCardClick = useCallback(
     (index: number, model: EntityCardModel) => {
-      if (!analyticsEnabled || !isCatalogMusicCard(model)) {
+      if (!analyticsEnabled || !isCatalogReleaseCard(model)) {
         return;
       }
 
       track('catalog_carousel_listen_click', {
-        release_id: resolveReleaseId(model),
+        release_id: model.releaseId,
         index,
         artist_handle: artistHandle,
         artist_id: artistId,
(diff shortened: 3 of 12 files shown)


itstimwhite:

### Logic Errors

**1. `EntityCarousel.tsx:33` — `items.entries()` called on `items` but `items` is `readonly` array; `for...of` works fine, but the `key` for deduplication uses `model.id` which is the slug, not a stable DB id.**

The impression key is `${model.kind}-${model.id}-${index}`. Since `model.id` is the slug (e.g., `"catalog-cut"`), if two different releases share the same slug (unlikely but possible in preview/mock data), impressions would be deduplicated incorrectly. More critically, if the same release appears at two different indices (shouldn't happen, but the dedup key includes `index` so this is actually safe). **This is fine on closer inspection — the index disambiguates.** No issue.

**2. `ProfileHomeRail.tsx:211-214` — `featuredReleaseId` is resolved but never used.**

```ts
const featuredReleaseId =
  featuredReleaseSlug === null
    ? null
    : (releases.find(release => release.slug === featuredReleaseSlug)?.id ??
      null);
```

This variable is computed but never referenced. It was likely intended to be passed as the `id` field on the featured release's `ReleaseEntityInput`, but the `releaseToEntityCard` call on line 218 doesn't use it. The featured release gets `id: featuredReleaseId ?? undefined` (line 220), but `featuredReleaseId` is always `null` here because `latestRelease` comes from `releaseVisibility`, not from the `releases` array — so `releases.find(...)` won't match.

Wait — re-reading: `latestRelease` is a separate prop (`ProfilePrimaryActionCardRelease`), not from the `releases` array. So `featuredReleaseId` will always be `null`, and the featured card's `releaseId` will be `undefined`. This means the featured release card will never trigger analytics tracking (since `isCatalogReleaseCard` requires `releaseId` to be a non-empty string).

**Severity: Medium.** The featured release — the most important card in the carousel — will be invisible to impression/click analytics.

**Suggested fix:** Either pass the `id` from `latestRelease` directly (it likely has an `id` field already), or resolve it from the `releases` array by matching on slug and passing the found `id` into the `releaseToEntityCard` call.

**3. `ProfileHomeRail.tsx:250-253` — Back-catalog loop skips releases with empty slug but the condition `release.slug === ''` is fragile.**

```ts
for (const release of releases) {
  if (release.slug === '' || release.slug === featuredReleaseSlug) {
    continue;
  }
  items.push(releaseToEntityCard(release, { handle: artist.handle, now }));
}
```

If a release has `slug === null` (possible from the DB), this check won't catch it and will pass `null` to `releaseToEntityCard`, which expects `string`. The `releaseToEntityCard` function uses `release.slug` as the `id` field in the returned model. A `null` slug would produce `id: null`, which is a type violation.

**Severity: Low-Medium.** Depends on whether `PublicRelease.slug` can ever be null. If the type says `string`, this is fine. If the type allows `null`, this is a runtime bug.

**4. `EntityCarousel.tsx:38-40` — IntersectionObserver fallback path fires all impressions synchronously on mount.**

When `IntersectionObserver` is undefined (e.g., SSR or old browser), the fallback immediately calls `onCardImpression` for every card. This means all cards are "impressed" at once, even those not visible. This is a known trade-off but worth noting for analytics accuracy.

**Severity: Low.** Acceptable fallback behavior, but worth documenting.

**5. `EntityCarousel.tsx:55-56` — `items` array is iterated with `for...of` but `itemRefs.current` is populated via callback refs. The refs may not be populated when the effect runs on first render.**

The `useEffect` depends on `[items, onCardImpression]`. On the initial render, `itemRefs.current` is an empty array. The `for (const node of itemRefs.current)` loop at line 55 will iterate over nothing because the ref callbacks haven't fired yet. The effect won't re-run after refs are populated because ref changes don't trigger re-renders.

**Severity: Medium.** Impression tracking may not work on the initial render in environments where `IntersectionObserver` exists. The observer will only catch cards that mount *after* the effect runs. In practice, React's synchronous ref callbacks fire before the browser paints, and the `useEffect` fires after paint, so the refs *should* be populated. But this is a subtle ordering dependency.

**6. `ReleaseCatalogCarousel.tsx:28` — `isCatalogReleaseCard` checks `model.kind === 'music' || model.kind === 'video'`, but `releaseToEntityCard` can also return `kind: 'video'`. The `releaseId` field is only set for releases, not for other video types (e.g., a video card from a different adapter). This is fine since the type guard checks both kind and releaseId presence.**

No issue here.

### Security

**1. `ReleaseCatalogCarousel.tsx:35-36` — `model.releaseId` is passed directly to analytics without sanitization.**

The `release_id` field in the analytics event is the raw DB id. If these ids are predictable (e.g., UUIDs), this is fine. If they're sensitive, this could leak internal identifiers. This is a low concern for most analytics platforms but worth noting.

**Severity: Low.** Standard practice for analytics.

**2. No XSS, CSRF, SQL injection, or auth bypass vectors found.** The PR is purely UI/analytics, no DB queries, no user input handling, no auth changes.

### Data Integrity

**1. `adapters.ts:82` — `ReleaseEntityInput.id` is optional (`readonly id?: string`), but `releaseToEntityCard` always sets `releaseId: release.id`, which could be `undefined`.**

```ts
return {
  id: release.slug,
  releaseId: release.id,  // could be undefined
  ...
};
```

The `EntityCardModel.releaseId` is also optional (`readonly releaseId?: string`), so this is type-safe. But the `isCatalogReleaseCard` type guard checks `typeof model.releaseId === 'string' && model.releaseId.length > 0`, so cards with `undefined` releaseId will correctly be excluded from analytics. **No issue.**

### Error Handling

**1. `EntityCarousel.tsx:44-46` — `entry.target.getAttribute('data-carousel-index')` returns `null` if the attribute is missing, and `Number(null)` is `NaN`.**

```ts
const index = Number(entry.target.getAttribute('data-carousel-index'));
const model = items[index];  // items[NaN] is undefined
```

If `model` is undefined, the code correctly continues. **No issue — handled gracefully.**

**2. `EntityCarousel.tsx:55` — `for (const node of itemRefs.current)` includes `null` values in the array.**

The ref callback pattern `ref={node => { itemRefs.current[index] = node; }}` can leave `null` entries if items are removed. The `if (node)` check on line 57 handles this. **No issue.**

### Cross-File Impact

**1. `EntityCarousel` is now a client component** (added `'use client'` directive). Any server component that imports `EntityCarousel` will now need to be a client component or use a client boundary. Check that all callers are already client components or that this doesn't force server components to become client components.

**Severity: Medium.** This could cause unexpected rendering behavior if a server component imports `EntityCarousel` directly.

**2. `EntityCard` now accepts an `onClick` prop** which is passed through to both `<Link>` and `<a>` elements. If `onClick` is provided alongside `href`, the click handler will fire *in addition to* navigation. This is intentional for analytics, but could cause double-tracking if callers also have their own click handlers.

**Severity: Low.** The `onClick` is only used in the new `ReleaseCatalogCarousel`, so no existing callers are affected.

---

### Summary of Actionable Issues

| # | File:Line | Severity | Issue |
|---|-----------|----------|-------|
| 1 | `ProfileHomeRail.tsx:211-214` | **Medium** | `featuredReleaseId` resolves to `null` because `latestRelease` is not from the `releases` array; featured card's `releaseId` is always `undefined`, so it's invisible to analytics. |
| 2 | `EntityCarousel.tsx:1` | **Medium** | Adding `'use client'` to `EntityCarousel` may force parent server components to become client components. Verify all callers. |
| 3 | `ProfileHomeRail.tsx:250` | **Low** | `release.slug === ''` check won't catch `null` slugs if `PublicRelease.slug` allows null. |

itstimwhite (commented):

itstimwhite:

### Logic Errors

**1. `EntityCarousel.tsx:33` — `items.entries()` called on `items` array but `items` is the prop, not guaranteed stable across renders**

The `useEffect` depends on `[items, onCardImpression]`. If the parent re-creates the `items` array on every render (which `ProfileHomeRail` does — it builds `carouselItems` inside the component body without `useMemo`), the effect re-runs every render. The `trackedImpressionKeys` ref prevents duplicate tracking, but the observer is torn down and re-created on every render, which is wasteful and could cause missed impressions during the teardown/rebuild window.

**Why it matters:** Performance churn and potential missed impression events during rapid re-renders.

**Suggested fix:** Wrap `carouselItems` in `useMemo` in `ProfileHomeRail` (it already has a `useMemo` block — add the items construction to it), or use a stable identity check in the effect.

---

**2. `ProfileHomeRail.tsx:208-214` — `featuredReleaseId` lookup uses `slug` but `releases` array items have `slug` as the `id` field in `releaseToEntityCard`**

At line 208, `featuredReleaseSlug` is set to `latestRelease.slug`. Then at line 211, it searches `releases.find(release => release.slug === featuredReleaseSlug)`. This is correct — `PublicRelease` has a `slug` field. However, the `releaseToEntityCard` adapter at `adapters.ts:120` sets `id: release.slug`, so the `EntityCardModel.id` is the slug. The `releaseId` is the database id. This means the `featuredReleaseId` variable correctly resolves to the database id. **No bug here** — the logic is consistent.

---

**3. `ProfileHomeRail.tsx:250-255` — Back-catalog loop skips releases with empty slug but does not skip non-release cards**

```ts
for (const release of releases) {
  if (release.slug === '' || release.slug === featuredReleaseSlug) {
    continue;
  }
  items.push(releaseToEntityCard(release, { handle: artist.handle, now }));
}
```

The `release.slug === ''` guard is defensive, but `slug` is required on `PublicRelease` (not optional). If a release somehow has an empty slug, `releaseToEntityCard` would produce a card with `id: ''`, which would collide with the `key` `${model.kind}-${model.id}` in `EntityCarousel` (line 108). This is a latent key-collision risk.

**Why it matters:** If two cards share the same `id` key, React will warn and may drop one from the DOM.

**Suggested fix:** Instead of silently skipping, filter with a more explicit guard or ensure the slug is always non-empty at the type level. The current guard is fine as a safety net, but worth noting.

---

**4. `ReleaseCatalogCarousel.tsx:28` — `isCatalogReleaseCard` type guard asserts `model.releaseId` is `string`, but `releaseId` is optional (`readonly releaseId?: string`)**

The type predicate `model is EntityCardModel & { readonly releaseId: string }` narrows to a non-optional `string`. The check `typeof model.releaseId === 'string' && model.releaseId.length > 0` is correct at runtime. The TypeScript narrowing is valid because the `&` intersection with `readonly releaseId: string` is compatible after the runtime check. **No bug here.**

---

**5. `EntityCarousel.tsx:33-41` — Fallback path when `IntersectionObserver` is undefined fires all impressions synchronously during `useEffect`**

When `IntersectionObserver` is not available (e.g., SSR, old browser), all cards are tracked as impressed immediately. This is intentional for SSR, but the `useEffect` only runs client-side, so SSR is fine. **No bug.**

---

**6. `EntityCarousel.tsx:55` — `threshold: 0.5` means a card must be 50% visible to trigger impression**

This is a design choice, not a bug. But worth noting: in a horizontal carousel with 180px-wide cards, a card may need to be scrolled significantly before triggering. If the intent is to track "seen" cards, a lower threshold (e.g., `0.1`) or `rootMargin` would be more appropriate.

**Why it matters:** Cards at the edge of the viewport may not register impressions until scrolled further.

**Suggested fix:** Consider `threshold: 0.1` or adding `rootMargin: '0px 0px 0px 0px'` with a lower threshold.

---

### Security

**None found.** No hardcoded secrets, no SQL injection vectors, no XSS sinks. The `track()` calls pass structured data, not user-controlled HTML. The `artistHandle` and `artistId` values come from server-side data.

---

### Data Integrity

**7. `ProfileHomeRail.tsx:208-214` — `featuredReleaseId` may be `null` even when `hasFeaturedRelease` is `true`**

If `latestRelease.slug` matches a slug in the `releases` array, `featuredReleaseId` resolves to the database id. But if the `releases` array does not contain the featured release (e.g., the featured release is newer than the catalog query), `featuredReleaseId` is `null`. The featured card is still pushed at line 217 with `id: featuredReleaseId ?? undefined`, so `releaseId` on the `EntityCardModel` will be `undefined`. This means the featured card won't be tracked by `isCatalogReleaseCard` (which requires `releaseId` to be a non-empty string).

**Why it matters:** The featured release card won't get impression/click tracking, which is likely unintended since it's the most prominent card.

**Suggested fix:** Consider always including the featured release in the `releases` array, or pass the featured release's database id separately so it's always available for tracking.

---

### Error Handling

**None found.** The component gracefully handles empty `releases`, missing `IntersectionObserver`, and `analyticsEnabled=false`. No async operations that could throw.

---

### Cross-File Impact

**8. `EntityCarousel.tsx` — Adding `onClick` prop to `EntityCard` changes the card's click behavior globally**

The `onClick` prop is passed through to both `<Link>` and `<a>` elements in `CardShell`. When `onClick` is provided, it fires alongside the navigation. This is fine for analytics, but if any existing consumer of `EntityCarousel` (outside `ReleaseCatalogCarousel`) also passes `onClick`, the behavior changes. Currently only `ReleaseCatalogCarousel` passes it, so no existing consumers are affected. **No current impact, but worth monitoring.**

---

### Summary

| 1 | Medium | `EntityCarousel.tsx:33` | Effect re-runs on every parent render due to unstable `items` ref — observer churn + potential missed impressions |
| 3 | Low | `ProfileHomeRail.tsx:250` | Empty-slug guard is defensive but if triggered, produces duplicate React keys |
| 6 | Low | `EntityCarousel.tsx:55` | `threshold: 0.5` may miss impressions for cards at viewport edge |
| 7 | Medium | `ProfileHomeRail.tsx:208-214` | Featured release card may lack `releaseId` for analytics when not in `releases` array |

itstimwhite (commented):

itstimwhite:

### Logic Errors

**1. `EntityCarousel.tsx` — `itemRefs.current` never cleaned up on unmount or items change**
- **File:** `apps/web/components/organisms/entity-card/EntityCarousel.tsx`
- **Lines:** The `useEffect` at the top of the component body (the one with `IntersectionObserver`) returns `() => observer.disconnect()`, which is correct. However, `itemRefs.current` is never reset when `items` changes. If items shrink (e.g., from 5 to 2), stale refs to unmounted DOM nodes remain in the array. The `IntersectionObserver` is disconnected and recreated, so this is mostly benign — but the `ref` callback `itemRefs.current[index] = node` will write to indices that may no longer correspond to valid items if the list changes. More critically, if items are removed, the `data-carousel-index` attribute on remaining nodes still matches the old index, so `onCardImpression` could fire with a stale/wrong index after re-render.
- **Why it matters:** If the `items` prop changes at runtime (e.g., filtering), the `data-carousel-index` on DOM nodes won't match the new `items` array positions, causing analytics to report wrong indices or wrong models.
- **Suggested fix:** Reset `itemRefs.current = []` at the top of the render (before the `.map`), and add a `key` prop to the `<li>` that includes the index so React re-mounts list items when the list changes. Or, use a `Map<Element, number>` instead of an array indexed by position.

**2. `ProfileHomeRail.tsx` — `featuredReleaseId` lookup is fragile**
- **File:** `apps/web/components/features/profile/ProfileHomeRail.tsx`
- **Lines:** ~208–213
  ```ts
  const featuredReleaseSlug = hasFeaturedRelease && latestRelease ? latestRelease.slug : null;
  const featuredReleaseId = featuredReleaseSlug === null
    ? null
    : (releases.find(release => release.slug === featuredReleaseSlug)?.id ?? null);
  ```
- **Why it matters:** This looks up the featured release's DB `id` from the `releases` array by matching on `slug`. But `latestRelease` comes from a different data source (the `latestRelease` prop) than `releases` (the new prop). If the `releases` array doesn't contain the featured release (which is possible — the featured release is the *latest* release, and `releases` is the *back catalog*), then `featuredReleaseId` will be `null`, and the `id: featuredReleaseId ?? undefined` passed to `releaseToEntityCard` will be `undefined`, meaning the featured release card won't have a `releaseId` and won't get analytics tracking. This is a silent data integrity issue — the featured release won't be tracked.
- **Suggested fix:** Either include the featured release in the `releases` array passed from the parent, or pass the featured release's DB id separately so the lookup isn't needed.

**3. `ReleaseCatalogCarousel.tsx` — `isCatalogReleaseCard` type guard is incorrect**
- **File:** `apps/web/components/features/profile/ReleaseCatalogCarousel.tsx`
- **Lines:** ~22–27
  ```ts
  function isCatalogReleaseCard(
    model: EntityCardModel
  ): model is EntityCardModel & { readonly releaseId: string } {
    return (
      (model.kind === 'music' || model.kind === 'video') &&
      typeof model.releaseId === 'string' &&
      model.releaseId.length > 0
    );
  }
  ```
- **Why it matters:** The type guard narrows to `EntityCardModel & { readonly releaseId: string }`, but `releaseId` is optional on `EntityCardModel` (`readonly releaseId?: string`). The runtime check is correct, but the TypeScript narrowing is fine — this is actually okay. No bug here, but worth noting that `video` kind cards with a `releaseId` will also be tracked, which may or may not be intentional.

**4. `EntityCarousel.tsx` — `data-carousel-index` attribute may not match after list mutation**
- **File:** `apps/web/components/organisms/entity-card/EntityCarousel.tsx`
- **Lines:** The `ref` callback sets `itemRefs.current[index] = node` and `data-carousel-index={index}` is set on the `<li>`. But the `ref` callback is called *after* the `data-carousel-index` attribute is set. If React reorders DOM nodes (which it can do with keyed lists), the `data-carousel-index` could be stale by the time the observer fires.
- **Why it matters:** Low risk in practice because the list is rendered in order, but it's a latent correctness issue.

### Security

**None found.** No hardcoded secrets, no user-input-driven XSS vectors, no auth bypass, no SQL injection. The `track()` calls pass structured data, not raw user input.

### Data Integrity

**5. `adapters.ts` — `releaseId` is set to `undefined` when `release.id` is not provided**
- **File:** `apps/web/components/organisms/entity-card/adapters.ts`
- **Line:** ~120
  ```ts
  releaseId: release.id,
  ```
- **Why it matters:** When `release.id` is `undefined`, the resulting model will have `releaseId: undefined`. The `isCatalogReleaseCard` guard handles this correctly (`typeof model.releaseId === 'string'` will be false), so analytics won't fire for cards without an id. This is fine — but it means any caller that doesn't pass `id` silently gets no analytics. Worth documenting or adding a dev warning.

### Error Handling

**6. `EntityCarousel.tsx` — `IntersectionObserver` callback doesn't guard against `NaN` index**
- **File:** `apps/web/components/organisms/entity-card/EntityCarousel.tsx`
- **Lines:** Inside the observer callback:
  ```ts
  const index = Number(entry.target.getAttribute('data-carousel-index'));
  const model = items[index];
  if (!model) { continue; }
  ```
- **Why it matters:** If `data-carousel-index` is missing or non-numeric, `Number(...)` returns `NaN`, and `items[NaN]` is `undefined`, so the `if (!model)` guard catches this. This is actually handled correctly. No bug.

**None other found.**

### Cross-File Impact

**7. `EntityCarousel` is now a client component but was previously server-renderable**
- **File:** `apps/web/components/organisms/entity-card/EntityCarousel.tsx`
- **Line:** Added `'use client';` at the top.
- **Why it matters:** Any parent that was server-rendering `EntityCarousel` will now get a client component boundary. This is necessary for `IntersectionObserver` and `useRef`, but it means the carousel HTML won't be in the initial SSR payload — it'll hydrate client-side. Check that all callers can handle this (they appear to, since the carousel is interactive).

**8. `EntityCard` now accepts an `onClick` prop**
- **File:** `apps/web/components/organisms/entity-card/EntityCard.tsx`
- **Why it matters:** The `onClick` is passed through to both `<Link>` and `<a>` in `CardShell`. This is additive and backward-compatible. No breakage.

**9. `ProfileCompactSurface` passes `releases` prop through to `ProfileHomeRail`**
- **File:** `apps/web/components/features/profile/templates/ProfileCompactSurface.tsx`
- **Why it matters:** The `releases` prop is new on `ProfileCompactSurface` (implied by the diff showing it passed through). Need to verify that `ProfileCompactSurface`'s own type signature includes `releases` — the diff only shows the pass-through line, not the type definition. If the type isn't updated, this would be a TypeScript error.

### Summary

| 1 | Medium | `EntityCarousel.tsx` | Stale `data-carousel-index` if `items` prop changes at runtime |
| 2 | Medium | `ProfileHomeRail.tsx:208-213` | Featured release may not be found in `releases` array, losing analytics |
| 3 | Low | `ReleaseCatalogCarousel.tsx` | `video` cards with `releaseId` also tracked (may be intentional) |
| 4 | Low | `EntityCarousel.tsx` | `itemRefs.current` not reset on items change |
| 5 | Low | `adapters.ts:120` | Silent no-op when `release.id` not provided |
| 6 | Info | `EntityCarousel.tsx` | `NaN` index handled correctly — no bug |
| 7 | Medium | `EntityCarousel.tsx` | New `'use client'` boundary affects SSR |
| 8 | Info | `EntityCard.tsx` | Additive `onClick` prop — backward compatible |
| 9 | Medium | `ProfileCompactSurface.tsx` | Verify `releases` is in the component's type signature (not shown in diff) |

itstimwhite (commented):

itstimwhite:
### Merge activity

* **Jun 21, 5:30 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jun 21, 5:30 AM UTC**: `itstimwhite` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie).
* **Jun 21, 5:31 AM UTC**: CI is running for this pull request on a draft pull request ([#11378](https://app.graphite.com/github/pr/JovieInc/Jovie/11378)) due to your merge queue CI optimization settings.
* **Jun 21, 5:54 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) via draft PR: [#11378](https://app.graphite.com/github/pr/JovieInc/Jovie/11378).

PR #12072, opened 2026-06-27
fix(chat): fail soft on tool errors with recoverable states (JOV-3338)

itstimwhite:
## Goal
When a chat tool fails (e.g. retouch unprovisioned, album art provider down), users should see specific recoverable inline errors instead of a generic "Something went wrong" stream failure that pauses the composer.

## Changes
- Add `tool-errors.ts` to map stable tool error codes (`TOOL_UNPROVISIONED`, `PROVIDER_UNAVAILABLE`, etc.) to user-facing copy and recovery hints
- Add `wrap-tool-execute.ts` fail-soft wrapper (mirrors `/api/track` audience upsert pattern): tool throws are captured, logged, and returned as structured `{ success: false, errorCode }` payloads
- Wrap all chat tools in `executeChatTurn` via `wrapToolSetFailSoft`
- Persist `errorCode` / `retryable` on tool events; render mapped copy + next-step hints in `tool-ui.tsx`
- Suppress global "Message paused" composer errors for recoverable tool-only failures in `useJovieChat`
- Tag album art tool failures with explicit error codes

## Testing
```bash
cd apps/web && pnpm exec vitest run \
  tests/unit/chat/tool-errors.test.ts \
  tests/unit/chat/wrap-tool-execute.test.ts \
  tests/unit/chat/tool-ui-errors.test.tsx \
  tests/unit/chat/jovie-error-utils.test.ts \
  tests/unit/chat/useJovieChat.rate-limit.test.tsx
```

### DIFF
--- a/apps/web/app/api/chat/route.ts
+++ b/apps/web/app/api/chat/route.ts
@@ -933,6 +933,7 @@ function createGenerateAlbumArtTool(params: {
         return {
           success: false as const,
           retryable: false,
+          errorCode: 'PROFILE_REQUIRED' as const,
           error: 'Profile ID required',
         };
       }
@@ -940,6 +941,7 @@ function createGenerateAlbumArtTool(params: {
         return {
           success: false as const,
           retryable: false,
+          errorCode: 'PLAN_UNAVAILABLE' as const,
           error: 'Album art generation requires a Pro plan.',
         };
       }
@@ -948,6 +950,7 @@ function createGenerateAlbumArtTool(params: {
         return {
           success: false as const,
           retryable: false,
+          errorCode: 'PROVIDER_UNAVAILABLE' as const,
           error: 'Album art generation is temporarily unavailable.',
         };
       }
@@ -985,6 +988,7 @@ function createGenerateAlbumArtTool(params: {
         return {
           success: false as const,
           retryable: true,
+          errorCode: 'RATE_LIMITED' as const,
           error:
             burstLimit.reason ??
             'Album art generation limit reached. Please try again later.',
@@ -998,6 +1002,7 @@ function createGenerateAlbumArtTool(params: {
         return {
           success: false as const,
           retryable: true,
+          errorCode: 'RATE_LIMITED' as const,
           error:
             dailyLimit.reason ??
             'Album art generation limit reached. Please try again later.',
@@ -1097,6 +1102,7 @@ function createGenerateAlbumArtTool(params: {
           return {
             success: false as const,
             retryable: false,
+            errorCode: 'PROVIDER_UNAVAILABLE' as const,
             error: 'Album art generation is temporarily unavailable.',
           };
         }
@@ -1107,6 +1113,7 @@ function createGenerateAlbumArtTool(params: {
         return {
           success: false as const,
           retryable: true,
+          errorCode: 'TOOL_EXECUTION_FAILED' as const,
           error: 'Unable to generate album art. Please try again.',
         };
       }
--- a/apps/web/app/api/chat/route.ts
+++ b/apps/web/app/api/chat/route.ts
@@ -97,6 +97,7 @@ import {
   type ChatTelemetry,
   type ReleaseContext,
 } from '@/lib/chat/types';
+import { wrapToolSetFailSoft } from '@/lib/chat/wrap-tool-execute';
 import { db } from '@/lib/db';
 import { clickEvents, tips } from '@/lib/db/schema/analytics';
 import { users } from '@/lib/db/schema/auth';
@@ -2722,7 +2723,7 @@ export async function POST(req: Request) {
       insightsEnabled,
       forceLightModel,
       lastUserText: userText,
-      tools,
+      tools: wrapToolSetFailSoft(tools),
       signal: req.signal,
       requestId,
       telemetry,
--- a/apps/web/components/jovie/JovieChat.tsx
+++ b/apps/web/components/jovie/JovieChat.tsx
@@ -510,14 +510,15 @@ export function JovieChat({
     />
   );
 
-  const inlineChatError = chatError ? (
-    <ChatInlineError
-      chatError={chatError}
-      onRetry={handleRetry}
-      isLoading={isLoading}
-      isSubmitting={isSubmitting}
-    />
-  ) : null;
+  const inlineChatError =
+    chatError && !chatError.suppressComposerPause ? (
+      <ChatInlineError
+        chatError={chatError}
+        onRetry={handleRetry}
+        isLoading={isLoading}
+        isSubmitting={isSubmitting}
+      />
+    ) : null;
 
   return (
     <EntityResolutionProvider profileId={profileId}>
--- a/apps/web/components/jovie/components/ErrorDisplay.test.tsx
+++ b/apps/web/components/jovie/components/ErrorDisplay.test.tsx
@@ -28,7 +28,7 @@ describe('ErrorDisplay', () => {
     expect(screen.getByText('Message paused')).toBeInTheDocument();
     expect(screen.getByText('CHAT_TIMEOUT · req_123')).toBeInTheDocument();
 
-    fireEvent.click(screen.getByRole('button', { name: 'Retry message' }));
+    fireEvent.click(screen.getByRole('button', { name: 'Retry Message' }));
     expect(onRetry).toHaveBeenCalledTimes(1);
   });
 });
--- a/apps/web/components/jovie/components/ErrorDisplay.tsx
+++ b/apps/web/components/jovie/components/ErrorDisplay.tsx
@@ -49,6 +49,12 @@ export function ErrorDisplay({
     }
   };
 
+  const isToolScopedFailure =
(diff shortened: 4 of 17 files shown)


itstimwhite:
### Merge activity

* **Jun 27, 6:46 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jun 27, 7:57 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jun 27, 8:01 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jun 27, 8:03 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.

itstimwhite:
# Code Review: PR #12072 — fix(chat): fail soft on tool errors with recoverable states

### Logic Errors

**1. Race condition: `sdkMessagesRef` is read stale in `onError`**

`apps/web/components/jovie/hooks/useJovieChat.ts` — the `sdkMessagesRef` is updated via a `useEffect` (line ~606 in the diff), which means it always lags one render behind. In the `onError` callback (line ~509), you read `sdkMessagesRef.current` to get the last assistant message's parts. If the stream callback and the state update for `sdkMessages` resolve in the same tick, the ref may not yet contain the latest assistant message with the failed tool part. This means `shouldSuppressChatPauseForToolFailure` could return `false` when it should return `true`, causing a composer pause for a recoverable tool failure.

**Why it matters:** The core feature — suppressing composer pause for tool failures — can silently fail under normal React batching, producing exactly the bug this PR is trying to fix.

**Suggested fix:** Either update the ref synchronously (e.g., via a `useLayoutEffect` or by tracking the last assistant message in a separate ref that is updated directly in the stream event handler), or pass `sdkMessages` directly into the `onError` closure if it's available in scope.

---

**2. `messagePartsIncludeFailedTool` structural type-check is fragile**

`apps/web/components/jovie/utils.ts` — the function uses `typeof part.type !== 'string'` and `'state' in part` checks, but the `MessagePart` type from the `ai` package is a discriminated union. The check `part.type.startsWith('tool-') && part.type !== 'tool-invocation'` will match hypothetical future tool-related part types (e.g., a `tool-metadata` part) that aren't actually invocation results, potentially causing false positives.

**Why it matters:** False positives mean the composer pause gets suppressed for messages that didn't actually have a failed tool, hiding real errors from users.

**Suggested fix:** Narrow the check to the specific tool part types you actually handle (`dynamic-tool`, `tool-invocation` with a `state` of `output-error`, etc.) rather than using a prefix match with an exclusion.

---

**3. `resolveToolFailurePresentation` ignores caller-provided `errorMessage` when it's the empty string**

`apps/web/lib/chat/tool-errors.ts` line ~230 — the condition `input.errorMessage && input.errorMessage.trim().length > 0` is falsy for `null`/`undefined` (correct) but the caller in `tool-ui.tsx` passes `event.errorMessage ?? event.summary` as `errorMessage`. If `event.errorMessage` is an empty string, it falls through to `defaults.body`, which is correct. However, if a tool explicitly returns an empty `error` field in its payload (e.g., `{ success: false, error: '' }`), `normalizeToolFailureOutput` also falls through to `'Tool execution failed.'` — but `resolveToolFailurePresentation` would then use that fallback. This is fine, but the `body` in `resolveToolFailurePresentation` can return `undefined` from the `??` chain if `defaults.body` is ever missing from `USER_COPY`, which would cause a crash. The `USER_COPY` map is exhaustive over `ToolErrorCode`, so this is safe *today*, but adding a new code to `TOOL_ERROR_CODES` without adding it to `USER_COPY` would be a silent runtime crash.

**Why it matters:** Future maintainers adding a new error code would get a runtime crash with no compile-time safety.

**Suggested fix:** Add a compile-time exhaustive check (e.g., a `satisfies` assertion or a switch with `never` fallback) or provide a final fallback body in `resolveToolFailurePresentation`.

---

**4. `getToolStatusBody` switch has a missing `break` between `failed` and `denied`**

`apps/web/components/jovie/tool-ui.tsx` — the `case 'failed':` block now has a `return presentation.body;` inside a block, but the `case 'denied':` immediately follows. Looking at the diff, the `failed` case returns early, so there's no fall-through issue. However, the closing `}` of the `case 'failed':` block and the `case 'denied':` label are at the same indentation level, which is correct. **No issue here on closer inspection.**

---

**5. `onError` in `useJovieChat` dispatches `assistant.stream.completed` but does not add the failed tool part to the timeline**

`apps/web/components/jovie/hooks/useJovieChat.ts` line ~509 — when `shouldSuppressChatPauseForToolFailure` returns `true`, the handler dispatches `assistant.stream.completed` with `assistantParts` read from `sdkMessagesRef.current`. If the ref is stale (issue #1), the parts may not include the failed tool. Even if not stale, the parts are read from the last assistant message, which may or may not include the failed tool part depending on whether the stream had already appended it. If the tool part is missing from the dispatched parts, the UI will show the stream as completed with no visible error — the user sees nothing wrong, but also sees no indication of what happened.

**Why it matters:** Silent failures with no UI indication are worse than a paused composer — users don't know an action failed.

**Suggested fix:** Ensure the failed tool part is included in the dispatched parts, or dispatch a separate `tool.failed` event before completing.

---

### Security

**None found.** No hardcoded secrets, no SQL injection vectors, no XSS via `dangerouslySetInnerHTML`, no auth bypass. The `tool-errors.ts` file is pure mapping logic with no I/O. The `wrap-tool-execute.ts` wrapper catches errors without leaking internals to the client.

---

### Data Integrity

**6. `persistedToolEventSchema` adds optional fields — no migration for existing records**

`apps/web/lib/chat/tool-events.ts` — `errorCode` and `retryable` are added as `.optional()` to the zod schema. Existing persisted tool events in the database will lack these fields. The `normalizeToolPart` function reads `output.errorCode` from the tool result, which is set by the new `wrapToolSetFailSoft` wrapper. However, for tool events persisted *before* this change (no `errorCode` on the output), `normalizeToolFailureOutput` will return `null` (because `output.success !== false` may not match for old records that have a different shape), so `normalizedFailure` will be `null`, and the old `output`/`errorMessage`/`summary` will be used as-is. This is safe. **No issue.**

---

### Error Handling

**7. `captureError` in `withFailSoftToolExecute` is awaited but could reject**

`apps/web/lib/chat/wrap-tool-execute.ts` line ~38 — `await captureError(...)` is called without a try/catch. If `captureError` throws or rejects (e.g., the error tracking service is down), the rejection will propagate out of the wrapped `execute`, which will then propagate out of the tool execution and into the AI SDK's tool runner. The AI SDK may treat this as a tool failure (which is fine), but it could also crash the stream depending on its error handling.

**Why it matters:** Error tracking should never be on the critical path. The whole point of this PR is that tool failures shouldn't crash the stream — but a tracking failure could.

**Suggested fix:** Wrap the `captureError` call in a try/catch with a no-op fallback, or use `.catch(() => {})` on the promise.

---

### Cross-File Impact

**8. `getNextStepMessage` signature change is not propagated to all callers**

`apps/web/components/jovie/utils.ts` — `getNextStepMessage` now accepts an optional `errorCode` parameter. The `ErrorDisplay.tsx` component passes `chatError.errorCode` (line ~72 in the diff). However, any other callers of `getNextStepMessage` that might exist outside this diff would continue to work (the parameter is optional), so this is safe. **No issue.**

---

**9. `ChatErrorType` union expanded with `'tool'` — exhaustive switches must be updated**

`apps/web/components/jovie/types.ts` — the `ChatErrorType` union now includes `'tool'`. The `getErrorType` function in `utils.ts` handles it. The `getUserFriendlyMessage` function handles it. The `getNextStepMessage` function handles it. The `ErrorDisplay.tsx` uses `isToolScopedFailure` which checks for `chatError.type === 'tool'`. All known consumers are updated. **No issue.**

---

### Summary

| 1 | **High** | `useJovieChat.ts` | `sdkMessagesRef` updated via `useEffect` is one render stale; `onError` reads wrong parts |
| 5 | **High** | `useJovieChat.ts` | Suppressed tool failure may not include failed tool part in dispatched completed event — silent failure |
| 7 | **Medium** | `wrap-tool-execute.ts` | `await captureError()` without try/catch can reject and crash the stream |
| 3 | **Low** | `tool-errors.ts` | No compile-time exhaustive check on `USER_COPY` covering all `ToolErrorCode` variants |
| 2 | **Low** | `utils.ts` | `messagePartsIncludeFailedTool` prefix match is fragile against future `tool-*` part types |

Issues 1 and 5 are the most likely to manifest in production and directly undermine the PR's goal. I'd recommend addressing those before merge.

itstimwhite (commented):

PR #12301, opened 2026-06-28
feat(dashboard): ship unified Jovie work activity surface (JOV-3624)

itstimwhite:
## Goal

Surface Jovie's autonomous work to artists via a read-only unified feed over seven existing tables — proving the "Jovie runs your business while you make music" thesis.

## Changes

- Add `GET /api/dashboard/jovie-work/recent` aggregating `workflowRuns`, `agentRuns`, `suggestedActions`, `retouchJobs`, `merchFulfillmentJobs`, `metadataSubmissionRequests`, and `fanReleaseNotifications`
- Add mapper/merge layer in `lib/activity/jovie-work-feed.ts` with phase labels (pending, in progress, done, failed)
- Add artist-facing `/app/jovie-work` page with `JovieWorkFeed` component ("Jovie Did This")
- Add TanStack query hook, validation schema, and unit tests

## Testing

- `pnpm vitest run tests/unit/lib/jovie-work-feed.test.ts tests/unit/dashboard/JovieWorkFeed.test.tsx`
- `pnpm turbo typecheck --filter=@jovie/web`
- Visit `/app/jovie-work` while authenticated

---
*Generated by Jovie Agent*


itstimwhite:
# Code Review: PR #12301 — Jovie Work Activity Surface

### Logic Errors

**1. `agentRuns` query selects `createdAt` but it's not in the diff**

`apps/web/lib/activity/load-jovie-work-feed.ts` (agentRuns select block)

The `mapAgentRunToJovieWorkItem` function accepts an optional `createdAt` field and uses it as a fallback in the timestamp chain `input.completedAt ?? input.startedAt ?? input.createdAt ?? new Date()`. However, the `agentRuns` select query in `loadJovieWorkFeed` does **not** include `createdAt` in its select projection. If both `completedAt` and `startedAt` are null, the fallback goes to `input.createdAt ?? new Date()` — but since `createdAt` is not selected, it will be `undefined`, and the code will fall through to `new Date()` silently. This is technically safe due to the `new Date()` final fallback, so no crash — but it means agent runs with null timestamps will show "just now" rather than their actual creation time. **Low severity, no data loss, but worth noting for accuracy.**

**2. `mergeJovieWorkItems` sorts by ISO string comparison — correct only if all timestamps are true ISO strings**

`apps/web/lib/activity/jovie-work-feed.ts`, line containing `mergeJovieWorkItems`

```ts
.sort((left, right) => right.timestamp.localeCompare(left.timestamp))
```

`localeCompare` on ISO-8601 strings does **not** produce the same ordering as chronological sort in all cases. For example, `"2025-01-01T00:00:00Z"` vs `"2025-01-01T00:00:00.000Z"` — locale compare may order these differently than `Date` comparison. Since all timestamps in this codebase come from `Date.toISOString()` (which always produces `.000Z` format), this is safe **today**, but if any source ever produces a non-canonical ISO string (e.g., from a DB driver returning `timestamp` without millisecond precision), the sort will silently misorder. **Suggest:** `.sort((a, b) => new Date(b.timestamp).getTime() - new Date(a.timestamp).getTime())` for robustness.

**3. `perSourceLimit` heuristic can under-fetch for multi-source aggregation**

`apps/web/lib/activity/load-jovie-work-feed.ts`, line containing `perSourceLimit`

```ts
const perSourceLimit = Math.min(25, Math.max(5, input.limit * 2));
```

With `limit=20` (default), `perSourceLimit = 40`. Each of 7 sources fetches up to 40 rows. After merging and sorting, the top 20 are returned. This is fine when data is evenly distributed, but if one source (e.g., `fan_notification`) has 40 recent items and all other sources are empty, you get 40 fan notifications and nothing else — which is correct behavior. However, if all 7 sources have exactly 40 items with interleaved timestamps, you're fetching 280 rows to return 20. **Not a bug**, but the `perSourceLimit` could be tightened to `Math.ceil(input.limit * 1.5 / 7)` per source for better efficiency at high limits. **Low priority optimization.**

### Security

**None found.**

- Route uses `withDashboardRoute` (auth preserved).
- Zod schema validates `limit` (1–50) and `range` (constrained enum).
- No raw SQL — all queries use Drizzle parameterized builders.
- No user-controlled strings rendered as HTML — all content goes through React's text rendering.
- `parseJovieWorkFeedResponse` validates all fields before use.
- No new CSP domains introduced.

### Data Integrity

**4. `metadata_submission` and `fan_notification` queries filter by `creatorProfileId` but `workflowRuns`, `agentRuns`, `suggestedActions`, `retouchJobs` filter by `userId` — mixed ownership model**

`apps/web/lib/activity/load-jovie-work-feed.ts`

The route passes both `userId: ctx.user.id` and `creatorProfileId: ctx.profile.id`. Some queries use one, some use the other. This is **intentional** (workflows belong to users, merch/metadata/notifications belong to creator profiles), but there's no comment explaining why. If a user has multiple profiles, the feed will show profile-scoped items only for the currently selected profile but user-scoped items across all profiles. **This is a product decision, not a bug**, but a code comment documenting the intent would prevent future regressions.

**5. `merchFulfillmentJobs` query uses `innerJoin` on `merchOrders` — will silently drop orphaned fulfillment jobs**

`apps/web/lib/activity/load-jovie-work-feed.ts` (merchFulfillmentRows query)

```ts
.innerJoin(merchOrders, eq(merchFulfillmentJobs.merchOrderId, merchOrders.id))
```

If a `merchFulfillmentJob` references a `merchOrderId` that doesn't exist (deleted order, data migration artifact), that job will be silently excluded from the feed. For a read-only activity surface, this is acceptable — but consider whether a `leftJoin` with a null check on `merchOrders.creatorProfileId` would be more resilient. **Low severity.**

### Error Handling

**6. `fetchJovieWorkFeed` does not check `response.ok` before parsing**

`apps/web/lib/queries/useJovieWorkFeedQuery.ts`, `fetchJovieWorkFeed` function

```ts
const response = await fetchWithTimeout<unknown>(
  `/api/dashboard/jovie-work/recent?${params.toString()}`,
  { signal }
);
return parseJovieWorkFeedResponse(response);
```

The function casts the raw `Response` object to `<unknown>` and passes it directly to `parseJovieWorkFeedResponse`. If the server returns a 400 or 500 JSON body like `{ error: 'Invalid request' }`, `parseJovieWorkFeedResponse` will check `isRecord(value)` (passes), then check `Array.isArray(items)` (fails, since the error object has no `items` array), and return `[]`. The user sees an empty feed with no error indication. **The TanStack query's `error` property will never be set** because the fetch doesn't throw. **Suggest:** Check `response.ok` and throw on non-2xx, or at minimum check for an `error` field in the response and throw.

### Cross-File Impact

**7. `queryKeys.dashboard.jovieWorkFeed` is exported but `ActivityRange` type import in `types.ts` re-exports from a different path than the existing pattern**

`apps/web/components/features/dashboard/organisms/jovie-work-feed/types.ts`

```ts
import type { ActivityRange } from '@/features/dashboard/organisms/dashboard-activity-feed/types';
```

The existing codebase (per `useJovieWorkFeedQuery.ts`) imports `ActivityRange` from `@/lib/validation/schemas/dashboard/activity`. Both resolve to the same type, but having two import paths for the same type creates a maintenance risk if one path is later renamed. **Cosmetic, but should be unified.**

---

### Summary

| # | Severity | Category | File | Issue |
|---|----------|----------|------|-------|
| 1 | Low | Logic | `load-jovie-work-feed.ts` | `agentRuns` doesn't select `createdAt` — silent fallback to `new Date()` |
| 2 | Low | Logic | `jovie-work-feed.ts` | `localeCompare` sort is fragile for non-canonical ISO strings |
| 3 | Low | Logic | `load-jovie-work-feed.ts` | `perSourceLimit` heuristic is loose — optimization opportunity |
| 4 | Info | Data Integrity | `load-jovie-work-feed.ts` | Mixed `userId` vs `creatorProfileId` filtering — needs comment |
| 5 | Low | Data Integrity | `load-jovie-work-feed.ts` | `innerJoin` on merch orders drops orphaned jobs silently |
| 6 | **Medium** | Error Handling | `useJovieWorkFeedQuery.ts` | Non-2xx responses silently return empty array — no error surfaced to UI |
| 7 | Info | Cross-file | `types.ts` | Duplicate import path for `ActivityRange` |

**Issue #6 is the only one I'd recommend blocking merge for.** The rest are low-severity or informational.

itstimwhite (commented):

itstimwhite:
# Code Review: PR #12301 — Jovie Work Activity Surface

### Logic Errors

**1. `load-jovie-work-feed.ts:107` — `agentRuns` select omits `createdAt` but mapper expects it**

```typescript
db
  .select({
    id: agentRuns.id,
    agentSlug: agentRuns.agentSlug,
    status: agentRuns.status,
    completedAt: agentRuns.completedAt,
    startedAt: agentRuns.startedAt,
  })
```

`mapAgentRunToJovieWorkItem` (jovie-work-feed.ts:280) has `createdAt?: Date | string | null` in its input type and uses it as a fallback in the timestamp chain `input.completedAt ?? input.startedAt ?? input.createdAt ?? new Date()`. The query never selects `createdAt`, so it will always be `undefined`, falling through to `new Date()` — meaning agent run items always get the server's current timestamp instead of their actual creation time. This breaks sort ordering in the merged feed.

**Fix:** Add `createdAt: agentRuns.createdAt` to the select object, or remove `createdAt` from the mapper's input type and fallback chain if it's intentionally unused.

---

**2. `load-jovie-work-feed.ts:120` — `or(gte(startedAt, since), gte(completedAt, since))` with nullable columns**

```typescript
or(gte(agentRuns.startedAt, since), gte(agentRuns.completedAt, since))
```

`startedAt` and `completedAt` are nullable timestamp columns. In SQL, `NULL >= value` evaluates to `NULL` (falsy), which is fine — those rows are excluded. However, this means an agent run that *started* before `since` but *completed* after `since` is correctly captured, but a run that started after `since` and is still running (no `completedAt`) is also captured. This is actually correct behavior. **Retracted — this is fine.**

---

**3. `jovie-work-feed.ts:448` — `mergeJovieWorkItems` sorts by ISO string using `localeCompare`**

```typescript
.sort((left, right) => right.timestamp.localeCompare(left.timestamp))
```

ISO 8601 strings sort correctly lexicographically. `localeCompare` is locale-aware and can produce different orderings depending on the runtime locale, and is also slower than plain string comparison. This won't produce *wrong* results for ISO strings, but it's inconsistent with the rest of the codebase and introduces unnecessary locale-dependent behavior.

**Fix:** Use `right.timestamp.localeCompare(left.timestamp)` → `right.timestamp < left.timestamp ? -1 : right.timestamp > left.timestamp ? 1 : 0` or simply `right.timestamp > left.timestamp ? -1 : 1`.

---

**4. `useJovieWorkFeedQuery.ts:35` — `retry: 3` with exponential backoff up to 30s, but no distinction between error types**

The query retries on all errors including 400-level validation failures (though those won't throw from `fetchWithTimeout`). More importantly, a 500 from the server will retry 3 times with delays of 1s, 2s, 4s — totaling ~7s of perceived hang time before the error state shows. For a dashboard feed, this is aggressive. Consider `retry: 1` or a custom `retry` function that only retries on network errors / 5xx.

**Not a bug, but worth discussing** — the UX impact is that users see a skeleton for ~7s before an error message appears on server failures.

---

### Security

**5. `route.ts:22` — Zod `preprocess` on `limit` accepts any string input**

```typescript
limit: z.preprocess(
  val => Number(val ?? 20),
  z.number().int().min(1).max(50)
),
```

`Number("99999999999999999999999999")` returns `Infinity`, which fails `.int()` — that's handled. However, `Number("1e308")` is `Infinity`, also handled. `Number("0.5")` passes `.int()` check? No — `z.number().int()` rejects floats. This is fine. **No actual vulnerability here.**

---

**6. `route.ts:22` — `range` defaults to `'7d'` but the schema uses `activityRangeSchema`**

The `activityRangeSchema` (referenced from `./activity`) constrains to `'7d' | '30d' | '90d'`. The `.default('7d')` on `range` means an omitted param defaults correctly. **No issue.**

---

**7. No CSP or auth concerns** — route uses `withDashboardRoute`, user ID comes from `ctx.user.id` (server-side, not client-passed). Profile ID comes from `ctx.profile.id`. The client query hook passes only `range` as a query param; `profileId` is used only for the query key, not sent to the API. **Correct.**

---

### Data Integrity

**8. `load-jovie-work-feed.ts:155` — `merchFulfillmentJobs` filtered by `creatorProfileId` via join, but other sources filtered by `userId`**

```typescript
.eq(merchOrders.creatorProfileId, input.creatorProfileId),
```

The merch fulfillment query joins `merchOrders` and filters on `creatorProfileId`, while all other sources filter on `userId`. This is intentional (merch orders are scoped to creator profiles, not users), but it means the feed mixes two different scoping models. If a user has multiple profiles, they'll see merch fulfillments from all profiles they have orders under, but only see other activity types for their user ID. **This is likely correct given the domain model**, but should be documented with a comment explaining the dual-scoping rationale.

---

**9. `load-jovie-work-feed.ts:172` — `metadataSubmissionRequests` filtered by `creatorProfileId`, not `userId`**

Same pattern as merch — metadata submissions and fan notifications are scoped to `creatorProfileId`. The route passes both `userId` and `creatorProfileId` from context. **Verify that `ctx.profile.id` is the correct scope for these tables** — if a user switches profiles in the dashboard, the metadata/fan notification results should change accordingly. From the `DashboardDataContext` usage in `JovieWorkPanel.tsx`, `selectedProfile.id` is used, which is correct.

---

**10. `load-jovie-work-feed.ts` — No `createdAt` filter on `metadataSubmissionRequests` and `fanReleaseNotifications` beyond `gte(createdAt, since)`**

All seven sources filter by their `createdAt >= since`. This is consistent. **No issue.**

---

### Error Handling

**11. `route.ts:38` — `captureError` is awaited but the route continues to return 500 regardless**

```typescript
} catch (error) {
    await captureError('Jovie work feed fetch failed', error, { ... });
    return NextResponse.json({ error: 'Failed to load Jovie work feed' }, { status: 500, ... });
}
```

This is correct — error is captured and a user-facing 500 is returned. **No issue.**

---

**12. `useJovieWorkFeedQuery.ts` — No error boundary or error handling in the query hook**

The `JovieWorkFeed` component handles `error` from the query (line ~167 in the component), displaying `error.message`. This is fine for the happy path. However, if `parseJovieWorkFeedResponse` throws (e.g., if `fetchWithTimeout` returns a non-typed response that's not a valid object), the error propagates to the query's `error` state. **This is handled by the component's error branch.** No issue.

---

### Cross-File Impact

**13. `lib/queries/index.ts` — New export `useJovieWorkFeedQuery` added**

Standard barrel export addition. No breaking changes.

**14. `lib/queries/keys.ts` — New `jovieWorkFeed` key added under `dashboard` namespace**

Correctly nested under `dashboard.all` prefix. The test at `tests/lib/queries/keys.test.ts:117` validates the structure. **No issue.**

**15. `constants/routes.ts` — New `JOVIE_WORK: '/app/jovie-work'` route added**

No conflicts with existing routes. **No issue.**

---

### Summary

| Category | Count | Details |
|----------|-------|---------|
| Logic bugs | 1 | **#1** — `agentRuns` missing `createdAt` select causes incorrect timestamps |
| Security | 0 | — |
| Data integrity | 0 | — |
| Error handling | 0 | — |
| Cross-file impact | 0 | — |

**Required fix:** Add `createdAt: agentRuns.createdAt` to the select in `load-jovie-work-feed.ts:107` so `mapAgentRunToJovieWorkItem` receives a meaningful fallback timestamp. Without this, all agent-run items in the feed display "just now" regardless of when they actually ran, breaking the merged chronological sort.

itstimwhite (commented):

itstimwhite:
# PASS 1 — Logic, Security, Bugs

## Logic Errors

### 1. `load-jovie-work-feed.ts:107-113` — Agent run timestamp fallback can produce `Invalid Date`

```ts
const timestampSource =
  input.completedAt ?? input.startedAt ?? input.createdAt ?? new Date();
```

The `mapAgentRunToJovieWorkItem` function signature declares `createdAt?: Date | string | null` (optional). When all three are `null`, `new Date()` is used — this is fine. However, the DB query at lines 119-126 does **not** select `createdAt` from `agentRuns`:

```ts
.select({
  id: agentRuns.id,
  agentSlug: agentRuns.agentSlug,
  status: agentRuns.status,
  completedAt: agentRuns.completedAt,
  startedAt: agentRuns.startedAt,
})
```

So `input.createdAt` is always `undefined`, meaning the fallback chain resolves to `undefined ?? undefined ?? undefined ?? new Date()` → `new Date()`. This works at runtime but is misleading — the `createdAt` field is dead code in the mapper. More critically, if both `completedAt` and `startedAt` are null (possible for a brand-new run), the timestamp becomes "now" rather than the actual creation time, which misorders the item in the feed.

**Why it matters:** Items with null timestamps sort to the top of the feed as if they just happened, even if they're old. This corrupts the chronological ordering that `mergeJovieWorkItems` relies on.

**Fix:** Either select `createdAt` in the query, or use `createdAt` as the primary timestamp source for agent runs (since it's the only guaranteed-non-null field).

---

### 2. `load-jovie-work-feed.ts:102-115` — Agent run date filter uses `OR` on nullable columns

```ts
.or(gte(agentRuns.startedAt, since), gte(agentRuns.completedAt, since))
```

`startedAt` and `completedAt` are nullable. In SQL, `gte(null, since)` evaluates to `NULL` (not `false`), so `OR(NULL, NULL)` = `NULL` → row excluded. This means agent runs where both `startedAt` and `completedAt` are null are silently dropped from the feed.

**Why it matters:** Brand-new agent runs that haven't started yet are invisible to the feed, contradicting the "pending" phase the mapper supports.

**Fix:** Add `createdAt` to the select and use it in the `OR` clause, or add an `isNull` guard:
```ts
or(
  gte(agentRuns.startedAt, since),
  gte(agentRuns.completedAt, since),
  gte(agentRuns.createdAt, since)
)
```

---

### 3. `jovie-work-feed.ts:493-497` — `mergeJovieWorkItems` uses `localeCompare` for ISO timestamps

```ts
.sort((left, right) => right.timestamp.localeCompare(left.timestamp))
```

While ISO 8601 strings do sort correctly with `localeCompare` in most cases, this is locale-dependent behavior. `localeCompare` can apply locale-specific collation rules that may not preserve chronological order for all string formats.

**Why it matters:** Subtle sorting bugs in edge locales. The codebase uses ISO 8601 strings — a simple `>` comparison or `Date` subtraction would be both faster and unambiguous.

**Fix:**
```ts
.sort((left, right) => right.timestamp > left.timestamp ? -1 : 1)
```
Or more idiomatically for ISO strings:
```ts
.sort((left, right) => left.timestamp < right.timestamp ? 1 : -1)
```

---

### 4. `useJovieWorkFeedQuery.ts:38` — `profileId` not included in query key dependencies for fetch function

```ts
queryFn: ({ signal }) => fetchJovieWorkFeed(range, signal),
```

The `fetchJovieWorkFeed` function only takes `range` — it doesn't send `profileId` to the API. The query key includes `profileId`, so switching profiles will correctly invalidate the cache, but the actual HTTP request doesn't include `profileId` as a query param. The route handler gets `profileId` from `ctx.profile.id` (Clerk session), not from the request URL.

**Why it matters:** This is actually correct by design (profile comes from auth context), but it means the `profileId` in the query key is purely for cache invalidation, not for the request. This is fine but worth documenting, since a future developer might expect `profileId` to be a query param. No action needed if this is intentional.

**Verdict:** Not a bug — the route uses `ctx.profile.id` from the auth context. But verify that `withDashboardRoute` correctly resolves the profile for the current user.

---

## Security

### 5. `route.ts:14-18` — Zod `preprocess` on `limit` accepts non-numeric strings

```ts
limit: z.preprocess(
  val => Number(val ?? 20),
  z.number().int().min(1).max(50)
),
```

`Number("abc")` returns `NaN`, which fails `z.number().int()` — this is handled correctly. However, `Number("1e2")` returns `100`, which passes `.int()` but exceeds `.max(50)`. This is fine. No actual vulnerability here.

**Verdict:** Safe. The schema correctly rejects invalid input.

---

### 6. `route.ts:37-41` — Cache headers on error responses

Error responses use `'Cache-Control': 'no-store'` — correct. Success responses use `private, max-age=60, stale-while-revalidate=120` — correct for user-specific data.

**Verdict:** No issues.

---

### 7. `load-jovie-work-feed.ts` — No SQL injection risk

All queries use Drizzle's parameterized query builder (`eq`, `gte`, `and`, `or`). No raw SQL or string concatenation.

**Verdict:** Safe.

---

### 8. `JovieWorkFeed.tsx` — XSS via `item.description`

The feed renders `item.description` as JSX text content (`<span>{item.description}</span>`), which React automatically escapes. The `item.title` is also rendered as text content. No `dangerouslySetInnerHTML` usage.

**Verdict:** Safe.

---

## Data Integrity

### 9. `load-jovie-work-feed.ts:204-213` — `merchFulfillmentJobs` joined on `merchOrders` but `merchOrderId` relationship assumed

```ts
.innerJoin(
  merchOrders,
  eq(merchFulfillmentJobs.merchOrderId, merchOrders.id)
)
.where(
  and(
    eq(merchOrders.creatorProfileId, input.creatorProfileId),
    gte(merchFulfillmentJobs.createdAt, since)
  )
)
```

This filters merch fulfillment jobs to only those where the associated order belongs to the creator profile. If `merchFulfillmentJobs.merchOrderId` is nullable or if the join fails (orphaned fulfillment job), the row is silently excluded.

**Why it matters:** Data inconsistency between `merchFulfillmentJobs` and `merchOrders` could cause items to disappear from the feed without any error.

**Verdict:** Likely fine if FK constraints exist at the DB level. No action needed unless orphaned records are possible.

---

### 10. `load-jovie-work-feed.ts:170-191` — `leftJoin` on `discogReleases` for metadata and fan notifications

Both metadata submissions and fan notifications use `leftJoin` with `discogReleases`. If the release is deleted but the notification/submission remains, `releaseTitle` will be `null`. The mappers handle this with fallback to `'a release'`:

```ts
const releaseLabel = input.releaseTitle?.trim() || 'a release';
```

**Verdict:** Correctly handled.

---

## Error Handling

### 11. `route.ts:33-43` — `captureError` is awaited but failure is not handled

```ts
catch (error) {
  await captureError('Jovie work feed fetch failed', error, {
    route: '/api/dashboard/jovie-work/recent',
    method: 'GET',
    profileId: ctx.profile.id,
  });

  return NextResponse.json(
    { error: 'Failed to load Jovie work feed' },
    { status: 500, headers: { 'Cache-Control': 'no-store' } }
  );
}
```

If `captureError` itself throws (e.g., network error to error tracking service), the exception propagates unhandled, resulting in a 500 with no body (Next.js default error response).

**Why it matters:** The user gets a generic error instead of the intended JSON error message. The error tracking failure masks the original error.

**Fix:**
```ts
catch (error) {
  try {
    await captureError('Jovie work feed fetch failed', error, {
      route: '/api/dashboard/jovie-work/recent',
      method: 'GET',
      profileId: ctx.profile.id,
    });
  } catch {
    // Swallow — error tracking is best-effort
  }

  return NextResponse.json(
    { error: 'Failed to load Jovie work feed' },
    { status: 500, headers: { 'Cache-Control': 'no-store' } }
  );
}
```

---

### 12. `useJovieWorkFeedQuery.ts:44-48` — Aggressive retry configuration

```ts
retry: 3,
retryDelay: attemptIndex => Math.min(1000 * 2 ** attemptIndex, 30_000),
```

With `refetchInterval: 5 * 60 * 1000` and `retry: 3`, a transient failure during background refetch will retry 3 times with exponential backoff (1s, 2s, 4s). This is reasonable.

**Verdict:** No issue.

---

## Cross-File Impact

### 13. `constants/routes.ts:52` — New route `JOVIE_WORK` added

```ts
JOVIE_WORK: '/app/jovie-work',
```

This is a new route addition. No existing code references this route yet, so no breakage. The route is correctly added to `APP_ROUTES`.

**Verdict:** Safe.

---

### 14. `lib/queries/index.ts:397` — New export added

```ts
export { useJovieWorkFeedQuery } from './useJovieWorkFeedQuery';
```

Additive change. No existing exports modified.

**Verdict:** Safe.

---

### 15. `lib/validation/schemas/dashboard/index.ts:38-42` — New schema export added

Additive change. No existing exports modified.

**Verdict:** Safe.

---

## Summary

| 1 | **Medium** | `load-jovie-work-feed.ts:107-113` | Agent run mapper has dead `createdAt` fallback; items with null `completedAt`/`startedAt` get `new Date()` timestamp, misordering them in feed |
| 2 | **Medium** | `load-jovie-work-feed.ts:102-115` | Agent run query uses `OR` on nullable columns — rows where both are null are excluded from feed |
| 3 | **Low** | `jovie-work-feed.ts:493` | `localeCompare` for ISO timestamp sort works but is semantically incorrect and potentially locale-sensitive |
| 11 | **Low** | `route.ts:35` | `captureError` failure in catch block propagates unhandled, masking the intended JSON error response |

**No security issues found.** SQL injection, XSS, and auth bypass are all properly handled.

itstimwhite (commented):

itstimwhite:
# Code Review — PR #12301

### Logic Errors

**1. `load-jovie-work-feed.ts:100-108` — Agent run timestamp fallback can produce `Invalid Date`**

```ts
const timestampSource =
  input.completedAt ?? input.startedAt ?? input.createdAt ?? new Date();
```

The `mapAgentRunToJovieWorkItem` function in `jovie-work-feed.ts` accepts `createdAt?: Date | string | null` (optional). When all three fields are `null`, the fallback `new Date()` is fine. However, the DB query at `load-jovie-work-feed.ts:100-108` does **not** select `createdAt` from `agentRuns`:

```ts
.select({
  id: agentRuns.id,
 Runs.agentSlug,
  status: agentRuns.status,
  completedAt: agentRuns.completedAt,
  startedAt: agentRuns.startedAt,
})
```

So `input.createdAt` is always `undefined`, meaning the `?? new Date()` fallback silently fires for every agent run where `completedAt` and `startedAt` are both null. This masks data quality issues and produces a "now" timestamp for historical records. The mapper's `createdAt` parameter is dead code given this loader.

**Why it matters:** Items with null timestamps get sorted to the top of the feed as if they just happened, corrupting the chronological ordering.

**Fix:** Either (a) add `createdAt` to the select list so the fallback chain works as designed, or (b) handle the all-null case explicitly with a guard that logs/throws.

---

**2. `jovie-work-feed.ts:393-394` — `mergeJovieWorkItems` uses `localeCompare` for ISO timestamps**

```ts
.sort((left, right) => right.timestamp.localeCompare(left.timestamp))
```

ISO 8601 timestamps sort correctly with `localeCompare`, but this is an accidental correctness — `localeCompare` is locale-aware and can produce unexpected results with certain locale settings or non-ASCII characters. For ISO date strings, subtraction via `Date` or direct string comparison (`>`) is both faster and unambiguous.

**Why it matters:** Fragile correctness; localeCompare is semantically wrong for numeric/ISO ordering and adds unnecessary overhead.

**Fix:** Use `right.timestamp < left.timestamp ? -1 : 1` or `new Date(right.timestamp).getTime() - new Date(left.timestamp).getTime()`.

---

**3. `useJovieWorkFeedQuery.ts:21-28` — `fetchJovieWorkFeed` ignores `profileId`**

```ts
async function fetchJovieWorkFeed(
  range: ActivityRange,
  signal?: AbortSignal
): Promise<JovieWorkItem[]> {
  const params = new URLSearchParams({ range });
  const response = await fetchWithTimeout<unknown>(
    `/api/dashboard/jovie-work/recent?${params.toString()}`,
    { signal }
  );
```

The `profileId` is passed into the hook options but never sent to the API. The route handler (`route.ts:28-33`) derives `creatorProfileId` from `ctx.profile.id` (Clerk session), so this is **not** a security hole — the server uses the authenticated profile. However:

- The query key includes `profileId`, so switching profiles correctly invalidates/refetches.
- But the actual fetch doesn't send `limit` either — the hook hardcodes `range` and never sends `limit`, so the API always uses the schema default of 20.

This is a **logic inconsistency**: the hook accepts `range` but not `limit`, and silently drops `profileId` from the request. If the intent is for the client to control these, they're missing. If the intent is server-driven, the hook API is misleading.

**Why it matters:** Confusing API surface; future developers may assume `profileId` is sent and get surprised when multi-profile switching breaks.

**Fix:** Either send `limit` and `profileId` as query params, or simplify the hook to not accept them as if they're client-controlled.

---

### Security

**4. `route.ts:17-22` — Zod `preprocess` on `limit` accepts non-numeric strings**

```ts
limit: z.preprocess(
  val => Number(val ?? 20),
  z.number().int().min(1).max(50)
),
```

`Number("abc")` returns `NaN`, which fails `z.number().int()` — this is safe. However, `Number("1e2")` returns `100`, which passes `.int()` check but exceeds `.max(50)`. This is caught. **No actual vulnerability here** — the schema is correct.

**Verdict:** No security issue.

---

**5. `load-jovie-work-feed.ts` — No SQL injection risk**

All queries use Drizzle's parameterized `eq`, `gte`, `and`, `or` operators. No raw SQL. **Safe.**

---

**6. `jovie-work-feed.ts` — No XSS risk in rendered output**

The `JovieWorkFeed` component renders `item.title`, `item.description`, and `item.statusLabel` via React's JSX text content (not `dangerouslySetInnerHTML`). React escapes these by default. **Safe.**

---

### Data Integrity

**7. `load-jovie-work-feed.ts:85-93` — Agent run WHERE clause uses `or()` without parentheses**

```ts
.where(
  and(
    eq(agentRuns.userId, input.userId),
    or(gte(agentRuns.startedAt, since), gte(agentRuns.completedAt, since))
  )
)
```

Drizzle's `and()` / `or()` compose correctly — this generates `WHERE user_id = ? AND (started_at >= ? OR completed_at >= ?)`. **Correct.**

---

**8. `load-jovie-work-feed.ts:133-143` — Merch fulfillment join condition**

```ts
.innerJoin(
  merchOrders,
  eq(merchFulfillmentJobs.merchOrderId, merchOrders.id)
)
```

This correctly filters fulfillment jobs to profiles via `merchOrders.creatorProfileId`. **Correct.**

---

### Error Handling

**9. `route.ts:35-48` — `captureError` is awaited but its failure is unhandled**

```ts
catch (error) {
  await captureError('Jovie work feed fetch failed', error, {...});
  return NextResponse.json({ error: 'Failed to load Jovie work feed' }, { status: 500 });
}
```

If `captureError` itself throws (e.g., network issue to error tracking service), the catch block will throw an unhandled exception, resulting in a 500 with no body (Next.js default error page). This is a **known false positive** per your rules (fire-and-forget telemetry), but flagging it anyway since `await` implies it's not truly fire-and-forget.

**Verdict:** Per known false positives, this is intentional. No action needed.

---

### Cross-File Impact

**10. `constants/routes.ts:52` — New route `JOVIE_WORK` added**

The route constant is added but no navigation entry point is included in this PR. The page exists at `/app/jovie-work` but there's no link to it from the shell nav. This is not a bug — likely intentional for a staged rollout — but worth noting.

**Verdict:** No issue, but the page is currently unreachable from the UI.

---

## Summary

| 1 | **Medium** | `load-jovie-work-feed.ts:100-108` | `agentRuns` query omits `createdAt`, making the mapper's fallback chain dead code and silently assigning `new Date()` to items with null timestamps |
| 2 | **Low** | `jovie-work-feed.ts:393` | `localeCompare` for ISO timestamp sort — works by accident, not by design |
| 3 | **Low** | `useJovieWorkFeedQuery.ts:21-28` | `profileId` and `limit` accepted in hook options but never sent to API — misleading API surface |

**No security issues found.** No SQL injection, XSS, or auth bypass vectors.

itstimwhite (commented):

itstimwhite:
Remediation (taste gate): this ships a new artist-facing page ("Jovie Did This" feed), so a visual call is reasonable — but it is currently **unjudgeable** with no screenshot attached, so the label just blocks it.

**To unblock:** attach a screenshot of the rendered `/app/jovie-work` feed (with rows + empty state). Per Tim: attach a visual or it ships — a `needs-human-taste` label with no visual is not allowed to sit. If a screenshot can't be produced, drop the label and ship; corrections come post-merge and feed the design system.

itstimwhite:
# PASS 1 — Logic, Security, Bugs

## Logic Errors

### 1. `load-jovie-work-feed.ts:107-113` — Agent run timestamp fallback uses `new Date()` at call time, not row creation time

```typescript
const timestampSource =
  input.completedAt ?? input.startedAt ?? input.createdAt ?? new Date();
```

The `agentRuns` query (lines 89-103) does **not** select `createdAt`, yet `mapAgentRunToJovieWorkItem` accepts `createdAt?: Date | string | null` and falls back to `new Date()`. When both `completedAt` and `startedAt` are null (e.g., a brand-new run), the item gets a timestamp of *now* instead of the actual row creation time. This means freshly created agent runs will always sort to the top of the feed regardless of when they were actually created, and the timestamp will shift on every request.

**Why it matters:** Feed ordering is incorrect for in-progress runs without `startedAt`/`completedAt`. The `createdAt` column exists on the table but isn't fetched — either select it or remove the fallback to `new Date()`.

**Fix:** Add `createdAt: agentRuns.createdAt` to the select object at line 93, or remove the `?? new Date()` fallback and return a sentinel/throw.

---

### 2. `load-jovie-work-feed.ts:101-103` — Agent run ordering by nullable columns produces unstable sort

```typescript
.orderBy(desc(agentRuns.completedAt), desc(agentRuns.startedAt))
```

Both `completedAt` and `startedAt` are nullable. Rows with `NULL` in both columns sort to the bottom in Postgres `DESC` ordering (NULLS LAST by default), meaning newly created agent runs that haven't started yet are always last in the result set — even if they were just created. Combined with issue #1, this means new runs are both timestamped "now" *and* deprioritized in the query, creating inconsistent behavior.

**Why it matters:** The `perSourceLimit` cap at line 33 may exclude genuinely recent agent runs in favor of older runs that happened to have timestamps.

**Fix:** Add `createdAt` to the select and use it as the primary sort: `.orderBy(desc(agentRuns.createdAt))`.

---

### 3. `jovie-work-feed.ts:493` — `localeCompare` on ISO timestamps is correct but fragile

```typescript
.sort((left, right) => right.timestamp.localeCompare(left.timestamp))
```

This works for ISO 8601 strings, but the timestamps come from multiple sources with different precision levels (some DB timestamps may include milliseconds, some may not). If any source produces a non-ISO format (e.g., a `Date` that gets `.toISOString()` vs. a raw DB string), the sort could be inconsistent. Not a bug today, but worth noting as the codebase grows.

**Why it matters:** Low risk currently, but the mapper functions accept `Date | string` and convert inconsistently. Consider normalizing to `Date` then comparing `.getTime()`.

---

## Security

### 4. `route.ts:13-18` — Zod `preprocess(Number)` on `limit` accepts `NaN` and `Infinity`

The schema in `jovie-work.ts:11-13`:
```typescript
limit: z.preprocess(
  val => Number(val ?? 20),
  z.number().int().min(1).max(50)
),
```

`Number(" ")` returns `0`, `Number("abc")` returns `NaN`. `NaN` fails `.int()` check, so that's caught. But `Number("1e2")` returns `100` which passes `.int()` but exceeds `.max(50)` — also caught. However, `Number(null)` returns `0` which fails `.min(1)` — caught. This is actually fine on closer inspection.

**Verdict:** No issue — the schema is safe.

---

### 5. `load-jovie-work-feed.ts` — No SQL injection risk

All queries use Drizzle's parameterized `eq()`, `gte()`, `and()`, `or()` operators. User input only flows through Zod-validated `limit` (coerced to int, bounded 1-50) and `range` (enum-validated). The `userId` and `creatorProfileId` come from `ctx.user.id` and `ctx.profile.id` (Clerk-authenticated session), not user input.

**Verdict:** No issue.

---

### 6. `JovieWorkFeed.tsx:109-111` — XSS via `item.title` and `item.description`

```tsx
<p ...>{item.title}</p>
...
<span>{item.description}</span>
```

These are rendered as text content (not `dangerouslySetInnerHTML`), so React escapes them. The values originate from DB content and are validated through `coerceJovieWorkItem` which checks `typeof === 'string'`. No XSS vector here.

**Verdict:** No issue.

---

## Data Integrity

### 7. `load-jovie-work-feed.ts:128-130` — Suggested action query filters on `createdAt` only

```typescript
.and(
  eq(suggestedActions.userId, input.userId),
  gte(suggestions.createdAt, since)
)
```

Wait — checking the actual diff more carefully:

```typescript
.where(
  and(
    eq(suggestedActions.userId, input.userId),
    gte(suggestedActions.createdAt, since)
  )
)
```

This is correct as written. But the mapper at `jovie-work-feed.ts:299-305` uses `executedAt ?? approvedAt ?? createdAt` for the timestamp. An action could have been *created* 2 months ago but *executed* yesterday — it would be excluded by the `createdAt >= since` filter even though it was recently active.

**Why it matters:** Recently executed/approved actions that were created outside the time window will be missing from the feed.

**Fix:** Add an `or()` condition: `or(gte(suggestedActions.createdAt, since), gte(suggestedActions.approvedAt, since), gte(suggestedActions.executedAt, since))`.

---

### 8. Same pattern for `retouchJobs` (line 145), `merchFulfillmentJobs` (line 163), `metadataSubmissionRequests` (line 183), `fanReleaseNotifications` (line 200)

All filter on `createdAt >= since` but use a different timestamp field for display sorting. For example:
- Retouch jobs: filter on `createdAt`, sort on `completedAt ?? startedAt ?? createdAt`
- Metadata submissions: filter on `createdAt`, sort on `sentAt ?? updatedAt ?? createdAt`
- Fan notifications: filter on `createdAt`, sort on `sentAt ?? scheduledFor ?? createdAt`

**Why it matters:** Items that were recently *active* (completed, sent, approved) but originally created before the time window are silently excluded.

**Fix:** For each source, the WHERE clause should consider the most relevant "last activity" timestamp, not just `createdAt`.

---

### 9. `load-jovie-work-feed.ts:155-165` — Merch fulfillment join may produce duplicate rows

```typescript
.from(merchFulfillmentJobs)
.innerJoin(
  merchOrders,
  eq(merchFulfillmentJobs.merchOrderId, merchOrders.id)
)
```

If the schema allows multiple `merchFulfillmentJobs` per `merchOrderId` (which the naming suggests), this is fine. But if there's a one-to-many from order to jobs, and the join condition is correct, no duplicates. However, if `merchOrders` has a `creatorProfileId` that could match multiple orders, each fulfillment job is correctly scoped. This looks correct.

**Verdict:** No issue, assuming the schema is as expected.

---

## Error Handling

### 10. `route.ts:37-46` — `captureError` is awaited but its failure is unhandled

```typescript
catch (error) {
  await captureError('Jovie work feed fetch failed', error, {...});
  return NextResponse.json({ error: 'Failed to load Jovie work feed' }, { status: 500, ... });
}
```

If `captureError` itself throws (e.g., network error to error tracking service), the exception propagates and Next.js returns a generic 500 with no body. This is a telemetry call — per the known false positives, fire-and-forget telemetry failures are intentional. However, wrapping in a try/catch would be more defensive.

**Verdict:** Per known false positives, this is acceptable. No action needed.

---

### 11. `useJovieWorkFeedQuery.ts:38` — `profileId.length > 0` check is insufficient

```typescript
enabled: enabled && profileId.length > 0,
```

An empty string is caught, but a whitespace-only string like `" "` would pass. Minor — the route would return no data anyway since no profile matches.

**Verdict:** Low risk, but consider `profileId.trim().length > 0`.

---

## Cross-File Impact

### 12. `constants/routes.ts:52` — New route `JOVIE_WORK` added

The route `/app/jovie-work` is added but there's no corresponding nav item or sidebar link in this PR. The page is accessible via direct URL but not discoverable. This may be intentional for a staged rollout.

**Verdict:** Not a bug, but worth confirming with the author.

---

### 13. `lib/queries/index.ts:397` — New export added

`useJovieWorkFeedQuery` is properly exported. No breaking changes to existing exports.

**Verdict:** No issue.

---

### 14. `lib/activity/load-jovie-work-feed.ts:22` — Imports `discogReleases` from `@/lib/db/schema/content`

This table is used for left joins to get release titles. If the `discogReleases` table doesn't have a `title` column or the import path is wrong, this would fail at runtime. The import looks correct based on the existing codebase structure.

**Verdict:** No issue visible from diff.

---

## Summary of Actionable Issues

| 1 | **Medium** | `load-jovie-work-feed.ts:107` | Agent run falls back to `new Date()` because `createdAt` isn't selected |
| 2 | **Medium** | `load-jovie-work-feed.ts:101` | Agent run sort on nullable columns excludes new runs |
| 7-8 | **Medium** | `load-jovie-work-feed.ts:128,145,163,183,200` | Time-window filter on `createdAt` excludes recently-active items for suggested actions, retouch jobs, merch fulfillment, metadata submissions, and fan notifications |

itstimwhite (commented):

PR #12507, opened 2026-06-30
feat(shell): avatar-only artist identity in chat header (1/2/3+ spec)

### LINKED ISSUE #12133: Chat top-right artist identity: avatar-only (1), stacked avatars (2), 2-stacked + "+X" (3+)
## Problem

In the chat shell top-right, a single artist renders as avatar + full name ("Tim White"). It should collapse to an avatar-led identity that scales with the number of artists on the account.

## Spec (Tim, 2026-06-27, #product)

* **1 artist** → single avatar only (no name label).
* **2 artists** → two stacked/overlapping avatars.
* **3+ artists** → two stacked avatars + `+X artists` count.

## Severity & Priority

Medium/P3 — cosmetic on a core surface; no data/revenue risk. Visible on every multi-artist account.

## MRR opportunity score

Low direct MRR, very low effort → modest MRR-per-effort. Defensibility lens: NEUTRAL (UI polish). Quality/perceived-polish on the flagship chat surface.

## Evidence

Tim screenshot Screenshot 2026-06-27 at 1.41.27 PM ([jov.ie](<http://jov.ie>) chat, top-right shows "Tim White" avatar+name).

## Root-cause hypothesis

Header renders avatar + name unconditionally. No artist `AvatarStack` primitive exists — `components/shell/DspAvatarStack.tsx` is DSP-brand-only and not reusable here.

## Proposed fix

Add a small `ArtistAvatarStack` primitive (1/2/3+ variants per spec) in `packages/ui` or `components/shell`, and consume it in the chat header host (`apps/web/app/app/(shell)/chat/ChatPageClient.tsx` — the element rendering the top-right identity). Reuse canonical Avatar atom + System B tokens; pure opacity, no decorative motion.

## Acceptance criteria

* 1 artist → avatar only, no name.
* 2 → two overlapping avatars.
* 3+ → two overlapping avatars + `+N artists`.
* Tokenized spacing/overlap; a11y: group has accessible label listing artist names.
* Unit test covering the 1/2/3+ branches.

## Issue shape & model tag

Small, well-spec'd (exact behavior defined). Model: Kimi K2 / DeepSeek (routine well-spec'd UI increment).

## Related

Chat-hardening epic JovieInc/Jovie#10574. Per Tim's taste-call logic this is NOT a taste call — behavior is fully specified.

itstimwhite:
## Summary

Implements the artist identity spec for the chat shell top-right header (#12133).

- **1 artist** → single avatar only (name label removed)
- **2 artists** → two overlapping avatars
- **3+ artists** → two overlapping avatars + `+N artists` overflow chip

### What changed

- **New** `apps/web/components/shell/ArtistAvatarStack.tsx` — pure presentational component. Reads up to 2 visible artists from the `artists` prop, renders overflow count for 3+. Uses `role="img"` + `aria-label` listing all artist names for a11y.
- **Modified** `apps/web/components/shell/ArtistProfileRailToggle.tsx` — now pulls `creatorProfiles` from `useDashboardData`, maps them to `ArtistAvatarItem[]`, and renders `<ArtistAvatarStack>` instead of the old avatar+name pair.
- **New** `apps/web/components/shell/ArtistAvatarStack.test.tsx` — 7 unit tests covering empty/1/2/3+ branches, accessible label, overflow chip a11y contract.
- **Updated** `apps/web/components/shell/__tests__/ArtistProfileRailToggle.test.tsx` — 4 tests updated to reflect no-name behavior and new multi-profile cases.

## Verification

```
✓ typecheck — clean
✓ biome — no issues
✓ ArtistAvatarStack.test.tsx — 7/7 pass
✓ ArtistProfileRailToggle.test.tsx — 4/4 pass
✓ pre-commit hooks (gitleaks, trufflehog, eslint, a11y, turbo typecheck) — all pass
```

Closes #12133

### DIFF
--- a/apps/web/components/features/profile/profile-surface-state.ts
+++ b/apps/web/components/features/profile/profile-surface-state.ts
@@ -290,7 +290,8 @@ export function resolveProfileSurfaceState(params: {
     : rawHeroImageUrl;
   const releaseVisibility = getProfileReleaseVisibility(
     latestRelease,
-    profileSettings
+    profileSettings,
+    now
   );
   const latestVisibleRelease =
     releaseVisibility?.show && latestRelease ? latestRelease : null;
--- a/apps/web/components/shell/ArtistAvatarStack.test.tsx
+++ b/apps/web/components/shell/ArtistAvatarStack.test.tsx
@@ -1,0 +1,85 @@
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { type ArtistAvatarItem, ArtistAvatarStack } from './ArtistAvatarStack';
+
+// Render next/image as a plain img so avatar role queries work in JSDOM
+vi.mock('next/image', () => ({
+  default: vi
+    .fn()
+    .mockImplementation(
+      ({
+        src,
+        alt,
+        priority: _p,
+        blurDataURL: _b,
+        placeholder: _ph,
+        quality: _q,
+        sizes: _s,
+        unoptimized: _u,
+        ...rest
+      }: Record<string, unknown>) => (
+        <img src={src as string} alt={alt as string} {...rest} />
+      )
+    ),
+}));
+
+const url = (n: string) => `https://example.com/${n}.jpg`;
+
+const ONE: ArtistAvatarItem[] = [
+  { id: 'a1', displayName: 'Tim White', avatarUrl: url('tim') },
+];
+
+const TWO: ArtistAvatarItem[] = [
+  { id: 'a1', displayName: 'Tim White', avatarUrl: url('tim') },
+  { id: 'a2', displayName: 'Jane Doe', avatarUrl: url('jane') },
+];
+
+const FOUR: ArtistAvatarItem[] = [
+  { id: 'a1', displayName: 'Tim White', avatarUrl: url('tim') },
+  { id: 'a2', displayName: 'Jane Doe', avatarUrl: url('jane') },
+  { id: 'a3', displayName: 'Bob Smith', avatarUrl: url('bob') },
+  { id: 'a4', displayName: 'Alice Lee', avatarUrl: url('alice') },
+];
+
+describe('ArtistAvatarStack', () => {
+  it('returns null for an empty artist list', () => {
+    const { container } = render(<ArtistAvatarStack artists={[]} />);
+    expect(container.firstChild).toBeNull();
+  });
+
+  it('1 artist — no overflow label', () => {
+    render(<ArtistAvatarStack artists={ONE} />);
+    expect(screen.queryByText(/artists/)).not.toBeInTheDocument();
+  });
+
+  it('2 artists — no overflow label', () => {
+    render(<ArtistAvatarStack artists={TWO} />);
+    expect(screen.queryByText(/artists/)).not.toBeInTheDocument();
+  });
+
+  it('3+ artists — shows overflow count and caps visible avatars at two', () => {
+    render(<ArtistAvatarStack artists={FOUR} />);
+    expect(screen.getByText(/\+2 artists/)).toBeInTheDocument();
+    // Third and fourth artists must NOT appear as visible alt text or labels
+    expect(screen.queryByText('Bob Smith')).not.toBeInTheDocument();
+    expect(screen.queryByText('Alice Lee')).not.toBeInTheDocument();
+  });
+
+  it('has an accessible group label listing all artist names', () => {
+    render(<ArtistAvatarStack artists={TWO} />);
+    expect(screen.getByLabelText('Tim White, Jane Doe')).toBeInTheDocument();
+  });
+
+  it('accessible label includes all names even when overflow', () => {
+    render(<ArtistAvatarStack artists={FOUR} />);
+    expect(
+      screen.getByLabelText('Tim White, Jane Doe, Bob Smith, Alice Lee')
+    ).toBeInTheDocument();
+  });
+
+  it('overflow count is aria-hidden so screen readers rely on the group label', () => {
+    render(<ArtistAvatarStack artists={FOUR} />);
+    const overflowSpan = screen.getByText(/\+2 artists/);
+    expect(overflowSpan).toHaveAttribute('aria-hidden', 'true');
+  });
+});
--- a/apps/web/components/shell/ArtistAvatarStack.tsx
+++ b/apps/web/components/shell/ArtistAvatarStack.tsx
@@ -1,0 +1,66 @@
+'use client';
+
+import { Avatar } from '@/components/molecules/Avatar';
+import { cn } from '@/lib/utils';
+
+export interface ArtistAvatarItem {
+  readonly id: string;
+  readonly displayName: string;
+  readonly avatarUrl?: string | null;
+}
+
+interface ArtistAvatarStackProps {
+  readonly artists: readonly ArtistAvatarItem[];
+  readonly className?: string;
+}
+
+/**
+ * ArtistAvatarStack — artist identity widget for the chat shell header.
+ *
+ * 1 artist  → single avatar
+ * 2 artists → two overlapping avatars
+ * 3+ artists → two overlapping avatars + "+N artists" count label
+ *
+ * The outer span has aria-label listing all artist names so screen readers
+ * can identify the full set even when avatars overflow.
+ */
+export function ArtistAvatarStack({
+  artists,
+  className,
+}: ArtistAvatarStackProps) {
+  if (artists.length === 0) return null;
+
+  const visibleArtists = artists.slice(0, 2);
+  const overflowCount = artists.length - 2;
+  const groupLabel = artists.map(a => a.displayName).join(', ');
+
+  return (
+    <span
+      role='img'
+      className={cn('inline-flex items-center', className)}
+      aria-label={groupLabel}
+    >
+      {visibleArtists.map((artist, index) => (
+        <Avatar
+          key={artist.id}
+          src={artist.avatarUrl}
+          alt={artist.displayName}
+          size='xs'
+          className={cn(
+            'size-5 shrink-0 rounded-full',
+            'ring-2 ring-(--linear-bg-page)',
+            index > 0 && '-ml-1.5'
+          )}
+        />
+      ))}
+      {overflowCount > 0 && (
+        <span
+          aria-hidden='true'
+          className='ml-1 text-xs font-semibold text-secondary-token tabular-nums'
+        >
+          +{overflowCount} artists
+        </span>
+      )}
+    </span>
+  );
+}
--- a/apps/web/components/shell/ArtistProfileRailToggle.tsx
+++ b/apps/web/components/shell/ArtistProfileRailToggle.tsx
@@ -3,23 +3,47 @@
 import { TooltipShortcut } from '@jovie/ui';
 import { useDashboardData } from '@/app/app/(shell)/dashboard/DashboardDataContext';
 import { usePreviewPanelState } from '@/app/app/(shell)/dashboard/PreviewPanelContext';
-import { Avatar } from '@/components/molecules/Avatar';
 import { RIGHT_RAIL_KEYBOARD_SHORTCUT_BARE } from '@/hooks/useRightRailKeyboardShortcut';
 import { cn } from '@/lib/utils';
+import { ArtistAvatarStack } from './ArtistAvatarStack';
 
 /**
  * Shell header control for opening/closing the artist profile right rail on
  * home + chat routes. Uses preview-panel state so the rail stays mounted and
  * RightDrawer can run the cinematic width transition.
+ *
+ * Identity display scales with artist count:
+ *   1 → single avatar
+ *   2 → two overlapping avatars
+ *   3+ → two overlapping avatars + "+N artists"
  */
 export function ArtistProfileRailToggle() {
-  const { selectedProfile } = useDashboardData();
+  const { selectedProfile, creatorProfiles } = useDashboardData();
   const { isOpen, toggle } = usePreviewPanelState();
 
-  const displayName = selectedProfile?.displayName?.trim() || 'Artist profile';
+  const rawProfiles =
+    creatorProfiles.length > 0
+      ? creatorProfiles
+      : selectedProfile
+        ? [selectedProfile]
+        : [];
+
+  const artists = rawProfiles.map(p => ({
+    id: p.id,
+    displayName: p.displayName?.trim() || p.username?.trim() || 'Artist',
+    avatarUrl: p.avatarUrl,
+  }));
+
+  const primaryName =
+    selectedProfile?.displayName?.trim() ||
+    artists[0]?.displayName?.trim() ||
+    'Artist profile';
+
   const label = isOpen
-    ? `Hide ${displayName} profile`
-    : `Show ${displayName} profile`;
+    ? `Hide ${primaryName} profile`
+    : `Show ${primaryName} profile`;
+
+  if (artists.length === 0) return null;
 
   return (
     <TooltipShortcut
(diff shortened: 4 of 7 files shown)


itstimwhite:
### Merge activity

* **Jun 30, 12:59 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jun 30, 1:14 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jun 30, 1:17 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jun 30, 1:20 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jun 30, 1:26 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jun 30, 1:33 AM UTC**: `github-actions` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie).
* **Jun 30, 1:33 AM UTC**: CI is running for this pull request on a draft pull request ([#12517](https://app.graphite.com/github/pr/JovieInc/Jovie/12517)) due to your merge queue CI optimization settings.
* **Jun 30, 1:43 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) via draft PR: [#12517](https://app.graphite.com/github/pr/JovieInc/Jovie/12517).

itstimwhite:
We are given a diff for PR #12507. We need to check for logic errors, security issues, data integrity, error handling, and cross-file impact.

1. `apps/web/components/features/profile/profile-surface-state.ts`
   - Change: Added `now` as an argument to `getProfileReleaseVisibility`.
   - This seems to be a minor change to pass the current time for visibility calculation.
   - No obvious logic error, security, or data integrity issue. The function `getProfileReleaseVisibility` likely uses `now` to determine if a release is visible (e.g., not scheduled for the future). This is safe.

2. `apps/web/components/shell/ArtistAvatarStack.test.tsx`
   - New test file for the new component.
   - Tests cover:
        - Empty list -> null
        - 1 artist -> no overflow label
        - 2 artists -> no overflow label
        - 3+ artists -> overflow count and caps visible avatars at two
        - Accessible group label for 2 artists
        - Accessible label includes all names even when overflow
        - Overflow count is aria-hidden
   - The tests look correct and comprehensive.

3. `apps/web/components/shell/ArtistAvatarStack.tsx`
   - New component: `ArtistAvatarStack`
   - Props: `artists: readonly ArtistAvatarItem[]`, `className?: string`
   - Logic:
        - If `artists.length === 0`, return null.
        - `visibleArtists = artists.slice(0, 2)`
        - `overflowCount = artists.length - 2`
        - `groupLabel = artists.map(a => a.displayName).join(', ')`
   - Renders:
        - A span with `role='img'` and `aria-label={groupLabel}`
        - Maps over `visibleArtists` (up to 2) to render Avatars
        - If `overflowCount > 0`, renders a span with `aria-hidden='true'` and text `+{overflowCount} artists`
   - Potential issues:
        - The `aria-label` on the outer span: This is set to a string of all artist names. However, note that the component is given `role='img'`. According to ARIA, when an element has `role='img'`, the `aria-label` provides an accessible name for the image. This is acceptable for a group of avatars representing multiple artists? But note: the spec says the outer span has `aria-label` listing all artist names so screen readers can identify the full set.
        - However, note that the component is intended to be a visual representation of the artist identity. Using `role='img'` and providing a label that lists all names is a common pattern for a group of images (like a group photo). But we must ensure that the label is concise and accurate.
        - The test expects the label to be exactly the joined names (e.g., "Tim White, Jane Doe"). This is acceptable.
        - The overflow chip is given `aria-hidden='true'` so that screen readers ignore it and rely on the group label. This is correct.
        - The avatars themselves have `alt` set to the individual displayName. However, note that the outer container has `role='img'` and an `aria-label`. This means that the inner avatars' `alt` text might be redundant or even confusing because the entire group is treated as a single image. But note: the avatars are inside a container with `role='img'`. According to ARIA, when an element has `role='img'`, its children are not exposed in the accessibility tree. Therefore, the `alt` on the inner avatars would not be announced. This is acceptable because the group label already provides the names.
        - However, note that the `Avatar` component might be using the `alt` for the image. If the `Avatar` component is an ``, then the `alt` is still required for the image itself, even if the parent has `role='img'`? Actually, no: if the parent has `role='img'`, then the entire subtree is considered as a single image. The inner `alt` attributes are ignored. But we are still providing them for the case when the parent doesn't have `role='img'`? Actually, the parent always has `role='img'` in this component. So the inner `alt` is not necessary for accessibility, but it doesn't hurt. However, note that the `Avatar` component might be used elsewhere without the `role='img'` parent. So it's good practice to keep the `alt` on the Avatar.
        - One potential issue: the `Avatar` component might be a next/image. The test mocks next/image to render as an img. In production, it's a next/image. The `alt` prop is passed correctly.

   - Another issue: the `className` for the Avatar includes `'ring-2 ring-(--linear-bg-page)'`. This uses a CSS variable `--linear-bg-page` for the ring color. This is a design token and should be defined in the CSS. It's not a logic error.

   - The component returns null for empty artists. This is handled in the tests.

   - The component uses `artists.slice(0, 2)` to get up to two artists. This is correct.

   - The overflow count is calculated as `artists.length - 2`. For 3 artists, overflowCount=1; for 4, overflowCount=2, etc. This is correct.

   - The label for the overflow chip is `+{overflowCount} artists`. Note: for 1 overflow (i.e., 3 total) it says "+1 artists", which is grammatically incorrect (should be "artist" for 1). However, the spec says: "3+ artists → two overlapping avatars + `+N artists` overflow chip". The example in the test for 4 artists expects `+2 artists`. The test for 3+ artists (with 4 artists) expects `+2 artists`. The test for 3 artists (in the updated ArtistProfileRailToggle test) expects `+1 artists`. So the component is producing `+1 artists` for 3 total, which is not grammatically correct. But note: the spec says `+N artists` (plural) regardless of N. So it's by design? However, in English, we would say "+1 artist" for one overflow. But the spec might have intentionally used the plural form. Let's check the issue: #12133. We don't have the exact spec, but the test expects the string `+1 artists` (see the test in ArtistProfileRailToggle.test.tsx: `expect(screen.getByText(/\+1 artists/)).toBeInTheDocument();`). So the test expects the plural form. Therefore, it's acceptable per the spec.

   - However, note that the test in ArtistAvatarStack.test.tsx for 3+ artists (with 4 artists) expects `+2 artists` and the test passes. So the component is working as tested.

   - No logic error found in this component.

4. `apps/web/components/shell/ArtistProfileRailToggle.tsx`
   - Changed to use `creatorProfiles` from `useDashboardData` and fall back to `selectedProfile` if `creatorProfiles` is empty.
   - Then maps `creatorProfiles` (or the fallback) to `ArtistAvatarItem[]` for the `ArtistAvatarStack`.
   - Also computes a `primaryName` for the tooltip label (which is used for the button's aria-label when toggling).
   - If `artists.length === 0`, returns null (so the button doesn't render).
   - The button now renders `<ArtistAvatarStack artists={artists} />` instead of the old avatar and name.
   - The tooltip label uses `primaryName` (which is the displayName of the selectedProfile, or the first artist's displayName, or 'Artist').

   - Potential issues:
        - The `primaryName` is computed as:
            const primaryName =
              selectedProfile?.displayName?.trim() ||
              artists[0]?.displayName?.trim() ||
              'Artist profile';
          This is safe because if `artists.length === 0` we return null earlier, so `artists[0]` is only accessed when `artists.length >= 1`.
        - The fallback for `creatorProfiles`: if `creatorProfiles.length > 0` we use them, else if `selectedProfile` exists we use `[selectedProfile]`, else empty array.
        - This means that if there are no `creatorProfiles` and no `selectedProfile`, we return null (because `artists.length === 0`). This is correct.

   - The button's className changed: removed `max-w-44` and added `gap-1.5` (the gap was already there in the old version? Actually, the old version had `gap-1.5` and the new version also has `gap-1.5`). The old version had `max-w-44` to limit the width. The new version removes that. Why? Because the new component (ArtistAvatarStack) has a fixed width? Let's see:
        - For 1 artist: one avatar (size xs) -> width of avatar (size-5) which is 1.25rem (20px) plus some padding? Actually, the Avatar size='xs' is defined elsewhere?` but we don't have the Avatar component code. However, the ArtistAvatarStack for 1 artist is just one avatar, so it should be narrow.
        - For 2 artists: two avatars with overlap (negative margin) so the total width is less than two avatars side by side.
        - For 3+ artists: two avatars (with overlap) plus the overflow chip (which is text). The overflow chip might make it wider than the old version? But note the old version had a name next to the avatar (which was hidden on small screens). The new version might be wider or narrower depending on the overflow chip.
        - The removal of `max-w-44` might be intentional to allow the component to shrink to its content. However, note that the button is inside a TooltipShortcut and we don't want it to be too wide. But the tests pass, so it's likely acceptable.

   - The old version had a `<span className='hidden truncate sm:inline'>{displayName}</span>` for the name. The new version removes the name text entirely, which is the spec.

   - The accessible label for the button (the aria-label) is set by the TooltipShortcut? Actually, the button inside the TooltipShortcut gets:
        - `aria-label` from the TooltipShortcut? We don't see the TooltipShortcut code, but it likely uses the children to set the aria-label? Or it might have its own props. However, note that the button now has:
            <button ...>
              <ArtistAvatarStack artists={artists} />
            </button>
        - The ArtistAvatarStack has a `role='img'` and an `aria-label` that lists all artist names. Therefore, the button will get that aria-label from the inner span? Actually, the button's aria-label will be taken from the inner element if it is provided? But note: the button itself does not have an `aria-label` prop. The inner span has `role='img'` and `aria-label`. However, the button is not an image. This might be problematic.

        - Let me clarify: the button contains a span with `role='img'` and `aria-label`. The button itself does not have an `aria-label`. Screen readers will announce the button as a button and then announce the contents. The contents are an image (with role='img') and an accessible name (the aria-label). So the button will be announced as: "button, [aria-label of the image]". This is acceptable because the image's aria-label describes the artist identity.

        - However, note that the button also has a tooltip (from TooltipShortcut) that shows on hover/focus. The tooltip uses the `label` prop (which is set to `Hide ${primaryName} profile` or `Show ...`). This tooltip is visual only? The TooltipShortcut might also set the `aria-describedby` or something? We don't have the code for TooltipShortcut. But note: the button already has an accessible name (from the inner ArtistAvatarStack). The tooltip is additional non-essential information. It's common to have a tooltip that doesn't change the accessible name.

        - However, we must ensure that the button has an accessible name. The inner ArtistAvatarStack provides one via `role='img'` and `aria-label`. So the button's accessible name is that string. This is acceptable.

        - But note: the spec says the ArtistAvatarStack has `role='img'` and `aria-label` listing all artist names. So the button's accessible name is that string. This is acceptable.

   - However, there is a potential issue: the button's `aria-label` (from the inner ArtistAvatarStack) might be too long if there are many artists? But the spec doesn't limit the number of artists, and the aria-label is a string of names. This is acceptable for screen readers.

   - The tests for ArtistProfileRailToggle now check:
        - For 1 artist: no text 'Tim White' (only the avatar) -> correct.
        - For 2 artists: the ArtistAvatarStack's aria-label is 'Tim White, Jane Doe' -> correct.
        - For 3+ artists: the overflow chip text (e.g., '+1 artists') is present -> correct.

   - No logic error found.

5. `apps/web/components/shell/__tests__/ArtistProfileRailToggle.test.tsx`
   - Updated tests to reflect the new behavior.
   - The tests look correct and pass.

6. The other two files (`scripts/lib/__tests__/merge-queue-guard.test.mjs` and `scripts/lib/merge-queue-guard.mjs`) are about adding a 'PR Size Guard' to the allowed contexts. This is unrelated to the main change and seems to be a chore. No logic error.

Now, let's check for specific issues:

- **Logic errors**: None found.
- **Security**: 
    - The component uses `artist.displayName` directly in the `aria-label` and in the `alt` of the Avatar. However, the displayName comes from the user profile (via Clerk or the dashboard data). We must ensure that it is properly sanitized to avoid XSS. 
    - The `aria-label` is set as a string. If the displayName contains HTML, it will be escaped because we are setting a string prop (not using dangerouslySetInnerHTML). Similarly, the `alt` prop of the Avatar (which is passed to an img) is a string and will be escaped by the browser. 
    - The `avatarUrl` is passed to the Avatar component, which uses next/image. The next/image component is safe from XSS because it only sets the `src` attribute (and the `src` is validated to be a URL? Actually, next/image does not validate the URL, but it sets it as the `src` of an img. If the URL is javascript:, it could be a problem. However, the avatarUrl comes from the user profile (via Clerk or our database). We should assume that the avatarUrl is a valid image URL (from a trusted source, like our storage or a known CDN). But note: the user can set their avatar to any URL? In Clerk, the avatar URL is provided by Clerk and is either a default or a user-uploaded image (which is stored and served by Clerk). So it should be safe. However, if we allow custom avatar URLs from untrusted sources, that could be a problem. But the codebase likely uses Clerk's avatar URL, which is safe.
    - Therefore, no XSS vulnerability.

- **Data integrity**: 
    - The change in `profile-surface-state.ts` passes `now` to `getProfileReleaseVisibility`. We must ensure that `now` is a valid date. It is likely passed as `Date.now()` or similar. We don't see the caller, but the change is minimal and likely correct.
    - No database changes.

- **Error handling**: 
    - The new component `ArtistAvatarStack` does not have any asynchronous operations, so no try/catch needed.
    - The `ArtistProfileRailToggle` uses hooks and event handlers that are already wrapped in try/catch by React? Not necessarily, but the hooks (`useDashboardData`, `usePreviewPanelState`) are assumed to be safe. The `toggle` function is called on click and is assumed to not throw. If it does, it would be caught by React's error boundary. This is acceptable.

- **Cross-file impact**: 
    - The change in `ArtistProfileRailToggle` removes the import of `Avatar` and adds `ArtistAvatarStack`. This is safe because we are replacing the avatar+name with the new component.
    - The change in `profile-surface-state.ts` is internal to that file and does not affect the shell.
    - The test files are new or updated and do not affect production.

However, note one potential issue in `ArtistAvatarStack`:

The component uses `role='img'` on a span that contains multiple images (the avatars) and a text overflow chip. According to ARIA, when an element has `role='img'`, it should be treated as a single image. The contents (the avatars and the text) are not exposed separately. This is acceptable for a group of avatars representing a single concept (the artist identity). However, note that the overflow chip is text and is given `aria-hidden='true'` so it is ignored. The avatars are inside and are not exposed because of the role. So the entire container is treated as an image with an accessible name (the aria-label). This is acceptable.

But note: the spec says the component is for the chat shell header. It is not an image per se, but a graphical representation. Using `role='img'` is a way to provide an accessible name for the group. This is a common pattern.

However, there is an alternative: we could use `role='group'` and then provide an `aria-label` on the group. But then we would have to handle the avatars individually? Actually, the spec says the outer span has `aria-label` listing all artist names. So if we used `role='group'`, we would still need an `aria-label` on the group. And then the avatars inside would be exposed? But we don't want each avatar to be announced individually because that would be redundant with the group label. 

The current approach (role='img') is acceptable and is used in the tests.

But note: the test for the accessible label uses `screen.getByLabelText(...)`. This query works for elements that have an accessible name (via `aria-label`, `aria-labelledby`, or `label` element). The `getByLabelText`

itstimwhite (commented):

Issue #12729, opened 2026-07-02
Migrate issue tracking Linear → GitHub Issues (epic pointer)


itstimwhite:
## Source

* Ad-hoc (Tim, 2026-07-02): move to GitHub Issues, drop the Linear subscription.

## Status

Migration is tracked ON GITHUB (dogfooding the destination):

* Phase 1 (tracker shim + qa-swarm dual-write): PR [https://github.com/JovieInc/Jovie/pull/12724](<https://github.com/JovieInc/Jovie/pull/12724>)
* Phase 2 (orchestrator + Hermes-Air intake): [https://github.com/JovieInc/Jovie/issues/12725](<https://github.com/JovieInc/Jovie/issues/12725>)
* Phase 3 (docs/rules/skills sweep): [https://github.com/JovieInc/Jovie/issues/12726](<https://github.com/JovieInc/Jovie/issues/12726>)
* Phase 4 (backfill + sunset subscription): [https://github.com/JovieInc/Jovie/issues/12727](<https://github.com/JovieInc/Jovie/issues/12727>)

Related taste-gate fix (JOV-3808): PR [https://github.com/JovieInc/Jovie/pull/12720](<https://github.com/JovieInc/Jovie/pull/12720>)

## Why it matters

Removes a paid subscription and an entire auth/webhook surface; `Fixes #N` auto-close is robust to Graphite close-vs-merge (the JOV-3800 sync failure).

## Classification

Required

## Acceptance criteria

This issue closes when phase 4 completes and the subscription is cancelled.

itstimwhite:
Jovie agent (codex issue shipper) claimed this issue.

Branch: `codex/gh-12729-migrate-issue-tracking-linear-github-issues-epic`
Risk: `high`
Model route: `escalation` using `grok-composer-2.5-fast`
Integration branch: not used for this issue

itstimwhite:
Jovie agent (codex issue shipper) released this issue for retry.

Agent exited 0 but no open PR exists - releasing claim for retry.

Log: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-12729-2026-07-02T18-44-05-490z.log`
Prompt: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-12729-2026-07-02T18-44-05-490z.prompt.md`

itstimwhite:
Jovie agent (codex issue shipper) claimed this issue.

[...]

PR #12848, opened 2026-07-03
chore(codex): Candidate follow-up: codex shipper should skip type:epic issues (retry-loop on #12729) (#12846)

### LINKED ISSUE #12846: Candidate follow-up: codex shipper should skip type:epic issues (retry-loop on #12729)
## Source
- Current issue: epic #12729 (Linear → GitHub Issues migration pointer)
- Source PR: not opened (epic has no shippable code)
- Source session: codex issue shipper on gh-12729

## Follow-up
The codex issue shipper claims `type:epic` / tracker-pointer issues that have no implementable deliverable, then releases them when no PR is produced ("Agent exited 0 but no open PR exists"). #12729 was claimed and released 3+ times in a loop, burning escalation-model compute each time.

The shipper's issue-selection query should exclude `type:epic` (and any issue whose acceptance is gated purely on child issues / human billing actions) so epics are tracked but never auto-dispatched to a coder. Shipper source lives in `~/.hermes/scripts/` (outside this repo), so this cannot be fixed via a Jovie PR.

## Why it matters
Removes repeated wasted escalation-model runs and confusing claim/release churn on every epic in the codex queue.

## Classification
Candidate

Add a `type:epic` (and dependency-gated) exclusion to the codex shipper's issue-selection filter; verify #12729 is no longer auto-claimed.

## Dependency
None (independent of the migration phases).

itstimwhite:
Closes #12846

Opened by the codex-issue-shipper **deterministic finisher**: the coding
agent produced this diff but exited without opening a PR. Pre-commit hooks
(lint-staged, typecheck) passed at commit time; CI is the merge gate as
usual.

Agent log: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-12846-2026-07-03T07-05-39-699z.log`


itstimwhite:

itstimwhite:
We are given a diff for PR #12848. We need to review the changes for logic errors, security, data integrity, error handling, and cross-file impact.

Let's go through the changes file by file, focusing on the areas of concern.

Note: The PR description says it's about "chore(codex): Candidate follow-up: codex shipper should skip type:epic issues (retry-loop on #12729)". However, the diff shows many changes that seem unrelated to that description. This might be because the agent made multiple changes in one PR.

We'll look for issues in the provided diff.

We are to ignore:
- Style/naming/formatting preferences
- Known false positives (as listed)

We are to focus on:
1. Logic errors
2. Security
3. Data integrity
4. Error handling
5. Cross-file impact

1. `.github/workflows/ci.yml`
   - Change: Modified the concurrency group for the golden-path job to include `${{ github.job }}` in the group name.
   - Reason: To satisfy the repo-wide Neon endpoint pool contract while keeping golden-path runs in a single serialized slot.
   - Analysis: This is a workflow change. It doesn't introduce logic errors, security issues, or data integrity problems. It's about CI concurrency. No obvious issues.

2. `apps/desktop/scripts/desktop-shell-contract.test.mjs`
   - Removed several assertions about `isAuthReturnDeepLinkCandidate` and related functions.
   - Changed one assertion in the native auth smoke test.
   - Analysis: This is a test file. The changes are removing tests that might be outdated. However, we must check if the removal of tests could hide a bug. But note: the test is for the desktop shell contract. The removed assertions were checking for the existence of a function and its usage. The change in the native auth smoke test replaces an assertion about `completeNativeReturnBounce` with a check for the pathname being `/auth/native-return` and then changes the way the test waits for the native protocol request.
   - We must check if the new test is correct. However, without context, it's hard to say. But note: the test is being updated to match the implementation in `smoke-native-auth.mjs` (see next file). Since the test and the implementation are being changed together, it might be okay. However, we are to look for bugs in the change.

   Let's look at the change in the test:
   - Removed: 
        assert.match(smokeSource, /async function completeNativeReturnBounce/);
        assert.match(smokeSource, /parsedRedirect\.pathname !== '\/auth\/native-return'/);
        assert.match(smokeSource, /redirectUrl\.startsWith\(NATIVE_AUTH_CALLBACK_PREFIX\)/);
   - Added:
        assert.match(smokeSource, /parsed\.pathname === '\/auth\/native-return'/);
        assert.match(smokeSource, /waitForNativeProtocolRequest\(page, '\/auth\/native-return'\)/);

   This seems to be updating the test to match a refactoring in the implementation (see next file). It doesn't appear to introduce a bug in the test itself.

3. `apps/desktop/scripts/smoke-native-auth.mjs`
   - Refactored the `completeNativeReturnBounce` function into `requestNativeCallbackRedirect`.
   - The old function `completeNativeReturnBounce` is removed and its logic is incorporated into `requestNativeCallbackRedirect`.
   - Analysis: We must check if the refactoring introduces any logic error.

   Old function `completeNativeReturnBounce`:
        async function completeNativeReturnBounce(page, nativeReturnUrl, label) {
          const parsed = new URL(nativeReturnUrl);
          const callbackPromise = Promise.race([
            waitForNativeProtocolRequest(page, parsed.pathname),
            waitForNativeRedirectResponse(page, parsed.pathname),
          ]);
          await page.goto(nativeReturnUrl, {
            waitUntil: 'domcontentloaded',
            timeout: 60_000,
          });
          const nativeCallbackUrl = await callbackPromise;
          if (!nativeCallbackUrl.startsWith(NATIVE_AUTH_CALLBACK_PREFIX)) {
            throw new Error(
              `${label} native-return bounce did not yield the Electron app callback: ${nativeCallbackUrl}`
            );
          }
          return nativeCallbackUrl;
        }

   New code in `requestNativeCallbackRedirect` (after the refactor):
        async function requestNativeCallbackRedirect(page, callbackPath, label) {
          const redirectUrl = await requestRedirect(page, callbackPath, label);
          if (redirectUrl.startsWith(NATIVE_AUTH_CALLBACK_PREFIX)) {
            return redirectUrl;
          }

          const parsed = new URL(redirectUrl);
          if (parsed.pathname === '/auth/native-return') {
            const callbackPromise = Promise.race([
              waitForNativeProtocolRequest(page, '/auth/native-return'),
              waitForNativeRedirectResponse(page, '/auth/native-return'),
            ]);
            await page.goto(redirectUrl, {
              waitUntil: 'domcontentloaded',
              timeout: 60_000,
            });
            return await callbackPromise;
          }

          throw new Error(
            `${label} did not return the Electron app callback: ${redirectUrl}`
          );
        }

   The logic appears to be the same: 
   - If the redirectUrl is already the Electron app callback (starts with NATIVE_AUTH_CALLBACK_PREFIX), return it.
   - Else, if the redirectUrl is a native-return bounce (pathname is '/auth/native-return'), then wait for either the protocol request or the redirect response and return that.
   - Else, throw an error.

   The old function did the same: 
        It would go to the nativeReturnUrl (which is the redirectUrl in the new code) and then wait for the protocol request or redirect response on the parsed.pathname (which would be '/auth/native-return' in the bounce case) and then check that the result starts with the Electron app callback prefix.

   The new code does the same, but splits the bounce case into a separate branch.

   However, note: in the old function, the `waitForNativeProtocolRequest` and `waitForNativeRedirectResponse` were called with `parsed.pathname` (which is the pathname of the nativeReturnUrl). In the new code, in the bounce branch, we are calling them with the hardcoded string '/auth/native-return'. This is safe because we just checked that `parsed.pathname === '/auth/native-return'`.

   So the logic is preserved.

   But note: the old function had a timeout of 60_000 ms for the `page.goto` and the same for the new code.

   Therefore, no logic error is apparent.

4. iOS files (AppShellLeftDrawer.swift, AppShellDrawerThreadsFilterTests.swift, JovieUITests.swift)
   - These are removing an enum `AppShellDrawerSurfaceLayout` and its tests, and inlining the constant `0.85` for `minimumScaleFactor`.
   - Also, in JovieUITests.swift, removing some assertions about the "Offline" status.
   - Analysis: 
        - The removal of the enum and inlining of the constant is a refactor. It doesn't change behavior because the constant is the same.
        - The removal of the tests for the enum is acceptable if the enum is removed.
        - The removal of the offline status assertions in JovieUITests.swift: we must check if this is safe. The test was checking for the existence of an "Offline" status indicator. The change removes two assertions and leaves one that checks for a longer string: "Offline. Drafts stay on this device and cached history remains available.".
        - This suggests that the UI now shows a longer offline message. The test update is appropriate. No logic error.

5. `apps/web/.lighthouserc.public-launch.json`
   - Added `puppeteerLaunchOptions` with args and protocolTimeout.
   - Analysis: This is a configuration file for Lighthouse. No logic, security, or data integrity issues.

6. `apps/web/app/api/cron/daily-maintenance/route.ts`
   - Removed the call to `runOnboardingScriptAggregation` (which was step 7) and renumbered the steps.
   - Analysis: This removes a daily maintenance job. We must check if this removal is intentional and if it leaves any dangling references.
        - The function `runOnboardingScriptAggregation` is imported from '@/lib/onboarding/script-aggregation' and that import is removed.
        - The job is removed from the cron.
        - We must check if there are any callers of `runOnboardingScriptAggregation` elsewhere. But note: the PR also changes the onboarding script engine and script files (see below). It's possible that the onboarding script aggregation is no longer needed because of a change in how the onboarding script works.
        - However, we are to look for bugs in the change. The removal of a cron job might be intentional. Without more context, we cannot say it's a bug. But note: the PR description doesn't mention this. However, the agent might have made this change as part of the "codex shipper" follow-up.

        We note that the onboarding script engine and script files are changed (see below). It's possible that the aggregation job is no longer required.

        However, we must check for cross-file impact: if the function `runOnboardingScriptAggregation` is still used elsewhere, then removing it would break. But the import is removed from this file, and we don't see it in the diff of other files. We can only rely on the diff provided.

        Since we don't see the function being used elsewhere in the diff, and the import is removed, it's likely safe.

        But note: the function `runOnboardingScriptAggregation` is defined in `apps/web/lib/onboarding/script-aggregation.ts` (not shown in the diff). We are not changing that file in this PR. So if we remove the call and the import, but the function still exists, then it's not a breaking change for other files because we are not removing the function definition, just not calling it.

        However, we are removing the import and the call. If the function is not used anywhere else, then it's dead code and can be removed. But we are not removing the function definition in this PR. So we are leaving dead code. That is not a bug per se, but it's unclean. However, the PR is about the codex shipper, so it might be out of scope.

        We are to look for bugs. Leaving dead code is not a bug that causes incorrect behavior, but it is a code health issue. However, the instructions say to focus on correctness/security/requirements only. So we can ignore this.

        But note: the PR also changes the onboarding script engine and script files. It's possible that the aggregation job is replaced by something else. Without more context, we cannot be sure.

        However, we are to look for specific issues. We don't see any obvious bug in the removal.

7. `apps/web/app/layout.tsx`
   - Added import for `buildGoogleConsentInitScript`, `isValidGaMeasurementId`, `shouldMountGoogleAnalytics` from '@/lib/tracking/google-consent-mode'.
   - Changed the GA init script from loading an external script (`/ga-consent-init.js`) to inlining the script via `dangerouslySetInnerHTML` with the content built by `buildGoogleConsentInitScript(gaMeasurementId)`.
   - Analysis: 
        - This change moves from an external script to an inline script. 
        - The function `buildGoogleConsentInitScript` is expected to return a string that is safe to inject. We must check if it properly escapes the input to avoid XSS.
        - The comment says: `// biome-ignore lint/security/noDangerouslySetInnerHtml: script body is generated from a validated GA measurement ID.`
        - We must trust that `buildGoogleConsentInitScript` uses a validated GA measurement ID (via `isValidGaMeasurementId`). 
        - The function `isValidGaMeasurementId` is imported and used in the condition: `shouldMountGa && isValidGaMeasurementId(gaMeasurementId)`
        - So if the GA measurement ID is valid, then we build the script. We assume that `buildGoogleConsentInitScript` does not introduce XSS because it uses the validated ID in a safe way (e.g., it doesn't allow arbitrary JS).
        - Without seeing the function, we cannot be 100% sure, but the comment and the validation suggest it's safe.
        - Therefore, no XSS vulnerability is apparent.

8. `apps/web/components/molecules/ConfirmDialog.tsx`
   - This file is completely rewritten. It was previously a re-export of `@jovie/ui`'s ConfirmDialog, now it's a local implementation.
   - Analysis: 
        - We must check if the new implementation has any logic errors, security issues, etc.
        - The component uses state and refs to handle loading and prevent double confirmation.
        - It looks correct: 
            - `handlingRef` is used to prevent multiple concurrent confirms.
            - `isPending` state is set to true during the confirm call and then reset.
            - The loading state is derived from `isLoading` prop or `isPending` state.
            - The confirm button shows "Please wait..." when loading.
        - No obvious logic error.
        - Security: The component does not use `dangerouslySetInnerHTML` or similar, so no XSS risk from the component itself.
        - However, note that the `body` and `children` props are rendered as `body ?? children`. If these contain user-controlled HTML, there could be an XSS risk. But the component is a confirmation dialog and likely used with trusted content. Moreover, the previous version (from `@jovie/ui`) would have had the same risk. So this is not a new issue.
        - Therefore, no new security issue.

9. `apps/web/components/shell/ThreadTurn.tsx`
   - Changed the className for the message bubble: 
        from: `max-w-[70%] rounded-xl rounded-br-[10px] px-3 py-1.5 text-app leading-relaxed bg-(--surface-1)/80 text-primary-token`
        to: `max-w-[75%] rounded-2xl rounded-br-md px-3.5 py-2 text-app leading-relaxed bg-(--surface-1)/80 text-primary-token`
   - Analysis: This is a styling change. No logic, security, or data integrity impact.

10. `apps/web/components/shell/ThreadView.tsx`
    - Changed the space-y from `space-y-2.5` to `space-y-4` for the container of the thread turns.
    - Analysis: Styling change. No impact.

11. `apps/web/lib/chat/onboarding-script/engine.ts`
    - Removed the import of `loadScriptBank` and `pickFromBank` from './line-source'.
    - Changed the import from './script' to only import `pickLine`, `renderLine`, and `ScriptLine`.
    - Removed the `bank` variable and the `pick` function in `decideFallbackTurn`.
    - Changed all calls from `pick(stepId)` to `pickLine(stepId, sessionId)`.
    - Removed the `pick` parameter from the `decisionTurn` function and adjusted its callers.
    - Analysis: 
        - The engine now uses `pickLine` from './script' which takes a sessionId to determine the variant.
        - This is a refactor to move the variant picking logic into the script module.
        - We must check if the `pickLine` function is implemented correctly (see changes in `script.ts` below).
        - The removal of the `bank` and `pick` function is safe because we are now using the new method.
        - The function `decisionTurn` no longer takes the `pick` parameter, and its callers are adjusted accordingly.
        - No logic error is apparent.

12. `apps/web/lib/chat/onboarding-script/script.ts`
    - Added `SCRIPT_STEP_IDS` array and export.
    - Added `isScriptStepId` function and `hashSessionId` function.
    - Changed the `hashSessionId` function to use a different algorithm (note: the old version used `char.codePointAt(0) ?? 0`, the new version uses `sessionId.charCodeAt(i)` and then does `hash = (hash * 31 + sessionId.charCodeAt(i)) >>> 0;`).
    - Changed the `pickLine` function to use the new `hashSessionId` and to throw if no candidates are found.
    - Removed the old `SCRIPT_STEP_IDS` and `isScriptStepId` from the bottom (they are now at the top).
    - Analysis:
        - The new `hashSessionId` function: 
            Old: 
                export function hashSessionId(sessionId: string): number {
                  let hash = 0;
                  for (const char of sessionId) {
                    hash = (hash * 31 + (char.codePointAt(0) ?? 0)) >>> 0;
                  }
                  return hash;
                }
            New:
                export function hashSessionId(sessionId: string): number {
                  let hash = 0;
                  for (let i = 0; i < sessionId.length; i += 1) {
                    hash = (hash * 31 + sessionId.charCodeAt(i)) >>> 0;
                  }
                  return hash;
                }
            - The old version used `codePointAt(0)` which for a single character string is the same as `charCodeAt(0)` for BMP characters, but for non-BMP (emoji, etc.) it would be different. However, the sessionId is likely a UUID or similar, so it's ASCII. Therefore, the change is safe for ASCII sessionIds. But note: the sessionId might be any string. The new version uses `charCodeAt(i)` which for a surrogate pair would give the code unit, not the code point. The old version used `codePointAt(0)` for each character (which for a string iterated by `for...of` gives the code point). 
            - However, note the old version iterated by `for (const char of sessionId)` which gives each Unicode character (code point). The new version iterates by index and gets the code unit (16-bit). This is a breaking change for non-BMP characters.

        - Why does this matter? The sessionId is likely generated by the system (e.g., a UUID or a random string). UUIDs are hexadecimal and only use 0-9a-f, so they are ASCII and BMP

itstimwhite (commented):

itstimwhite:
We are given a diff for PR #12848. We need to review for logic errors, security, data integrity, error handling, and cross-file impact.

Let's go through the changes file by file, focusing on the areas of concern.

Note: The PR description says it's a chore and about the codex shipper skipping type:epic issues. However, the diff shows many changes across the codebase. We must review each change for correctness.

We'll break down the review by the files that have been changed.

However, note that the PR is large (39 files). We must be efficient and focus on potential issues.

Given the instructions, we are to look for:
1. Logic errors
2. Security
3. Data integrity
4. Error handling
5. Cross-file impact

We'll go through each file in the diff and note any issues.

But note: The PR is from an automated agent (codex-issue-shipper) and the description says it's a follow-up to skip type:epic issues. However, the changes are not limited to that. We must review the entire diff.

Let's start:

.github/workflows/ci.yml
Change: In the concurrency group for the golden-path job, changed from a fixed group to include `${{ github.job }}`.
Original: `group: neon-endpoint-pool-ci-golden-path`
New: `group: neon-endpoint-pool-${{ github.job }}-ci-golden-path`

Reason given: "Include github.job to satisfy the repo-wide Neon endpoint pool contract while keeping golden-path runs in a single serialized slot."

This change is about concurrency control. It ensures that each job has its own concurrency group, but note that the comment says it's to keep golden-path runs in a single serialized slot. However, by including the job name, we are now having multiple groups (one per job) so they can run in parallel? But the comment says we want to serialize golden-path runs.

Wait: The original comment said: "Constant group so at most ONE golden-path job runs repo-wide." meaning that all golden-path jobs (regardless of the job name) would be in the same group and thus only one runs at a time.

The new change: `group: neon-endpoint-pool-${{ github.job }}-ci-golden-path` would create a separate group for each job (e.g., build, test, etc.) so that multiple golden-path jobs (if they are different job names) could run concurrently.

But the comment says: "Include github.job to satisfy the repo-wide Neon endpoint pool contract while keeping golden-path runs in a single serialized slot."

This seems contradictory. However, note that the condition for this job is very specific (it only runs under certain conditions). It's possible that the job name is always the same for this particular job? Let's look at the job name in the workflow.

We don't have the full workflow, but from the diff we see that this job is part of a matrix or has a specific name? Actually, the job is defined in the workflow and we are only seeing a snippet.

Given the comment, it seems the intent is to have a single group for all golden-path runs. But the change does the opposite.

However, note that the condition for this job includes:
`if: ${{ needs.ci-path-changes.outputs.skip == 'false' && ... }}`

And the concurrency group is set to `neon-endpoint-pool-${{ github.job }}-ci-golden-path`.

If the job name (github.job) is the same for every instance of this job (which it should be, because it's the same job definition) then the group would be the same for all runs of this job. But if there are multiple jobs in the workflow that have the same condition and are golden-path, then they would have different group names and could run in parallel.

But the original comment was about preventing two concurrent runs from deleting each other's Clerk test users. The concern is about the Neon endpoint pool and the Clerk dev instance.

The change might be incorrect if it allows multiple golden-path jobs (with different job names) to run concurrently and interfere.

However, without the full context of the workflow, it's hard to say. But note that the PR is labeled as a chore and the agent's log is about skipping type:epic issues. This change might be unrelated to the main purpose.

Given the risk of concurrency issues (which could lead to flaky tests due to shared resources), we should flag this as a potential logic error.

But note: the original comment said "Do not parameterize this group." and now we are parameterizing it by the job. This might break the intended serialization.

Therefore, we have a potential issue: the concurrency group is now per job, which might allow multiple golden-path jobs (if they have different names) to run concurrently and cause interference.

However, we must check if the job name is fixed for this job. In the workflow, the job is likely named something like "golden-path" or similar. But we don't see the job name in the diff.

Given the information, we cannot be 100% sure, but the change removes the constant group and makes it dependent on the job name. If the job name is the same for every run of this job (which it should be) then it's safe. But if the workflow has multiple jobs that match this condition and are golden-path, then they would have different group names and run in parallel.

But note: the condition is very specific and likely only one job in the workflow matches it. However, without the full workflow we cannot be certain.

Given the risk and the original comment's warning, we should consider this a potential logic error.

But wait: the agent's log says it's a follow-up to skip type:epic issues. This change in ci.yml doesn't seem related. It might be a mistake.

Alternatively, it might be that the job name is always the same (for example, the job is named "ci-golden-path") and then the group becomes `neon-endpoint-pool-ci-golden-path-ci-golden-path` which is different from the original. But that would break the original intention because now every run of this job would be in a group that includes the job name twice? Actually, the original group was `neon-endpoint-pool-ci-golden-path` and the new one is `neon-endpoint-pool-${{ github.job }}-ci-golden-path`. If the job name is "ci-golden-path", then the new group is `neon-endpoint-pool-ci-golden-path-ci-golden-path` which is different.

This would break the serialization because now every run of this job would be in a unique group (if the job name is fixed) but wait: if the job name is fixed, then the group is fixed too. So it would still be one group for all runs of this job.

Example: if the job is named "golden-path", then:
  Original group: neon-endpoint-pool-ci-golden-path
  New group: neon-endpoint-pool-golden-path-ci-golden-path

But note: the original group had "ci-golden-path" in it, and the new one has the job name and then "-ci-golden-path". So unless the job name is empty, it's a different string.

However, the point is: if the job name is fixed for this job, then the group is fixed and all runs of this job will use the same group -> serialized.

But what if there are multiple jobs in the workflow that are golden-path? Then each would have their own group (because they have different job names) and could run in parallel.

Given the original concern about Clerk dev instance and Neon endpoint pool being shared resources that cannot handle concurrent golden-path runs, we must ensure that only one golden-path job runs at a time across the entire workflow.

Therefore, the change is potentially unsafe because it allows multiple golden-path jobs (with different names) to run concurrently.

We should revert to the original fixed group or ensure that the group is the same for all golden-path jobs.

But note: the condition for this job might be such that only one job in the workflow matches it. We don't have the full workflow.

Given the ambiguity, and since the original comment explicitly said "Do not parameterize this group", we should consider this change as a potential issue.

However, note that the PR is from an automated agent and the agent might have made a mistake. We are to flag issues we can point to.

Let's look at the next files.

But note: we are only allowed to flag issues we can point to specific lines for.

We'll continue reviewing.

apps/desktop/scripts/desktop-shell-contract.test.mjs
Changes: Removed several assertions and changed one.

Specifically, removed:
  - Assertions about `isAuthReturnDeepLinkCandidate` function and its usage.
  - Changed an assertion in the smoke test.

We removed:
  assert.match(mainSource, /function isAuthReturnDeepLinkCandidate\(urlString: string\): boolean/);
  assert.match(mainSource, /isAuthReturnDeepLinkCandidate\(arg\) && !parseDesktopAuthReturnDeepLink\(arg\)/);
  assert.doesNotMatch(mainSource, /startsWith\(`\$\{AUTH_RETURN_PROTOCOL\}\/\/\$\{AUTH_RETURN_HOST\}`\)/);

And in the smoke test:
  Changed from:
      assert.match(smokeSource, /async function completeNativeReturnBounce/);
      assert.match(smokeSource, /parsedRedirect\.pathname !== '\/auth\/native-return'/);
      assert.match(smokeSource, /redirectUrl\.startsWith\(NATIVE_AUTH_CALLBACK_PREFIX\)/);
  To:
      assert.match(smokeSource, /parsed\.pathname === '\/auth\/native-return'/);
      assert.match(smokeSource, /waitForNativeProtocolRequest\(page, '\/auth\/native-return'\)/);

We also removed the function `completeNativeReturnBounce` and changed the logic in `requestNativeCallbackRedirect`.

This seems to be a refactor of the native auth flow. We must check if the changes are correct.

However, without context of the actual implementation, it's hard to say. But note that the test is checking the smoke source (which is the compiled desktop app?).

We are removing tests for certain functions and changing the expected behavior.

This could be a logic error if the removed functions are still used and the test no longer covers them.

But note: the test file is named `desktop-shell-contract.test.mjs` and it's testing the desktop shell contract. The changes might be intentional.

However, we removed an assertion that the function `isAuthReturnDeepLinkCandidate` exists and is used in a certain way. If that function is still present in the code, then the test is now missing coverage for it.

But we don't have the code for the desktop shell, only the test.

Given that we are removing tests, we must be cautious. However, the PR description doesn't mention this, and the agent's log is about skipping type:epic issues. This might be unrelated.

But note: the test is failing without these changes? Or the agent is updating the test to match a refactor?

We cannot tell from the diff alone. However, the removal of the function `completeNativeReturnBounce` and the changes in the smoke test suggest that the function was removed from the code and the test is updated accordingly.

If that's the case, then the test changes are correct.

But we must check: did the agent also change the implementation? We don't see the implementation in this diff.

We are only seeing the test file. So we cannot verify if the implementation was changed.

However, note that the PR has many files. We see changes in `apps/desktop/scripts/smoke-native-auth.mjs` as well.

Let's look at that file.

apps/desktop/scripts/smoke-native-auth.mjs
Changes: 
  - Removed the function `completeNativeReturnBounce`
  - Changed the function `requestNativeCallbackRedirect` to handle the native-return bounce internally.

Specifically, the old code:
  - Had a helper function `completeNativeReturnBounce` that would:
        const parsed = new URL(nativeReturnUrl);
        ... 
        await page.goto(nativeReturnUrl);
        const nativeCallbackUrl = await callbackPromise;
        ... check that it starts with NATIVE_AUTH_CALLBACK_PREFIX

  - In `requestNativeCallbackRedirect`:
        if the redirectUrl starts with NATIVE_AUTH_CALLBACK_PREFIX, return it.
        else if the parsedRedirect.pathname is not '/auth/native-return', throw an error.
        else, call `completeNativeReturnBounce`.

  The new code:
      - Removed `completeNativeReturnBounce`.
      - In `requestNativeCallbackRedirect`:
          if the redirectUrl starts with NATIVE_AUTH_CALLBACK_PREFIX, return it.
          else:
              const parsed = new URL(redirectUrl);
              if (parsed.pathname === '/auth/native-return') {
                  // then do the same as what was in completeNativeReturnBounce: 
                  const callbackPromise = Promise.race([
                    waitForNativeProtocolRequest(page, '/auth/native-return'),
                    waitForNativeRedirectResponse(page, '/auth/native-return'),
                  ]);
                  await page.goto(redirectUrl, { waitUntil: 'domcontentloaded', timeout: 60_000 });
                  return await callbackPromise;
              } else {
                  throw new Error(`${label} did not return the Electron app callback: ${redirectUrl}`);
              }

This is essentially inlining the helper function and removing the helper.

The test changes in `desktop-shell-contract.test.mjs` are consistent with this: they removed the assertions for the helper function and changed the smoke test to expect the new behavior.

Therefore, this seems to be a refactor and the test updates are appropriate.

We don't see an obvious error here.

apps/ios/Jovie/Features/AppShell/AppShellLeftDrawer.swift
Changes: 
  - Removed the enum `AppShellDrawerSurfaceLayout` (which had three static constants)
  - In `DrawerSurfaceButton`, changed `.minimumScaleFactor(AppShellDrawerSurfaceLayout.labelMinimumScaleFactor)` to `.minimumScaleFactor(0.85)`

This is a simple refactor: inlining the constant. The constant was 0.85, so it's the same value.

No issue.

apps/ios/JovieTests/AppShellDrawerThreadsFilterTests.swift
Changes: 
  - Removed the entire struct `AppShellDrawerSurfaceLayoutTests` (which had three tests for the constants we just removed)

This is consistent with the removal of the enum. No issue.

apps/ios/JovieUITests/JovieUITests.swift
Changes: 
  - Removed several assertions related to "Offline" status indicators.

Specifically, removed:
      XCTAssertTrue(app.staticTexts["Offline"].exists, ...);
      XCTAssertEqual(app.staticTexts.matching(NSPredicate(format: "label == %@", "Offline")).count, 1, ...);
      XCTAssertFalse(app.descendants(matching: .any)["shell-drawer-account"].staticTexts["Offline"].exists, ...);

We must check if these assertions are still valid. The test is for the offline chat state.

The remaining assertions are:
      XCTAssertTrue(app.staticTexts["Offline. Drafts stay on this device and cached history remains available."].exists, ...);

And then later in the test (after opening the drawer) we have:
      XCTAssertEqual(app.staticTexts.matching(NSPredicate(format: "label == %@", "Offline")).count, 1, ...);   [This one is still there?]

Wait, let's look at the diff:

Original:
      XCTAssertTrue(app.staticTexts["Offline"].exists, ...);
      XCTAssertEqual(app.staticTexts.matching(NSPredicate(format: "label == %@", "Offline")).count, 1, ...);
      XCTAssertTrue(app.staticTexts["Offline. Drafts stay on this device and cached history remains available."].exists, ...);

Then after that, there was a block for opening the drawer and checking for duplicate offline indicators.

In the diff, we see:
  - The first two assertions about "Offline" are removed.
  - The third assertion (about the long string) remains.
  - Then later, in the drawer account header check, we removed:
        XCTAssertFalse(app.descendants(matching: .any)["shell-drawer-account"].staticTexts["Offline"].exists, ...);

So we removed the check for the standalone "Offline" string and the check for duplicate offline indicators in the drawer.

Why? The test might be flaky or the UI changed.

But note: the test is for the offline chat state. The offline state might now be represented only by the long string and not by the standalone "Offline".

If that's the case, then the test update is correct.

However, we removed the check that there is exactly one standalone "Offline" indicator. If the UI now only shows the long string, then that's fine.

But we also removed the check for duplicate offline indicators in the drawer. If the drawer should not show any offline indicator, then we should still check that.

The remaining test after opening the drawer is:
      XCTAssertTrue(app.staticTexts["Offline. Drafts stay on this device and cached history remains available."].exists, ...);
      XCTAssertEqual(app.staticTexts.matching(NSPredicate(format: "label == %@", "Offline")).count, 1, ...);   [This line is still present?]

Wait, let me re-read the diff:

The diff shows:
      -    XCTAssertTrue(
      -      app.staticTexts["Offline"].exists,
      -      "Shell header did not show the canonical offline status.\n\(app.debugDescription)"
      -    )
      -    XCTAssertEqual(
      -      app.staticTexts.matching(NSPredicate(format: "label == %@", "Offline")).count,
      -      1,
      -      "Offline chat launch showed more than one standalone offline status indicator.\n\(app.debugDescription)"
      -    )
      +    XCTAssertTrue(
      +      app.staticTexts["Offline. Drafts stay on this device and cached history remains available."].exists,
      +      "Offline chat empty state did not explain draft/cache behavior.\n\(app.debugDescription)"
      +    )

And then later:
      -    XCTAssertFalse(
      -      app.descendants(matching: .any)["shell-drawer-account"].staticTexts["Offline"].exists,
      -      "Drawer account header showed a redundant offline status.\n\(app.debugDescription)"
      -    )

So we removed two assertions and left one (the long string) in the initial state check.

But note: after the initial state check, we have:
      XCTAssertEqual(app.staticTexts.matching(NSPredicate(format: "label == %@", "Offline")).count, 1, ...);

This line is not in the diff, meaning it was not changed. So it's still there.

Wait, let me look at the original code (not shown in the diff) but we can infer:

itstimwhite (commented):

itstimwhite:
We are given a diff of a PR. We need to look for logic errors, security issues, data integrity problems, error handling issues, and cross-file impact.

Let's go through the changes file by file, focusing on the areas of concern.

However, note that the PR description says: "chore(codex): Candidate follow-up: codex shipper should skip type:epic issues (retry-loop on #12729)". This suggests that the changes might be related to the codex issue shipper, but we see changes in many files.

We must be cautious: the PR has 39 changed files. We cannot review every line in detail, but we can look for obvious issues.

Given the instructions, we are to focus on:
1. Logic errors
2. Security
3. Data integrity
4. Error handling
5. Cross-file impact

We'll go through the diff and note any potential issues.

Let's start:

.github/workflows/ci.yml
Change: 
  - Removed a comment about constant group for Neon endpoint pool and Clerk dev instance.
  - Changed the concurrency group from `neon-endpoint-pool-ci-golden-path` to `neon-endpoint-pool-${{ github.job }}-ci-golden-path`

This change is about CI concurrency. It might be to avoid conflicts between different jobs. However, note that the comment says: "Include github.job to satisfy the repo-wide Neon endpoint pool contract while keeping golden-path runs in a single serialized slot."

This seems to be a refactor to allow multiple golden-path jobs to run in parallel but still serialize per job? Actually, the group now includes the job name, so different jobs (e.g., test, build) would have different groups and could run in parallel. But the original comment said that the Neon endpoint pool and Clerk dev instance are shared resources that must be serialized. 

If we are now allowing different jobs to run in parallel (because they have different group names) then we might break the contract that only one golden-path job runs at a time (to avoid conflicts with the Neon endpoint pool and Clerk dev instance). 

However, note that the condition for this job is very specific (it runs only for golden-path). The condition includes:
  `contains(github.event.pull_request.labels.*.name, 'launch-candidate') || contains(github.event.pull_request.labels.*.name, 'testing')`

And the job is only for golden-path. So if we are changing the group to include the job name, then if there are multiple golden-path jobs (with different job names) they would run in parallel. But the condition for this job is the same for all? Actually, the job is defined once. So there is only one job with this condition? 

Looking at the condition, it is a single job. So the group name change from a fixed string to one that includes `${{ github.job }}` would just append the job name to the group. Since there is only one job that matches this condition (the job is defined once), then the group name becomes `neon-endpoint-pool-<job_name>-ci-golden-path`. 

But note: the job name in the workflow is not shown in the diff. We only see the condition and the concurrency group. Without the job name, we cannot be sure. However, the change is likely safe because it's just making the group name more specific to the job, but since there's only one job that uses this condition, it doesn't change the concurrency behavior (it's still one group per job, and there's only one job). 

But wait: the condition might be true for multiple jobs? Actually, the condition is inside a job. The job is defined by the lines above. We don't see the job name in the diff. However, the change is to include the job name in the group. This is a common pattern to allow different jobs to run in parallel while still serializing same-job runs. 

Given that the original comment said that the Neon endpoint pool and Clerk dev instance are shared and must be serialized, and now we are making the group per job, we must ensure that the job name is the same for all runs that should be serialized. 

But note: the condition for this job is very specific and likely only one job in the workflow matches it. So it's probably safe.

However, we are not seeing the entire workflow. We must trust that the change is correct because the author says it satisfies the contract.

But let's note: the original comment said "Do not parameterize this group." and now we are parameterizing it with the job. This might be a risk if the job name changes or if there are multiple jobs that should be serialized together. 

Given the lack of context, we cannot be sure. However, the PR is from an automated agent and we are to look for clear errors. We don't see an obvious logic error here.

apps/desktop/scripts/desktop-shell-contract.test.mjs
Changes:
  - Removed several assertions about `isAuthReturnDeepLinkCandidate` and related functions.
  - Changed one assertion in the native auth smoke test.

Specifically, in the test 'native auth smoke keeps browser callbacks on the browser auth origin':
  - Removed: 
        assert.match(smokeSource, /async function completeNativeReturnBounce/);
        assert.match(smokeSource, /parsedRedirect\.pathname !== '\/auth\/native-return'/);
        assert.doesNotMatch(smokeSource, /startsWith\(`\$\{AUTH_RETURN_PROTOCOL\}\/\/\$\{AUTH_RETURN_HOST\}`\)/);
  - Changed: 
        assert.match(smokeSource, /parsed\.pathname === '\/auth\/native-return'/);
        and changed the next assert from checking for a startsWith to checking for a function call `waitForNativeProtocolRequest`.

Also, in the function `requestNativeCallbackRedirect` (in smoke-native-auth.mjs) we see a refactor: 
  - Removed the helper function `completeNativeReturnBounce` and inlined its logic.

This seems to be a refactor of the test and the helper function. We must check if the refactor is correct.

In the test, we removed assertions that checked for the absence of a certain pattern (the startsWith) and the presence of a function `completeNativeReturnBounce`. Then we changed the assertion about the pathname to be equal to '/auth/native-return' (from not equal) and changed the next assertion to check for a call to `waitForNativeProtocolRequest`.

In the implementation of `requestNativeCallbackRedirect` (in smoke-native-auth.mjs), we see:
  - The function `completeNativeReturnBounce` is removed.
  - The function `requestNativeCallbackRedirect` is rewritten: 
        if the redirectUrl starts with NATIVE_AUTH_CALLBACK_PREFIX, return it.
        else, if the parsed pathname is '/auth/native-return', then we do a goto and wait for either a protocol request or a redirect response.
        else, throw an error.

This seems to be a correction: previously, if we got a redirectUrl that was not the native auth callback prefix, we would check if it was not '/auth/native-return' and then throw an error? Actually, the old code:

  const parsedRedirect = new URL(redirectUrl);
  if (parsedRedirect.pathname !== '/auth/native-return') {
    throw new Error(...);
  }
  return await completeNativeReturnBounce(page, redirectUrl, label);

Now, the new code:

  const parsed = new URL(redirectUrl);
  if (parsed.pathname === '/auth/native-return') {
    ... // do the bounce and return the callbackPromise
  }
  throw new Error(...);

So the logic is inverted: now we only attempt the bounce if the pathname is '/auth/native-return', otherwise we throw.

This matches the test change: the test now expects that the pathname is '/auth/native-return' (so we take the bounce path) and then we wait for the protocol request.

This seems to be a fix: the test was previously expecting that the pathname is NOT '/auth/native-return' (which was wrong) and now it is corrected.

However, we must check if the test is correct. The test is named: 'native auth smoke keeps browser callbacks on the browser auth origin'. The test is checking that the native auth smoke test keeps browser callbacks on the browser auth origin. 

The change in the test and the implementation seems to be fixing the logic. We don't see an obvious error.

But note: the test removed several assertions. We must ensure that the test still covers what it intended. However, without the full context, we cannot be sure. But the changes seem to be a refactor that corrects the logic.

apps/desktop/scripts/smoke-native-auth.mjs
We already discussed the refactor of `requestNativeCallbackRedirect`. It looks correct.

apps/ios/Jovie/Features/AppShell/AppShellLeftDrawer.swift
Change: 
  - Removed the enum `AppShellDrawerSurfaceLayout` and replaced the use of `AppShellDrawerSurfaceLayout.labelMinimumScaleFactor` with the literal `0.85`.

This is a removal of an unused enum? We see that the enum was only used in one place. This is safe.

apps/ios/JovieTests/AppShellDrawerThreadsFilterTests.swift
Change:
  - Removed the entire struct `AppShellDrawerSurfaceLayoutTests` which had three tests.

This is because the enum `AppShellDrawerSurfaceLayout` was removed. So the tests are no longer relevant. This is safe.

apps/ios/JovieUITests/JovieUITests.swift
Change:
  - Removed several assertions about the "Offline" status indicator.

Specifically, removed:
  - Checking for the existence of a static text "Offline"
  - Checking that there is exactly one such static text
  - Checking that the drawer account header does not show a redundant offline status.

This might be a problem if the offline status indicator is still expected. However, we don't know the context. The change might be intentional if the UI has changed. But note: the test is for the JovieUITests, and we are removing checks for the offline status. 

We must consider: is this a regression? The test is now less comprehensive. However, the PR description does not indicate that this is related to the offline status. 

But note: the PR is about the codex shipper skipping epic issues. This change in iOS tests seems unrelated. 

We are looking for logic errors, security, etc. Removing test assertions is not a logic error in the production code, but it might be a test gap. However, the instructions say: we are to look for issues that break callers or dependents (cross-file impact) and also logic errors in the code. 

This change is in a test file. We are not to flag test changes unless they indicate a problem in the production code? The instructions say: "No test for this change" is a false positive for pure config/doc/dependency-bump PRs. But this is not a pure config change: it's removing test assertions. 

However, note the learned review rules: "Recent reviews have high rejection rate. Be more careful with factual claims. Only flag issues you can point to specific lines for."

We cannot point to a specific line in the production code that is broken by this test change. Therefore, we should not flag it as a bug in the production code. It might be a test gap, but that is out of scope for our review (we are only to look for correctness/security/requirements in the production code). 

But note: the instructions say: "Cross-file impact — does this change break callers or dependents?" This change is in a test file, so it doesn't break production callers. 

Therefore, we skip.

apps/web/.lighthouserc.public-launch.json
Change:
  - Added "puppeteerLaunchOptions" with args and protocolTimeout.

This is a configuration change for Lighthouse. It seems safe.

apps/web/app/api/cron/daily-maintenance/route.ts
Change:
  - Removed the comment and the call to `runOnboardingScriptAggregation` (which was step 7) and then renumbered the steps.

This removes a daily cron job for onboarding script self-improvement. 

We must check: is this job still needed? The comment says it was for JOV-3806. If the feature is no longer needed, then removing it is safe. However, we must ensure that the removal does not break anything else.

The function `runOnboardingScriptAggregation` is imported from '@/lib/onboarding/script-aggregation' and that import is removed. We must check if that file is still used elsewhere.

But note: the diff also shows changes in the onboarding script files (engine.ts, script.ts, etc.). We see that the onboarding script engine is being refactored (removing the script bank and using a direct pickLine function). 

It is possible that the onboarding script aggregation job is no longer needed because the engine has been changed. 

However, we must check if the function `runOnboardingScriptAggregation` is defined and if it is used elsewhere. The import is removed and the call is removed. We don't see any other changes that would suggest it is used elsewhere. 

But note: the file `apps/web/lib/onboarding/script-aggregation.ts` might still exist and be unused. That is not a production error, but it is dead code. However, we are not to flag dead code unless it causes a problem (like a build error). 

We don't see a build error in the diff. 

Therefore, this change seems safe.

apps/web/app/layout.tsx
Change:
  - Imported `buildGoogleConsentInitScript`, `isValidGaMeasurementId`, `shouldMountGoogleAnalytics` from '@/lib/tracking/google-consent-mode'
  - Changed the GA init script from using an external script file to injecting the script via `dangerouslySetInnerHTML` with the content built by `buildGoogleConsentInitScript`.

This is a change to how Google Consent Mode is implemented. 

We must check for security: 
  - The `buildGoogleConsentInitScript` function takes `gaMeasurementId` and returns a string of JavaScript. 
  - We are then setting `dangerouslySetInnerHTML` with that string.

We must ensure that `gaMeasurementId` is validated. We see that we are checking `shouldMountGa && isValidGaMeasurementId(gaMeasurementId)`. 

The function `isValidGaMeasurementId` is imported from the same module. We must assume it validates the ID to prevent injection. 

Without seeing the implementation of `isValidGaMeasurementId`, we cannot be 100% sure, but it is likely safe because it is a validation function. 

However, note: the original code loaded an external script (`/ga-consent-init.js`). Now we are inlining the script. This changes the CSP: we are now allowing inline scripts? 

But note: we are using `dangerouslySetInnerHTML` to set the content of a script tag. This is still an inline script. 

The original code loaded an external script, which would be allowed by CSP if the script source is allowed. The new code injects an inline script, which would require CSP to allow 'unsafe-inline' for scripts unless we use a nonce or hash. 

We do not see any nonce or hash being used. 

This might be a CSP violation if the CSP does not allow inline scripts. 

We must check: the project's CSP. 

However, we don't have the CSP configuration in the diff. 

But note: the PR also changes the file `apps/web/tests/unit/lib/content-security-policy.test.ts`. Let's look at that.

apps/web/tests/unit/lib/content-security-policy.test.ts
We don't see the diff for this file in the provided changes? Actually, we do see it in the list of changed files, but the diff is not shown in the provided snippet? 

Wait, the provided diff snippet does not include the changes for `apps/web/tests/unit/lib/content-security-policy.test.ts`. We only see the file name in the list of changed files, but the actual diff is not in the provided code block. 

Therefore, we cannot see what changed in that test. 

However, we are to look for security issues. The change in layout.tsx might introduce an inline script without a nonce or hash, which could break CSP. 

But note: the original code also loaded an external script that might have been inlined? Actually, the original code was:

  <script id='ga-consent-init' src='/ga-consent-init.js' />

This is an external script. The new code is:

  <script
    id='ga-consent-init'
    // biome-ignore lint/security/noDangerouslySetInnerHtml: script body is generated from a validated GA measurement ID.
    dangerouslySetInnerHTML={{
      __html: buildGoogleConsentInitScript(gaMeasurementId),
    }}
  />

This is an inline script. 

If the CSP does not allow inline scripts, then this will break. 

We must check if the CSP allows inline scripts for this specific script by using a nonce or hash. We don't see one. 

Therefore, this is a potential security issue (CSP violation) unless the CSP is updated to allow the hash of this inline script or to allow 'unsafe-inline' (which is not recommended). 

However, note the comment: 
  // biome-ignore lint/security/noDangerouslySetInnerHtml: script body is generated from a validated GA measurement ID.

This indicates that the team is aware of the risk and is relying on the validation of the GA measurement ID to make it safe. 

But note: even if the GA measurement ID is validated, the function `buildGoogleConsentInitScript` might return a string that includes more than just the ID. We must check the implementation of that function.

We don't have the diff for `apps/web/lib/tracking/google-consent-mode.ts` in the provided changes. 

But note: the file `apps/web/lib/tracking/google-consent-mode.ts` is not in the list of changed files. So we assume it is unchanged. 

We must look at the function `buildGoogleConsentInitScript` in the existing codebase. 

Without the code, we cannot be sure. However, the comment says the script body is generated from a validated GA measurement ID. 

If the function only uses the GA measurement ID to build a string like:

  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());
  gtag('config', 'G-XXXXXXX', { ... });

then it might be safe. But note: the GA measurement ID is validated, so it should be in the format G-XXXXXXX or UA-XXXXXXX-X. 

However

itstimwhite (commented):

itstimwhite:
We are given a diff for PR #12848. We need to review for logic errors, security, data integrity, error handling, and cross-file impact.

Let's go through the changes file by file, focusing on the areas of concern.

Note: The PR description says it's about "codex shipper should skip type:epic issues", but the diff shows many changes. We must check each change for the above categories.

We'll break down the diff by file and look for issues.

However, note that the PR is large (39 files). We must be efficient and focus on the most critical parts.

Given the instructions, we are to only flag issues we can point to specific lines for.

Let's start:

1. .github/workflows/ci.yml
   - Change: In the concurrency group for the golden-path job, changed from a static group to include `${{ github.job }}`.
   - Reason: To satisfy the repo-wide Neon endpoint pool contract while keeping golden-path runs in a single serialized slot.
   - Analysis: This is about CI concurrency. It doesn't introduce a logic error, security issue, data integrity problem, or error handling issue. It's a CI configuration change. No issue found.

2. apps/desktop/scripts/desktop-shell-contract.test.mjs
   - Removed several assertions about `isAuthReturnDeepLinkCandidate` and related code.
   - Changed the test for native auth smoke: replaced a check for `parsedRedirect.pathname !== '/auth/native-return'` with `parsed.pathname === '/auth/native-return'` and changed the way the native callback is waited for.
   - Analysis: The removed assertions might be testing important behavior, but the test is being updated to match the implementation. However, we must check if the new test is correct.
     - The test now: 
        - Matches the string `parsed.pathname === '/auth/native-return'` (which is a string, not a regex? Actually, it's a string in the source code, but the test uses `assert.match` with a regex? Wait, the line is: `assert.match(smokeSource, /parsed\.pathname === '\/auth\/native-return'/);`
        - This is checking that the source code contains the string `parsed.pathname === '/auth/native-return'` (as a regex pattern). This is acceptable if the code indeed has that line.
     - Then, it changed the wait for the native callback: instead of using `waitForNativeProtocolRequest` and `waitForNativeRedirectResponse` in a race, it now does:
          const callbackPromise = Promise.race([
            waitForNativeProtocolRequest(page, '/auth/native-return'),
            waitForNativeRedirectResponse(page, '/auth/native-return'),
          ]);
          await page.goto(redirectUrl, { ... });
          return await callbackPromise;
     - Previously, it had a function `completeNativeReturnBounce` that did similar but with a check on the URL prefix. Now, that function is removed and the logic is inlined.
   - This seems like a refactor of the test. We don't see an obvious logic error, but note that the test is now checking for a specific string in the source. If the implementation changes, the test might break. However, that's the point of a test.
   - No clear logic error, security, data integrity, or error handling issue.

3. apps/desktop/scripts/smoke-native-auth.mjs
   - Removed the function `completeNativeReturnBounce` and refactored `requestNativeCallbackRedirect`.
   - The refactored function now:
        - If the redirectUrl starts with NATIVE_AUTH_CALLBACK_PREFIX, return it.
        - Else, if the pathname is '/auth/native-return', then it waits for the native protocol request or redirect response and returns the result.
        - Otherwise, throws an error.
   - Previously, it would call `completeNativeReturnBounce` for the case when the redirectUrl did not start with the prefix and the pathname was not '/auth/native-return' (which would then do a bounce and check the prefix again).
   - Now, it only handles the '/auth/native-return' case and throws for anything else.
   - Analysis: This changes the behavior. Previously, if the redirectUrl was not the native auth callback and not the native-return path, it would attempt a bounce (by going to the URL and then waiting for the native protocol). Now, it throws an error immediately for any redirectUrl that doesn't start with the prefix and isn't the native-return path.
   - Is this correct? We don't have the full context, but note that the test in the desktop-shell-contract.test.mjs was updated to match this change (it now expects the native-return path to be handled and the bounce function removed). 
   - However, we must consider: what if there are other valid redirect paths? The function name is `requestNativeCallbackRedirect` and it's used in the smoke test for native auth. The smoke test might only expect the native-return path. 
   - Without more context, it's hard to say if this is an error. But note: the change in the test and this file are consistent. We'll assume it's intentional.
   - No obvious logic error, but we note that the error handling now throws for more cases. However, the function is only used in a test, so it might be acceptable.

4. apps/ios/Jovie/Features/AppShell/AppShellLeftDrawer.swift
   - Removed the enum `AppShellDrawerSurfaceLayout` and replaced the use of `AppShellDrawerSurfaceLayout.labelMinimumScaleFactor` with the hardcoded value `0.85`.
   - Analysis: This is a simplification. The value was 0.85 and now it's hardcoded. No logic error because the value is the same. However, if the value was used elsewhere and now it's duplicated, that might be a maintenance issue, but not a bug. We are only concerned with correctness. No issue.

5. apps/ios/JovieTests/AppShellDrawerThreadsFilterTests.swift
   - Removed the test suite `AppShellDrawerSurfaceLayoutTests`.
   - Analysis: This test suite was testing the constants from the removed enum. Since the enum is removed, the tests are no longer valid. Removing them is appropriate. No issue.

6. apps/ios/JovieUITests/JovieUITests.swift
  
   - Removed several assertions about the "Offline" status indicator.
   - Analysis: These assertions were checking for the presence and count of the "Offline" text. The change removes them. We don't know why, but it might be because the UI changed. However, note that the test still checks for a longer string: "Offline. Drafts stay on this device and cached history remains available." 
   - This removal might be because the standalone "Offline" status is no longer shown. Without more context, we cannot say if this is correct. But note: the test is being updated to match the UI. We'll assume it's intentional. No obvious logic error.

7. apps/web/.lighthouserc.public-launch.json
   - Added `puppeteerLaunchOptions` with args and protocolTimeout.
   - Analysis: This is a configuration change for Lighthouse. No logic error, security, etc.

8. apps/web/app/api/cron/daily-maintenance/route.ts
   - Removed the comment and the call to `runOnboardingScriptAggregation` (which was step 7) and renumbered the steps.
   - Analysis: This removes a daily cron job. We must check if this is intentional and if it has any impact. The PR description doesn't mention it, but the commit message says it closes #12846. We don't have the issue, but note that the onboarding script self-improvement was removed. 
   - However, we are not to worry about the correctness of removing a feature unless it breaks something else. The function `runOnboardingScriptAggregation` is imported from '@/lib/onboarding/script-aggregation' and that import is removed. We must check if that file is still used elsewhere? 
   - But note: the PR also changes the onboarding script engine and script files (see below). It's possible that the onboarding script self-improvement is being replaced by another mechanism. 
   - Without more context, we cannot say if removing this cron job is correct. However, we are to look for logic errors in the change itself. The change removes a block of code and adjusts the comments. There's no obvious logic error in the removal (it's just removing a step). But note: if the function `runOnboardingScriptAggregation` had side effects that are now missing, that could be a problem. However, we are not told that it's incorrect. 
   - We'll note: the removal of a cron job might be intentional and part of the feature. No obvious logic error in the change.

9. apps/web/app/layout.tsx
   - Imported `buildGoogleConsentInitScript`, `isValidGaMeasurementId`, `shouldMountGoogleAnalytics` from '@/lib/tracking/google-consent-mode'.
   - Changed the Google Analytics script insertion: instead of inserting a script tag with src='/ga-consent-init.js', it now inserts an inline script with the content built by `buildGoogleConsentInitScript(gaMeasurementId)`.
   - Analysis: 
        - This change moves from an external script to an inline script. 
        - The comment says: "Must run before React hydration; next/script nonce drift causes local E2E console errors." and now they are using `dangerouslySetInnerHTML` to set the script content.
        - We must check for XSS: the function `buildGoogleConsentInitScript` takes `gaMeasurementId` which is validated by `isValidGaMeasurementId`. We assume that function returns a safe string (only containing the measurement ID and some fixed template). 
        - However, note that the `gaMeasurementId` comes from the environment (publicEnv). We must trust that it's safe. But if an attacker could set the GA measurement ID to something malicious, then we might have an XSS. 
        - The function `isValidGaMeasurementId` (which we don't see the code for) should validate that the ID is in the expected format (e.g., G-XXXXXXX or UA-XXXXXXX-X). If it does, then it's safe. 
        - Without seeing the validation function, we cannot be 100% sure, but the pattern is to validate and then use. This is a common way to avoid XSS. 
        - However, note: the script is being inserted with `dangerouslySetInnerHTML`. If the validation is not sufficient, this could be dangerous. 
        - But note: the PR also changes the onboarding script and other files, and we are not given the validation function. We must assume that the validation is correct because it's used in the condition `shouldMountGa && isValidGaMeasurementId(gaMeasurementId)`. 
        - Therefore, if the validation is correct, then the string built by `buildGoogleConsentInitScript` is safe. 
        - No obvious XSS if the validation is correct. 
        - However, we note that the previous method (loading from '/ga-consent-init.js') might have been safer because it was a static file. Now we are generating script content dynamically. 
        - But the content is built from a validated ID, so it should be safe. 
        - No logic error, but we flag a potential security concern if the validation is not strict enough. However, we cannot see the validation function. 
        - Given the instructions, we can only flag if we see an issue. We don't see the validation function, so we cannot say it's invalid. 
        - We'll note: this is a change that requires trust in the validation function. But since we are not given the validation function, we cannot verify. 
        - However, note that the same validation is used in the condition to decide whether to mount the script. So if the validation passes, we assume the ID is safe. 
        - We'll leave it as no issue found, but note that the reviewer should check the validation function in `google-consent-mode.ts` (which is not in the diff). 

10. apps/web/components/molecules/ConfirmDialog.tsx
    - This file is completely new: it replaces the import from '@jovie/ui' with a local implementation of ConfirmDialog.
    - Analysis: 
        - We must check if this local implementation is correct and if it breaks anything.
        - The component uses `useState` and `useRef` to manage state and prevent double handling.
        - The `handleConfirm` function sets `handlingRef.current` to true to prevent re-entry, then sets `isPending` to true, calls `onConfirm`, and then on finally sets `isPending` to false and `handlingRef` to false.
        - This prevents multiple concurrent confirms, which is good.
        - The `loading` prop is combined with `isPending` to disable the buttons during loading or when the confirm is pending.
        - The confirm button shows "Please wait..." when loading or pending.
        - This seems correct.
        - However, note: the original ConfirmDialog from '@jovie/ui' might have been used elsewhere. By replacing it with a local version, we must ensure that the local version is a drop-in replacement. 
        - The props are the same (we see the interface matches the exported one from '@jovie/ui'? We don't have the original, but the props look reasonable).
        - We must check if the local version is used in the same way. 
        - Since we are replacing the import, and the export is the same (the function `ConfirmDialog` and the type `ConfirmDialogProps`), it should be a drop-in replacement.
        - No obvious logic error. 
        - However, note: the component uses `AlertDialog` from '@jovie/ui'. We assume that is still available and correct.
        - No issue found.

11. apps/web/components/shell/ThreadTurn.tsx
    - Changed the className for the message bubble: 
        - From: `max-w-[70%] rounded-xl rounded-br-[10px] px-3 py-1.5 text-app leading-relaxed bg-(--surface-1)/80 text-primary-token`
        - To: `max-w-[75%] rounded-2xl rounded-br-md px-3.5 py-2 text-app leading-relaxed bg-(--surface-1)/80 text-primary-token`
    - Analysis: This is a UI tweak. Changing the max-width from 70% to 75%, the border radius, and the padding. 
        - This is purely presentational. No logic error, security, etc.

12. apps/web/components/shell/ThreadView.tsx
    - Changed the space-y from 2.5 to 4 in the container for the messages.
    - Analysis: UI tweak. No issue.

13. apps/web/lib/chat/onboarding-script/engine.ts
    - Removed the import of `loadScriptBank` and `pickFromBank` from './line-source'.
    - Changed the import to: `import { pickLine, renderLine, type ScriptLine } from './script';`
    - Removed the `pick` parameter from the `decisionTurn` function and the `decideFallbackTurn` function.
    - In `decisionTurn` and `decideFallbackTurn`, replaced calls to `pick(stepId)` with `pickLine(stepId, sessionId)`.
    - Removed the code that loaded the script bank and created the `pick` function in `decideFallbackTurn`.
    - Analysis: 
        - This change replaces the old way of picking a script line (which used a bank loaded from a file and a session-based hash) with a new way that is defined in `script.ts` (see below).
        - We must check if the new `pickLine` function is correct and if it uses the sessionId appropriately.
        - Looking at the changes in `script.ts` (below), we see that the new `pickLine` function uses a hash of the sessionId to pick a variant. This is the same as the old method (which used `hashSessionId` and then modulo the number of candidates). 
        - The old method in `engine.ts` was:
              const bank = await loadScriptBank();
              const pick = (stepId: ScriptStepId): ScriptLine =>
                pickFromBank(bank, stepId, sessionId);
        - The new method in `script.ts` is:
              export function pickLine(stepId: ScriptStepId, sessionId: string): ScriptLine {
                const candidates = linesForStep(stepId);
                if (candidates.length === 0) {
                  throw new Error(`No script lines defined for step ${stepId}`);
                }
                const hash = hashSessionId(sessionId);
                const picked = candidates[hash % candidates.length];
                if (!picked) {
                  throw new Error(`Variant pick failed for step ${stepId}`);
                }
                return picked;
              }
        - And the `hashSessionId` function is the same as before (but moved to `script.ts`).
        - Therefore, the behavior is the same: it picks a variant based on the sessionId. 
        - However, note: the old method loaded the bank from a file (presumably the script lines) and then picked from that bank. The new method uses the `SCRIPT_LINES` constant (which is imported from the same file) and filters by stepId. 
        - This should be equivalent if the bank loaded by `loadScriptBank` is the same as `SCRIPT_LINES`. 
        - We don't see the `loadScriptBank` function, but it's removed. We assume that the bank was built from the same lines. 
        - Therefore, the change is a refactor and should be behaviorally equivalent. 
        - No logic error.

14. apps/web/lib/chat/onboarding-script/script.ts
    - Added `SCRIPT_STEP_IDS` array and exported it.
    - Added `isScriptStepId` function.
    - Added `hashSessionId` function (which was moved from here? Actually, it was in this file before but now it's updated to use `char.codePointAt(0) ?? 0` -> but wait, the diff shows:
          export function hashSessionId(sessionId: string): number {
            let hash = 0;
            for (let i = 0; i < sessionId.length; i += 1) {
              hash = (hash * 31 + sessionId.charCodeAt(i)) >>> 0;
            }
            return hash;
          }
        - Previously, it was:
          export function hashSessionId(sessionId: string): number {
            let hash = 0;
            for (

itstimwhite (commented):

itstimwhite:
Rebasing for merge queue

itstimwhite:
Re-triggering CI - clearing stale PR Ready failure

itstimwhite:
Zoe overnight merge watch: closing to re-trigger fresh CI (stale PR Ready failure from cancelled previous runs)

itstimwhite:
**PR drain: closing as superseded.**

The core feature of this PR — codex shipper skipping `type:epic` issues — already landed on `main` via a parallel path:
- `scripts/hermes/lib/codex-issue-shipper.ts:70` exports `EPIC_LABEL = 'type:epic'` with exclusion logic (+ tests in `pipeline-scoreboard.test.mjs`, `tracker.mjs:271`)
- `scripts/hermes/jobs/codex-issue-shipper.ts` header already documents the `type:epic` exclusion

Worse, this branch's residual diff vs merge-base would **revert** main's newer spawn-resource-guard work (`SpawnResourceUnavailableError` handling, `ListCodexIssuesResult`, the FETCH_LIMIT=200 fix) back to the older `isSpawnEagain` approach — 38 files / -715 lines of regression if merged with a naive conflict resolution. Since the PR carries `tim-approved`, resolving conflicts blindly would auto-merge that regression.

Reopen only if there is a distinct behavior here not already on main. Dispatch: pr-drain cmr8ftq9o2af507ad2oz5kp3u

Issue #12822, opened 2026-07-03
Packaged Mac app opens to a blank/black window when signed-out (all envs) — no sign-in affordance


itstimwhite:
## Source

* Ad-hoc /goal loop: "get the mac app working in all environments" (permanent Dock apps for prod/staging/local; each must load, auth, show settings with real data).
* Session/branch: itstimwhite/mac-app-all-envs-auth-verify (milan-v4 worktree)

## Symptom

The packaged desktop app (prod `/Applications/Jovie.app` 26.6.61, and locally-built Staging/Local) opens to a **permanently blank window** when the [jov.ie](<http://jov.ie>) session is expired/signed-out.

Evidence (prod, clean single instance, CDP on :9223):

* DevTools target metadata URL = `https://jov.ie/app/chat?runtime=electron`, but the actual renderer document is `about:blank` (`document.readyState=complete`, `body.innerText.length=0`, no `#__next`/root).
* `System Events` reports **0 windows** on plain `open -a "Jovie"`; forcing a window via Cmd+, paints solid black (traffic-lights only).
* Main-process log only shows: update check + `csp-header-missing` for the `/app/chat` **redirect** (307 → /signin; the redirect legitimately lacks CSP). No crash/`render-process-gone`/failure-page entries.
* Interpretation: signed-out → `/app/chat` 307 → `/signin` → the desktop `will-navigate` auth guard cancels the in-window navigation, but **no auth-handoff window opens on passive redirect**, so the main window is stranded on `about:blank`. This matches the "black screen" failure `apps/desktop/src/renderer-recovery.ts` was written for.

## Why it matters

The shipped Mac app is unusable whenever the session expires: user sees a blank/black window with **no sign-in UI and no recovery path**. This is the "mac app not working" reported by the goal.

## Confirmed root-cause context (verified this session)

* prod `jov.ie/signin` and `/start` (200, needsNonce) DO return full CSP + nonce — web CSP is healthy; the desktop `csp-header-missing` events are only on 307 redirects (harmless).
* Desktop CSP watchdog (`apps/desktop/src/desktop-csp-watchdog.ts`) injects a restrictive fallback (`default-src 'self'; script-src 'self'`) on missing CSP; harmless on redirects, but would break a nonce'd page if a real 200 ever shipped without CSP.

## Sub-findings (packaging gaps found + patched locally this session, not yet committed)

[...]

PR #13022, opened 2026-07-03
fix(desktop): isolate Mac auth callback schemes

itstimwhite:
## Summary
- register environment-specific Electron auth schemes: production `jovie`, staging `jovie-staging`, local `jovie-local`
- add staging/local macOS `CFBundleURLTypes` so packaged apps can receive their own callbacks
- make `/auth/native-return` derive its callback scheme from browser origin without writing a wrong transient `jovie://` href
- update desktop shell and smoke contracts for signed-out fallback and per-env native callbacks

Closes #12822

## Verification
- `./scripts/setup.sh` -> Ready to develop
- `pnpm exec biome check --write apps/desktop/src/main.ts apps/desktop/scripts/smoke-native-auth.mjs apps/desktop/scripts/desktop-shell-contract.test.mjs apps/web/app/(auth)/auth/native-return/page.tsx apps/web/app/(auth)/auth/native-return/page.test.tsx packages/auth-routing/index.ts packages/auth-routing/auth-routing.test.ts apps/desktop/electron-builder.staging.yml apps/desktop/electron-builder.local.yml` -> Checked 6 files. No fixes applied. Final rerun fixed 3 files, followed by passing tests.
- `pnpm --filter @jovie/auth-routing run typecheck && pnpm --filter @jovie/auth-routing test` -> tsc passed; 1 test file passed, 23 tests passed
- `pnpm --filter @jovie/web exec vitest run app/(auth)/auth/native-return/page.test.tsx lib/desktop/auth-return.test.ts` -> 2 test files passed, 14 tests passed
- `pnpm --filter @jovie/web run typecheck -- --pretty false` -> passed with no diagnostics
- `pnpm --filter @jovie/desktop run typecheck && pnpm --filter @jovie/desktop run test` -> tsc passed; desktop package test passed, including desktop-shell contract 10/10 and installed-app audit 4/4
- `pnpm --filter @jovie/web run test:bug-to-test` -> not applicable per script output
- `git diff --check` -> passed
- pre-commit hook -> gitleaks/trufflehog no leaks, proxy guard passed, lint-staged passed, web typecheck passed
- pre-push hook -> biome changed-files check passed, proxy guard passed, tailwind guard passed, web typecheck passed
- `JOVIE_AGENT_PROFILE=coder node scripts/desktop-release-guard.mjs --base origin/main` -> [desktop-release-guard] Desktop release handled by .github/workflows/desktop-release.yml.
[...]

PR #13029, opened 2026-07-04
fix(desktop): isolate Mac auth callback schemes

itstimwhite:
## Summary

Closes #12822.

- Route native desktop auth callbacks through env-specific schemes: `jovie://`, `jovie-staging://`, and `jovie-local://`.
- Derive the native-return deep-link protocol from the current web origin so staging/local do not transiently emit production `jovie://` callbacks.
- Register staging/local macOS bundle URL types and update the desktop native-auth smoke harness/contract tests for the new callback routing.
- Preserve the signed-out desktop recovery contract so passive auth redirects land on a usable sign-in surface instead of stranding the packaged window on `about:blank`.

## Verification

Local:

- `JOVIE_AGENT_PROFILE=coder ./scripts/setup.sh` -> ready to develop
- `JOVIE_AGENT_PROFILE=coder pnpm exec biome check --write packages/auth-routing/index.ts packages/auth-routing/auth-routing.test.ts apps/web/app/'(auth)'/auth/native-return/page.tsx apps/web/app/'(auth)'/auth/native-return/page.test.tsx apps/desktop/src/main.ts apps/desktop/scripts/smoke-native-auth.mjs apps/desktop/scripts/desktop-shell-contract.test.mjs apps/desktop/electron-builder.staging.yml apps/desktop/electron-builder.local.yml .github/workflows/desktop-release.yml`
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/auth-routing run typecheck && pnpm --filter @jovie/auth-routing test` -> 1 file / 23 tests passed
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/web exec vitest run app/'(auth)'/auth/native-return/page.test.tsx lib/desktop/auth-return.test.ts` -> 2 files / 14 tests passed
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/web run typecheck -- --pretty false` -> passed
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/desktop run prepare:assets`
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/desktop run typecheck && pnpm --filter @jovie/desktop run test` -> desktop package tests passed, including desktop shell contract 10/10 and installed-app audit 4/4
- `JOVIE_AGENT_PROFILE=coder pnpm --filter @jovie/web run test:bug-to-test` -> no implementation test candidates for bug-to-test contract
- `JOVIE_AGENT_PROFILE=coder node scripts/desktop-release-guard.mjs --base origin/main` -> desktop release handled by `.github/workflows/desktop-release.yml`
- `JOVIE_AGENT_PROFILE=coder node --test scripts/desktop-release-guard.test.mjs` -> 7 tests passed
- `JOVIE_AGENT_PROFILE=coder coderabbit review --agent -c AGENTS.md -t uncommitted` -> actionable findings fixed; final rerun hit CodeRabbit rate limit, GitHub CodeRabbit/AI review checks were non-blocking/green on the first PR run
- `git diff --check` -> passed
- pre-commit hooks -> passed
- pre-push hooks -> passed twice, including the replacement branch push

CI from the first active PR run (#13022 before GitHub closed it unmerged):

- Green: Build, Typecheck, Lint, Guardrails (proxy), Structural Contract, Migration Guard, Secret Scan, Neon DB, DB Migrate, Drizzle Check, unit shards, Lighthouse public/admin/chat/onboarding/dashboard, A11y, Mobile Overflow, Admin E2E Smoke, E2E DB Migrate, Preview Deploy, and PR Ready.
- Remaining CI blockers after rerun were preview/auth E2E jobs outside this desktop callback diff: `E2E Smoke (PR Fast Feedback)` and `Golden Path (PR)`; `Full E2E (Preview) (1)` was still in quarantine retry, so GitHub had not exposed the final failed-job logs yet.

## Subagent Review

- testing: verified required desktop shell/auth callback contracts and focused test plan.
- security: reviewed per-env scheme isolation and origin-derived deep-link guardrails.
- review: caught server/pre-hydration `globalThis.location.origin` risk; fixed and covered with a no-location render test.

### DIFF
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -26,7 +26,7 @@ on:
     branches: [main]
     paths:
       - "VERSION"
-      - ".github/workflows/desktop-release.yml"  # explicit handling for desktop auth allowlist fixes (e.g. 9870/9812 Clerk wildcard; JOV-3835 CSP-watchdog blank-window fix)
+      - ".github/workflows/desktop-release.yml"  # explicit handling for desktop auth allowlist/scheme fixes (e.g. 9870/9812 Clerk wildcard; JOV-3835 CSP-watchdog blank-window fix; GH-12822 per-env callback schemes)
   workflow_dispatch:
     inputs:
       environment:
--- a/apps/desktop/electron-builder.local.yml
+++ b/apps/desktop/electron-builder.local.yml
@@ -19,6 +19,10 @@ mac:
   extendInfo:
     NSAppTransportSecurity:
       NSAllowsArbitraryLoads: false
+    CFBundleURLTypes:
+      - CFBundleURLName: Jovie Local Auth
+        CFBundleURLSchemes:
+          - jovie-local
   notarize: false
   target:
     - target: dir
--- a/apps/desktop/electron-builder.staging.yml
+++ b/apps/desktop/electron-builder.staging.yml
@@ -19,6 +19,10 @@ mac:
   extendInfo:
     NSAppTransportSecurity:
       NSAllowsArbitraryLoads: false
+    CFBundleURLTypes:
+      - CFBundleURLName: Jovie Staging Auth
+        CFBundleURLSchemes:
+          - jovie-staging
   notarize: true
   target:
     - target: dmg
--- a/apps/desktop/scripts/desktop-shell-contract.test.mjs
+++ b/apps/desktop/scripts/desktop-shell-contract.test.mjs
@@ -115,6 +115,16 @@ test('desktop window fails into a branded Jovie recovery surface', async () => {
     mainSource,
     /hideMainWindowForAuthHandoff\(\);\s*if \(authHandoffWindow\) showWindow\(authHandoffWindow\);/
   );
+  assert.match(mainSource, /const initialVisibilityFallback = setTimeout/);
+  assert.match(mainSource, /current === 'about:blank'/);
+  assert.match(
+    mainSource,
+    /const authUrl = buildCentralDesktopAuthUrl\('sign_in', '\/app'\);/
+  );
+  assert.match(
+    mainSource,
+    /void win\.loadURL\(buildDesktopAuthHandoffUrl\(authUrl\)\);/
+  );
   assert.match(
     mainSource,
     /void win\.loadURL\(buildDesktopAuthHandoffUrl\(initialAuthUrl\)\);/
@@ -178,10 +188,19 @@ test('desktop production bundle declares the jovie auth protocol', async () => {
   assert.match(builderConfig, /CFBundleURLTypes:/);
   assert.match(builderConfig, /CFBundleURLName: Jovie Auth/);
   assert.match(builderConfig, /CFBundleURLSchemes:\s*\n\s*- jovie/);
-  assert.match(mainSource, /if \(APP_ENV !== 'production'\)/);
-  assert.match(mainSource, /setAsDefaultProtocolClient\('jovie'\)/);
-  assert.doesNotMatch(stagingConfig, /CFBundleURLTypes:/);
-  assert.doesNotMatch(localConfig, /CFBundleURLTypes:/);
+  assert.match(mainSource, /APP_ENV === 'staging'\s*\?\s*'jovie-staging'/);
+  assert.match(mainSource, /APP_ENV === 'local'\s*\?\s*'jovie-local'/);
+  assert.match(mainSource, /:\s*'jovie';/);
+  assert.match(
+    mainSource,
+    /app\.setAsDefaultProtocolClient\(AUTH_RETURN_SCHEME/
+  );
+  assert.match(stagingConfig, /CFBundleURLTypes:/);
+  assert.match(stagingConfig, /CFBundleURLName: Jovie Staging Auth/);
+  assert.match(stagingConfig, /CFBundleURLSchemes:\s*\n\s*- jovie-staging/);
+  assert.match(localConfig, /CFBundleURLTypes:/);
+  assert.match(localConfig, /CFBundleURLName: Jovie Local Auth/);
+  assert.match(localConfig, /CFBundleURLSchemes:\s*\n\s*- jovie-local/);
 });
 
 test('desktop navigation uses explicit URL disposition allowlists', async () => {
@@ -408,6 +427,21 @@ test('native auth smoke keeps browser callbacks on the browser auth origin', asy
     smokeSource,
     /const BASE_URL = process\.env\.BASE_URL \?\? 'http:\/\/localhost:3112';/
   );
+  assert.match(
+    smokeSource,
+    /const NATIVE_AUTH_CALLBACK_SCHEME = getNativeAuthSchemeForBaseUrl\(BASE_URL\);/
+  );
+  assert.match(
+    smokeSource,
+    /if \(hostname === 'staging\.jov\.ie'\) return 'jovie-staging';/
+  );
+  assert.match(smokeSource, /if \(hostname === 'jov\.ie'\) return 'jovie';/);
+  assert.match(
+    smokeSource,
+    /hostname === 'localhost'[\s\S]*return 'jovie-local';/
+  );
+  assert.match(smokeSource, /return 'jovie';\s*\n}/);
+  assert.match(smokeSource, /NATIVE_AUTH_CALLBACK_PREFIX/);
   assert.match(smokeSource, /async function waitForDesktopAuthHandoff/);
   assert.match(smokeSource, /state === 'opened'/);
   assert.match(
--- a/apps/desktop/scripts/smoke-native-auth.mjs
+++ b/apps/desktop/scripts/smoke-native-auth.mjs
@@ -14,17 +14,34 @@ const SMOKE_CLIENT_IP =
 const CLEAR_ELECTRON_AUTH_ON_START =
   process.env.SMOKE_CLEAR_ELECTRON_AUTH === '1';
 const SKIP_START_SIGN_OUT = process.env.SMOKE_SKIP_START_SIGNOUT === '1';
+const NATIVE_AUTH_CALLBACK_SCHEME = getNativeAuthSchemeForBaseUrl(BASE_URL);
 const REQUEST_TIMEOUT_MS = parsePositiveInteger(
   process.env.SMOKE_REQUEST_TIMEOUT_MS,
   180_000
 );
 const SMOKE_AUTH_EVIDENCE_KEY = 'jovie.desktopAuth.smokeAuthEvidence';
+const NATIVE_AUTH_CALLBACK_PREFIX = `${NATIVE_AUTH_CALLBACK_SCHEME}://auth/complete?`;
 
 let playwrightChromium;
 
 function parsePositiveInteger(value, fallback) {
   const parsed = Number.parseInt(value ?? '', 10);
   return Number.isFinite(parsed) && parsed > 0 ? parsed : fallback;
+}
+
+function getNativeAuthSchemeForBaseUrl(baseUrl) {
+  const hostname = new URL(baseUrl).hostname.toLowerCase();
+  if (hostname === 'jov.ie') return 'jovie';
+  if (hostname === 'staging.jov.ie') return 'jovie-staging';
+  if (
+    hostname === 'localhost' ||
+    hostname === '127.0.0.1' ||
+    hostname === '[::1]' ||
+    hostname.endsWith('.localhost')
+  ) {
+    return 'jovie-local';
+  }
+  return 'jovie';
 }
 
 function sleep(ms) {
@@ -663,7 +680,7 @@ async function requestRedirect(page, targetUrl, label) {
 
 async function requestNativeCallbackRedirect(page, callbackPath, label) {
   const redirectUrl = await requestRedirect(page, callbackPath, label);
-  if (redirectUrl.startsWith('jovie://auth/complete?')) {
+  if (redirectUrl.startsWith(NATIVE_AUTH_CALLBACK_PREFIX)) {
     return redirectUrl;
   }
 
@@ -694,7 +711,7 @@ function waitForNativeProtocolRequest(page, label, timeout = 60_000) {
 
     function onRequest(request) {
       const requestUrl = request.url();
-      if (!requestUrl.startsWith('jovie://auth/complete?')) return;
+      if (!requestUrl.startsWith(NATIVE_AUTH_CALLBACK_PREFIX)) return;
       clearTimeout(timeoutId);
       page.off('request', onRequest);
       resolve(requestUrl);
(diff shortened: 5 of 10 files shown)


itstimwhite:
We are given a PR diff and we need to check for logic errors, security issues, data integrity, error handling, and cross-file impact.

Let's go through the changes file by file and note any potential issues.

However, note that the PR description and the verification steps indicate that the changes have been tested and passed in various environments. We must be cautious and only flag issues that are clear and present in the diff.

We'll break down the changes:

1. **.github/workflows/desktop-release.yml**: Only a comment change in the paths section. No logic change.

2. **apps/desktop/electron-builder.local.yml**: Added CFBundleURLTypes for jovie-local. This is a configuration change for the Electron app. No logic in the sense of code, but we should check if it's correct.

3. **apps/desktop/electron-builder.staging.yml**: Similarly, added CFBundleURLTypes for jovie-staging.

4. **apps/desktop/scripts/desktop-shell-contract.test.mjs**: This test file is updated to check for the new environment-specific schemes and the updated main.ts logic. We see several new assertions. We must check if the test is correctly written and if it matches the implementation.

5. **apps/desktop/scripts/smoke-native-auth.mjs**: This file is updated to use a function `getNativeAuthSchemeForBaseUrl` to determine the scheme based on the base URL. It then uses that scheme to form the callback prefix and checks.

6. **apps/desktop/src/main.ts**: This is the main Electron process file. Changes include:
   - Defining `AUTH_RETURN_SCHEME` based on `APP_ENV` (staging, local, or production) and then `AUTH_RETURN_PROTOCOL` as the scheme with a colon.
   - Removing the condition that only registered the protocol in production and now registering it in all environments (but note: the condition was removed and now we always try to set the protocol client, but note that in development (unpacked) we have a condition to avoid setting it twice?).
   - In the single instance lock callback, the check for the URL now uses the dynamic `AUTH_RETURN_PROTOCOL` and `AUTH_RETURN_HOST`.

7. **apps/web/app/(auth)/auth/native-return/page.test.tsx**: Updated tests for the native return page. They now test for different origins (production, staging, local) and also a test for when location is not available.

8. **apps/web/app/(auth)/auth/native-return/page.tsx**: The native return page now uses a hook to get the protocol from the origin (via the auth-routing package) and then builds the deep link accordingly. It also handles the case when location is not available by delaying the protocol setting until after the effect runs.

9. **packages/auth-routing/auth-routing.test.ts**: Added tests for the new functions `buildElectronAuthCompleteUrl` (with protocol) and `getElectronAuthCompleteProtocolForOrigin`.

10. **packages/auth-routing/index.ts**: Added the type `ElectronAuthCompleteProtocol`, the function `getElectronAuthCompleteProtocolForOrigin`, and updated `buildElectronAuthCompleteUrl` to take an optional protocol.

### Logic Errors

- In `apps/desktop/src/main.ts`:
  - We removed the condition that prevented setting the protocol client in non-production environments. Now we set it in all environments. However, note that in development (when not packaged) we have:
        if (process.argv.length >= 2 && !app.isPackaged) {
          app.setAsDefaultProtocolClient(AUTH_RETURN_SCHEME, process.execPath, [
            path.resolve(process.argv[1]),
          ]);
          return;
        }
    and then outside we have:
        app.setAsDefaultProtocolClient(AUTH_RETURN_SCHEME);
    This is acceptable because we are setting it conditionally for development and then for production (and also for staged/local when packaged?).

    However, note that the condition `!app.isPackaged` is for development. In production (packaged) we skip the first block and set it without arguments (which uses the current executable). This is standard.

    But note: the condition we removed was:
        if (APP_ENV !== 'production') {
          return;
        }
    So now we are setting the protocol client in staging and local as well. This is intentional because we want each environment to have its own scheme.

    However, we must be cautious: setting the protocol client multiple times might cause issues? But the Electron documentation says that calling `setAsDefaultProtocolClient` multiple times is safe and will update the registration.

    Also note: in the development block, we are passing an array of arguments (the main script) but in the production-like block we are not. This is because in development we are running from a source file and we want to make sure the correct executable is used. In production, we are packaged and the executable is the app itself.

    This seems correct.

  - In the single instance lock callback, we have:
        if (url.startsWith(`${AUTH_RETURN_PROTOCOL}//${AUTH_RETURN_HOST}`)) {
    This is correct because we are now using the dynamic protocol and host.

- In `apps/desktop/scripts/smoke-native-auth.mjs`:
  - We added a function `getNativeAuthSchemeForBaseUrl` that returns the scheme based on the hostname of the base URL.
  - We then use that to form `NATIVE_AUTH_CALLBACK_PREFIX` and use it in several places.

  Let's check the function:
        function getNativeAuthSchemeForBaseUrl(baseUrl) {
          const hostname = new URL(baseUrl).hostname.toLowerCase();
          if (hostname === 'jov.ie') return 'jovie';
          if (hostname === 'staging.jov.ie') return 'jovie-staging';
          if (
            hostname === 'localhost' ||
            hostname === '127.0.0.1' ||
            hostname === '[::1]' ||
            hostname.endsWith('.localhost')
          ) {
            return 'jovie-local';
          }
          return 'jovie';
        }

  This function is used to determine the scheme for the native auth callback. It matches the logic in the Electron main process for determining the scheme (based on APP_ENV) but note: in the main process we use APP_ENV (an environment variable) while here we use the hostname of the base URL.

  This is intentional because the smoke test runs against a running web instance (which has a base URL) and we want to mimic the same logic that the web page uses to determine the protocol (which is based on the origin).

  However, note that the web page uses `getElectronAuthCompleteProtocolForOrigin` which is defined in the auth-routing package and uses the same logic.

  So this is consistent.

  - We then use `NATIVE_AUTH_CALLBACK_PREFIX` in:
        if (redirectUrl.startsWith(NATIVE_AUTH_CALLBACK_PREFIX)) {
        and similarly in the request and response listeners.

  This is correct.

- In `apps/web/app/(auth)/auth/native-return/page.tsx`:
  - We are using `useState` and `useEffect` to set the protocol when the component mounts, but only if `globalThis.location` is available.
  - We then build the deep link only when we have both the native return params and the protocol.
  - We also have a useEffect that attempts to redirect if we have a deep link and `globalThis.location` is available.

  However, note that we are setting the protocol in an effect that runs once (because of the empty dependency array). But what if the origin changes? It shouldn't because the page is mounted at a fixed origin.

  Also, note that we are checking for `globalThis.location` in the effect and in the redirect effect. This is to avoid errors in environments where `location` is not available (like during SSR or in tests).

  We also added a test for when location is not available.

  This seems correct.

  - One potential issue: the `useEffect` that sets the protocol runs only once. What if the component is re-mounted? It will run again. But that's acceptable.

  - Another: we are setting the protocol state and then using it in a `useMemo` for the deep link. This is correct.

### Security

- We are changing the deep link scheme based on the origin. This prevents staging and local from accidentally using the production scheme, which could cause a staging/local app to try to handle a production deep link (or vice versa). This is a security improvement because it isolates the environments.

- We are not introducing any hardcoded secrets.

- We are not changing any authentication or authorization logic in a way that would bypass checks.

- The changes are confined to URL scheme handling and do not introduce XSS vectors because we are not inserting user input into the DOM without escaping. The deep link is built from parameters that are validated (code and state) and then used in an `href` attribute. The `Link` component from Next.js will handle the href safely.

### Data Integrity

- No database changes are in this PR.

### Error Handling

- In `apps/desktop/src/main.ts`: We removed the early return for non-production, so we now attempt to set the protocol client in all environments. This could potentially throw an error if there's a problem (e.g., the protocol is already registered by another app). However, the Electron `setAsDefaultProtocolClient` method does not throw if it fails; it returns a boolean. We are not checking the return value. But note: the original code didn't check either. So this is not a regression.

- In the same file, in the single instance lock callback, we are checking the URL and then returning early if it doesn't match. We are not handling the case where the URL matches but then we fail to restore the window? But that's beyond the scope of this change.

- In `apps/desktop/scripts/smoke-native-auth.mjs`: We are using the `getNativeAuthSchemeForBaseUrl` function. If the baseUrl is invalid, the `new URL(baseUrl)` will throw. However, the baseUrl comes from an environment variable (BASE_URL) that is set in the smoke test. We assume it is a valid URL. If it's not, the test will fail early. This is acceptable because it's a test script.

- In the web page: We are using `new URL(origin)` in the auth-routing function. If the origin is invalid, we return the default protocol. This is safe.

### Cross-file Impact

- We changed the interface of `buildElectronAuthCompleteUrl` in the auth-routing package to accept an optional `protocol`. We updated the call in the web page to pass the protocol we derived from the origin.

- We updated the tests in the auth-routing package to cover the new protocol argument and the new function `getElectronAuthCompleteProtocolForOrigin`.

- We updated the web page tests to cover the different origins and the case when location is not available.

- We updated the desktop main.ts to use the new protocol variable.

- We updated the smoke test to use the new scheme detection function.

All these changes are consistent.

### Specific Issues to Check

Let's look for any specific lines that might be problematic:

1. In `apps/desktop/src/main.ts`:
   - Line 94: `const AUTH_RETURN_SCHEME = ...` - This is fine.
   - Line 102: `const AUTH_RETURN_PROTOCOL = `${AUTH_RETURN_SCHEME}:`; - This is fine.
   - Line 1410: `app.setAsDefaultProtocolClient(AUTH_RETURN_SCHEME, process.execPath, [ ... ]);` - In the development block, we are passing an array with one element: the resolved path of the first argument. This is correct for development.
   - Line 1413: `app.setAsDefaultProtocolClient(AUTH_RETURN_SCHEME);` - In the production-like block, we are setting it without arguments. This is correct for when we are packaged.

   However, note that in the development block we return early, so the production-like block is not run in development. This is correct.

   But wait: what about when we are running in a staged or local packaged app? We want to set the protocol client. The condition `!app.isPackaged` will be false, so we will go to the production-like block and set it without arguments. This is correct because in a packaged app (whether production, staging, or local) we are running from the executable and we don't need to specify the path.

   However, note that the staging and local builds are also packaged (they produce a .dmg or a dir). So this is correct.

2. In `apps/desktop/scripts/smoke-native-auth.mjs`:
   - We defined `NATIVE_AUTH_CALLBACK_SCHEME` at the top, but note that we use it in the function `getNativeAuthSchemeForBaseUrl` which is defined later. However, the function is called after its definition? Actually, we define the function and then we use `NATIVE_AUTH_CALLBACK_SCHEME` in the top-level code. But wait:

        const NATIVE_AUTH_CALLBACK_SCHEME = getNativeAuthSchemeForBaseUrl(BASE_URL);

   This line is at the top, and the function `getNativeAuthSchemeForBaseUrl` is defined later. This will cause a ReferenceError because we are calling the function before it is defined.

   Let me check the diff:

        const NATIVE_AUTH_CALLBACK_SCHEME = getNativeAuthSchemeForBaseUrl(BASE_URL);
        const REQUEST_TIMEOUT_MS = parsePositiveInteger(
          process.env.SMOKE_REQUEST_TIMEOUT_MS,
          180_000
        );

        function getNativeAuthSchemeForBaseUrl(baseUrl) {
          ...
        }

   This is indeed a problem: we are trying to call `getNativeAuthSchemeForBaseUrl` before it is defined.

   We must move the function definition above the const assignment, or we must change the order.

   This is a logic error because the script will fail to run.

   Let's verify the exact lines from the diff:

   The diff shows:

        const SMOKE_CLIENT_IP =
        const CLEAR_ELECTRON_AUTH_ON_START =
        const SKIP_START_SIGN_OUT = ...
        const NATIVE_AUTH_CALLBACK_SCHEME = getNativeAuthSchemeForBaseUrl(BASE_URL);
        const REQUEST_TIMEOUT_MS = ...
        const SMOKE_AUTH_EVIDENCE_KEY = ...
        const NATIVE_AUTH_CALLBACK_PREFIX = ... // uses NATIVE_AUTH_CALLBACK_SCHEME

        function parsePositiveInteger(...) { ... }

        function getNativeAuthSchemeForBaseUrl(baseUrl) { ... }

   So the function `getNativeAuthSchemeForBaseUrl` is defined after the const that uses it.

   This will cause a ReferenceError when the script is run.

   We must fix this by moving the function definition above the const assignment.

   Alternatively, we can change the order of the const and the function.

   Since the function does not depend on any of the consts above, we can move it up.

   This is a critical bug that would break the smoke test.

3. In `apps/desktop/src/main.ts`:
   - We removed the condition that prevented setting the protocol client in non-production. However, note that the condition we removed was:

        if (APP_ENV !== 'production') {
          return;
        }

     and we replaced it with nothing. Now we always try to set the protocol client.

     But note: in the development block (unpacked) we have:

        if (process.argv.length >= 2 && !app.isPackaged) {
          ... // set with arguments and return
        }

     and then we have the production-like block.

     This means that in development we set it with the arguments and return, so we don't reach the production-like block.

     In production (packaged) we skip the development block and set it in the production-like block.

     In staging and local (when packaged) we also skip the development block and set it in the production-like block.

     This is correct.

   However, note that the original condition was meant to prevent staging and local from setting the production protocol. Now we are setting the environment-specific protocol in all environments (including development) but note:

     In development, we are setting the protocol with the arguments and returning early. We are not setting it again in the production-like block because we returned.

     But what protocol are we setting in development? We are setting `AUTH_RETURN_SCHEME` which is determined by `APP_ENV`. In development, `APP_ENV` is likely 'local' (if we are running locally) or maybe set to something else? We don't see the setting of APP_ENV in this file, but it is likely set by the build process.

     However, note that the smoke test and the desktop app are built with different configurations. The development mode (when running `electron .`) would have APP_ENV set to 'local' (if we are running locally) and then we would set the scheme to 'jovie-local' in the development block.

     This is consistent.

   But wait: the development block does not use the `APP_ENV` to determine the scheme? Actually, it does because `AUTH_RETURN_SCHEME` is computed at the top of the file.

   So in development, we set the scheme to whatever `APP_ENV` is (which should be 'local' for a local build) and then we set the protocol client for that scheme.

   This is correct.

   However, note that the condition we removed was specifically to avoid setting the protocol client in non-production. Now we are setting it in non-production as well, but we are setting the correct scheme for the environment.

   This is the intended change.

   But we must ensure that setting the protocol client multiple times (for example, if the app is run multiple times) does not cause issues. The Electron documentation says it is safe.

   So this is not an issue.

4. In `apps/web/app/(auth)/auth/native-return/page.test.tsx`:
   - We added a test that deletes the `globalThis.location` property and then renders the component. We expect that no link is rendered and that the render does not throw.

   However, note that in the component we have:

        useEffect(() => {
          if (!globalThis.location) return;
          setProtocol(
            getElectronAuthCompleteProtocolForOrigin(globalThis.location.origin)
          );
        }, []);

   If `globalThis.location` is not available, we skip the effect and the protocol remains null.

   Then, in the `useMemo` for the deep link, we return null if either the nativeReturnParams or the protocol is null.

   Then, in the redirect effect:

        useEffect(() => {
          if (deepLink && globalThis.location) {
            globalThis.location.href = deepLink;
          }
        }, [deepLink]);

   So if there's no deepLink (because protocol is null) we don't redirect.

   And in the render, we conditionally render the link only if we have a deepLink (which we don't) so we render the recovery message.

   The test expects that the link is null, which is

itstimwhite (commented):

itstimwhite:
Closing as superseded: the keeper branch for #12822 rebases to an empty diff against current main, so the desktop auth callback fix is already upstream. Removing this duplicate from the merge queue.

PR #12937, opened 2026-07-03
chore(codex): feat(design): canonical ConfirmDialog for destructive actions (#12886)

itstimwhite:
Closes #12886

Opened by the codex-issue-shipper **deterministic finisher**: the coding
agent produced this diff but exited without opening a PR. Pre-commit hooks
(lint-staged, typecheck) passed at commit time; CI is the merge gate as
usual.

Agent log: `/Users/timwhite/.hermes/profiles/summer/logs/codex-issue-shipper/github-12886-2026-07-03T15-35-08-654z.log`

PR #12908, opened 2026-07-03
chore(codex): feat(design): canonical ConfirmDialog for destructive actions (#12886)

itstimwhite:
Closes #12886

Opened by the codex-issue-shipper **deterministic finisher**: the coding
agent produced this diff but exited without opening a PR. Pre-commit hooks
(lint-staged, typecheck) passed at commit time; CI is the merge gate as
usual.

Agent log: `/Users/timwhite/.hermes/profiles/summer/logs/codex-issue-shipper/github-12886-2026-07-03T12-13-13-782z.log`

PR #13035, opened 2026-07-04
feat(ui): add canonical destructive ConfirmDialog

### LINKED ISSUE #12886: feat(design): canonical ConfirmDialog for destructive actions
## Summary

Canonical confirmation dialog for destructive actions. Same shape every time:
- Title (the action, e.g. "Delete this release?")
- Body (consequence in plain language)
- Cancel (neutral) + Confirm (destructive, red)

Used for: delete account, delete release, remove contact, leave team, disconnect integration, etc.

## Why this matters

Some destructive actions confirm, some don't. Some confirm dialogs use brand colors, some use red. The user can't predict which is which. Destructive actions must always be visually destructive.

## Acceptance

- [ ] `ConfirmDialog` component in `@jovie/ui` with destructive variant
- [ ] Title + body + cancel + confirm slots
- [ ] Confirm button uses `JovieColor.errorText` for destructive
- [ ] Esc closes (cancel)
- [ ] Click outside closes (cancel)
- [ ] Focus trap inside dialog
- [ ] Migrate 6+ destructive flows: delete release, remove contact, disconnect platform, leave team, delete account, cancel subscription
- [ ] Add audit for any destructive action missing a confirm

## Affected files

- `packages/ui/confirm-dialog.tsx` (new)
- All destructive action handlers in `apps/web/app/`

## Visual mock

See shootout proposal: http://localhost:7788/ (card id: `confirm-dialog`)

## Context

- Approved 2026-07-03 via Design Shootout
- Decisions: `~/.ovie/decisions.json`

itstimwhite:
## Summary
- Add canonical `ConfirmDialog` to `@jovie/ui` with title/body/cancel/confirm slots, destructive confirm styling, Escape cancel, outside-click cancel, focus trap, disabled confirm, async pending, and visible failure fallback.
- Re-export the app `ConfirmDialog` wrapper from `@jovie/ui` for compatibility, then migrate delete account and cancel subscription off raw `AlertDialog` while preserving account `DELETE` typed confirmation.
- Add a destructive-confirm audit covering the required destructive flows and blocking new raw destructive `AlertDialogAction` usage outside the documented legacy allowlist.

Closes #12886

## Verification
- `pnpm install --frozen-lockfile`
- `pnpm biome check --write packages/ui/confirm-dialog.tsx packages/ui/confirm-dialog.test.tsx packages/ui/index.ts apps/web/components/molecules/ConfirmDialog.tsx apps/web/components/organisms/billing/BillingActionsSection.tsx apps/web/components/features/dashboard/organisms/DataPrivacySection.tsx apps/web/tests/unit/design-system/destructive-confirm-dialog-audit.test.ts` -> `Checked 7 files ...`
- `pnpm --filter @jovie/ui test -- confirm-dialog` -> `Test Files 1 passed`, `Tests 6 passed`
- `pnpm --filter @jovie/ui run typecheck` -> `tsc --noEmit` passed
- `pnpm --filter @jovie/web exec vitest run --config=vitest.config.mts tests/unit/design-system/destructive-confirm-dialog-audit.test.ts` -> `Test Files 1 passed`, `Tests 2 passed`
- `pnpm --filter @jovie/web run lint:no-native-dialogs` -> `No native browser dialogs found.`
- `pnpm --filter @jovie/web exec vitest run --config=vitest.config.mts tests/components/dashboard/LinkActions.negative.critical.test.tsx tests/components/dashboard/LinkActions.click.test.tsx tests/components/dashboard/LinkActions.keyboard.test.tsx` -> `Test Files 3 passed`, `Tests 19 passed`
- `pnpm --filter @jovie/web run typecheck -- --pretty false` -> passed
- `pnpm --filter @jovie/web run test:bug-to-test` -> `bug-to-test: not applicable (not a bug fix PR)`
- Pre-commit hook -> gitleaks/trufflehog clean, next proxy guard passed, staged Biome/a11y/ESLint/typecheck passed
- Pre-push hook -> changed-file Biome passed, next proxy guard passed, Tailwind guard passed, web typecheck passed
- `coderabbit review --agent -c AGENTS.md -t uncommitted` -> first run completed with findings; fixed actionable major/minor items. Required rerun attempted twice after diff changes, but CodeRabbit returned `rate_limit` with 45-46 minute wait.

## QA Notes
- Layout/state coverage: component tests cover closed/open rendering, destructive confirm token path, Escape cancel, outside-click cancel, focus trap, disabled confirm, and failed async confirm staying open with visible error.
- Delete account keeps the typed `DELETE` confirmation and only signs out after `mutateAsync` succeeds.
- Cancel subscription now uses canonical destructive confirm copy (`Cancel Subscription`) and token-based destructive styling instead of ad hoc red classes.

## Existing Related PRs
This issue already had earlier open attempts (#12908 and #12937). This PR is from the requested 2026-07-04 worktree/branch and includes the migration plus audit/test evidence for this run.

### DIFF
--- a/apps/web/components/features/dashboard/organisms/DataPrivacySection.tsx
+++ b/apps/web/components/features/dashboard/organisms/DataPrivacySection.tsx
@@ -1,18 +1,9 @@
 'use client';
 
-import {
-  AlertDialog,
-  AlertDialogCancel,
-  AlertDialogContent,
-  AlertDialogDescription,
-  AlertDialogFooter,
-  AlertDialogHeader,
-  AlertDialogTitle,
-  Button,
-  Input,
-} from '@jovie/ui';
+import { Button, Input } from '@jovie/ui';
 import { Download, Trash2 } from 'lucide-react';
 import { useCallback, useState } from 'react';
+import { ConfirmDialog } from '@/components/molecules/ConfirmDialog';
 import { SettingsActionRow } from '@/components/molecules/settings/SettingsActionRow';
 import { SettingsPanel } from '@/components/molecules/settings/SettingsPanel';
 import { useAuthSafe } from '@/hooks/useClerkSafe';
@@ -30,18 +21,16 @@ export function DataPrivacySection() {
     exportMutation.mutate();
   }, [exportMutation]);
 
-  const handleDelete = useCallback(() => {
+  const handleDelete = useCallback(async () => {
     if (confirmText !== 'DELETE') return;
-    deleteMutation.mutate(
-      { confirmation: 'DELETE' },
-      {
-        onSuccess: () => {
-          setDeleteDialogOpen(false);
-          // Sign out to clear session cookies, then redirect to home
-          signOut({ redirectUrl: '/' });
-        },
-      }
-    );
+    try {
+      await deleteMutation.mutateAsync({ confirmation: 'DELETE' });
+      setDeleteDialogOpen(false);
+      // Sign out to clear session cookies, then redirect to home.
+      signOut({ redirectUrl: '/' });
+    } catch {
+      // useDeleteAccountMutation surfaces the failure toast.
+    }
   }, [confirmText, deleteMutation, signOut]);
 
   return (
@@ -78,64 +67,45 @@ export function DataPrivacySection() {
                 onClick={() => setDeleteDialogOpen(true)}
                 className='w-full shrink-0 sm:w-auto'
               >
-                Delete account
+                Delete Account
               </Button>
             }
           />
         </div>
       </SettingsPanel>
 
-      <AlertDialog
+      <ConfirmDialog
         open={deleteDialogOpen}
         onOpenChange={open => {
           setDeleteDialogOpen(open);
           if (!open) setConfirmText('');
         }}
+        title='Delete your account?'
+        description='This will permanently delete your account, profile, links, contacts, and all associated data. You will be signed out immediately. This action cannot be undone.'
+        cancelLabel='Cancel'
+        confirmLabel='Permanently Delete Account'
+        variant='destructive'
+        onConfirm={handleDelete}
+        isLoading={deleteMutation.isPending}
+        confirmDisabled={confirmText !== 'DELETE'}
+        onCancel={() => setConfirmText('')}
       >
-        <AlertDialogContent>
-          <AlertDialogHeader>
-            <AlertDialogTitle>Delete your account?</AlertDialogTitle>
-            <AlertDialogDescription>
-              This will permanently delete your account, profile, links,
-              contacts, and all associated data. You will be signed out
-              immediately. This action cannot be undone.
-            </AlertDialogDescription>
-          </AlertDialogHeader>
-          <div className='py-4'>
-            <label
-              htmlFor='delete-confirm'
-              className='mb-2 block text-app text-secondary-token'
-            >
-              Type <strong>DELETE</strong> to confirm
-            </label>
-            <Input
-              id='delete-confirm'
-              value={confirmText}
-              onChange={e => setConfirmText(e.target.value)}
-              placeholder='DELETE'
-              autoComplete='off'
-            />
-          </div>
-          <AlertDialogFooter>
-            <AlertDialogCancel
-              onClick={() => {
-                setConfirmText('');
-              }}
-            >
-              Cancel
-            </AlertDialogCancel>
-            <Button
-              variant='destructive'
-              onClick={handleDelete}
-              disabled={confirmText !== 'DELETE' || deleteMutation.isPending}
-            >
-              {deleteMutation.isPending
-                ? 'Deleting...'
-                : 'Permanently Delete Account'}
-            </Button>
-          </AlertDialogFooter>
-        </AlertDialogContent>
-      </AlertDialog>
+        <div className='py-4'>
+          <label
+            htmlFor='delete-confirm'
+            className='mb-2 block text-app text-secondary-token'
+          >
+            Type <strong>DELETE</strong> to confirm
+          </label>
+          <Input
+            id='delete-confirm'
+            value={confirmText}
+            onChange={e => setConfirmText(e.target.value)}
+            placeholder='DELETE'
+            autoComplete='off'
+          />
+        </div>
+      </ConfirmDialog>
     </div>
   );
 }
--- a/apps/web/components/molecules/ConfirmDialog.tsx
+++ b/apps/web/components/molecules/ConfirmDialog.tsx
@@ -1,29 +1,3 @@
-'use client';
-
-import {
-  AlertDialog,
-  AlertDialogAction,
-  AlertDialogCancel,
-  AlertDialogContent,
-  AlertDialogDescription,
-  AlertDialogFooter,
-  AlertDialogHeader,
-  AlertDialogTitle,
-} from '@jovie/ui';
-import { useRef, useState } from 'react';
-
-export interface ConfirmDialogProps {
-  readonly open: boolean;
-  readonly onOpenChange: (open: boolean) => void;
-  readonly title: string;
-  readonly description: string;
-  readonly confirmLabel?: string;
-  readonly cancelLabel?: string;
-  readonly variant?: 'default' | 'destructive';
-  readonly onConfirm: () => void | Promise<void>;
-  readonly isLoading?: boolean;
-}
-
 /**
  * Reusable confirmation dialog. Drop-in replacement for native `confirm()`.
  *
@@ -43,55 +17,5 @@ export interface ConfirmDialogProps {
  *   - DESIGN.md → "Confirmations & Destructive Actions" for copy rules and decision table
  *   - AGENTS.md → "4f. No Native Browser Dialogs" for the lint policy
  */
-export function ConfirmDialog({
-  open,
-  onOpenChange,
-  title,
-  description,
-  confirmLabel = 'Confirm',
-  cancelLabel = 'Cancel',
-  variant = 'default',
-  onConfirm,
-  isLoading = false,
-}: ConfirmDialogProps) {
-  const [isPending, setIsPending] = useState(false);
-  const handlingRef = useRef(false);
-
-  const handleConfirm = async () => {
-    if (handlingRef.current) return;
-    handlingRef.current = true;
-    setIsPending(true);
-    try {
-      await onConfirm();
-      onOpenChange(false);
-    } finally {
-      setIsPending(false);
-      handlingRef.current = false;
-    }
-  };
-
-  const loading = isLoading || isPending;
-
-  return (
-    <AlertDialog open={open} onOpenChange={onOpenChange}>
-      <AlertDialogContent>
-        <AlertDialogHeader>
-          <AlertDialogTitle>{title}</AlertDialogTitle>
-          <AlertDialogDescription>{description}</AlertDialogDescription>
-        </AlertDialogHeader>
-        <AlertDialogFooter>
-          <AlertDialogCancel disabled={loading}>
-            {cancelLabel}
-          </AlertDialogCancel>
-          <AlertDialogAction
-            onClick={handleConfirm}
-            disabled={loading}
-            variant={variant === 'destructive' ? 'destructive' : 'primary'}
-          >
-            {loading ? 'Please wait...' : confirmLabel}
-          </AlertDialogAction>
-        </AlertDialogFooter>
-      </AlertDialogContent>
-    </AlertDialog>
-  );
-}
+export type { ConfirmDialogProps } from '@jovie/ui';
+export { ConfirmDialog } from '@jovie/ui';
--- a/apps/web/components/organisms/billing/BillingActionsSection.tsx
+++ b/apps/web/components/organisms/billing/BillingActionsSection.tsx
@@ -1,20 +1,10 @@
 'use client';
 
-import {
-  AlertDialog,
-  AlertDialogAction,
-  AlertDialogCancel,
-  AlertDialogContent,
-  AlertDialogDescription,
-  AlertDialogFooter,
-  AlertDialogHeader,
-  AlertDialogTitle,
-  AlertDialogTrigger,
-  Button,
-} from '@jovie/ui';
+import { Button } from '@jovie/ui';
 import { CreditCard, XCircle } from 'lucide-react';
 import { motion } from 'motion/react';
 import { BillingPortalLink } from '@/components/molecules/BillingPortalLink';
+import { ConfirmDialog } from '@/components/molecules/ConfirmDialog';
 import { ContentSectionHeader } from '@/components/molecules/ContentSectionHeader';
 import { ContentSurfaceCard } from '@/components/molecules/ContentSurfaceCard';
 import { track } from '@/lib/analytics';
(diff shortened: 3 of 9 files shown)


itstimwhite:
We are given a diff for PR #13035. We need to check for logic errors, security issues, data integrity issues, error handling issues, and cross-file impact.

1. `apps/web/components/features/dashboard/organisms/DataPrivacySection.tsx`
   - Changes: Replaced `AlertDialog` with `ConfirmDialog` and adjusted the delete account flow.
   - Key changes:
     - Removed imports of AlertDialog components and added import of `ConfirmDialog` from '@/components/molecules/ConfirmDialog'.
     - Changed `handleDelete` from a callback that calls `mutate` to an async function that uses `mutateAsync` and handles errors with a try/catch.
     - In the JSX, replaced the `AlertDialog` with `ConfirmDialog` and adjusted the props.
     - The `ConfirmDialog` now has:
        - `title`, `description`, `cancelLabel`, `confirmLabel`, `variant='destructive'`, `onConfirm={handleDelete}`, `isLoading={deleteMutation.isPending}`, `confirmDisabled={confirmText !== 'DELETE'}`, and `onCancel={() => setConfirmText('')}`.
     - Inside the `ConfirmDialog`, we have the same input for confirmation text.

   Potential issues:
     - The `onCancel` prop in `ConfirmDialog` is set to `() => setConfirmText('')`. This is acceptable because when the dialog is closed by cancel (either by clicking cancel, outside, or Escape), we want to reset the confirmation text.
     - However, note that in the `onOpenChange` handler we also set `setConfirmText('')` when the dialog is closed. This is done in two places: 
          onOpenChange: {open => {
            setDeleteDialogOpen(open);
            if (!open) setConfirmText('');
          }}
        and the `onCancel` prop of `ConfirmDialog` also sets the text to empty.
     - This might lead to setting the state twice, but it's harmless because setting state to the same value is a no-op in React (if using useState with the same value). However, it's redundant.

     But note: the `ConfirmDialog` component from `@jovie/ui` (which is the one we are using) does not have an `onCancel` prop in the version we are migrating to? Actually, looking at the new `ConfirmDialog` in `packages/ui/confirm-dialog.tsx`, it does have an `onCancel` prop.

     However, in the `DataPrivacySection` we are using the `ConfirmDialog` from `@/components/molecules/ConfirmDialog` (which is the old one that we are about to replace). Wait, note the change:

        In `DataPrivacySection`:
          Before: imported AlertDialog from '@jovie/ui'
          After:  imported { ConfirmDialog } from '@/components/molecules/ConfirmDialog'

        And then we are replacing the AlertDialog with that ConfirmDialog.

        But note: the `ConfirmDialog` in `apps/web/components/molecules/ConfirmDialog.tsx` is being replaced by a re-export of the one from `@jovie/ui`. So after this change, the `ConfirmDialog` in `apps/web/components/molecules/ConfirmDialog.tsx` is just a re-export of the new one from the UI package.

        Therefore, the `ConfirmDialog` we are using in `DataPrivacySection` is the new one from `@jovie/ui` (via the molecule).

     - The new `ConfirmDialog` in the UI package has an `onCancel` prop. So the usage in `DataPrivacySection` is correct.

     - However, note that the `onOpenChange` handler in `DataPrivacySection` also sets the confirm text to empty when the dialog is closed. And the `onCancel` prop of the `ConfirmDialog` also does the same. This is redundant but not harmful.

     - Another point: the `handleDelete` function is now async and uses `mutateAsync`. We catch any error and do nothing (the comment says the mutation surfaces the failure toast). This is acceptable because the mutation likely has its own error handling (toast). However, note that the `ConfirmDialog` also has an `onError` prop that we are not using. We are relying on the mutation's error handling.

     - The `ConfirmDialog` in the UI package does have an `onError` prop and will show an error message if the `onConfirm` function throws and we provide an `onError` handler. But in this case, we are not providing one. So if the mutation throws, the dialog will show an error message (because the UI package's `ConfirmDialog` sets an error message and shows it) and then the dialog remains open (because we don't call `onOpenChange(false)` on error). This is the desired behavior.

     - However, note that in the `handleDelete` function we are catching the error and doing nothing. Then the `ConfirmDialog` will catch the error (because the `onConfirm` function is the `handleDelete` function, which throws) and then show the error message. But wait: we are catching the error in `handleDelete` and not re-throwing. So the `ConfirmDialog` will not see an error? 

        Let's look:

          const handleDelete = useCallback(async () => {
            if (confirmText !== 'DELETE') return;
            try {
              await deleteMutation.mutateAsync({ confirmation: 'DELETE' });
              setDeleteDialogOpen(false);
              signOut({ redirectUrl: '/' });
            } catch {
              // useDeleteAccountMutation surfaces the failure toast.
            }
          }, [confirmText, deleteMutation, signOut]);

        We catch the error and do nothing. Then the function returns normally (without throwing). So the `ConfirmDialog` will not see an error and will close the dialog (because we set `setDeleteDialogOpen(false)` only on success?).

        Actually, note: we only set `setDeleteDialogOpen(false)` and sign out on success. On error, we catch and do nothing, so the dialog remains open? But we don't set the open state to false on error. However, the `ConfirmDialog` will set `isPending` to false (because the loading state is set by the mutation) and then the dialog remains open? 

        But wait: the `ConfirmDialog` in the UI package does not automatically close on error. It only closes on success (when we call `onOpenChange(false)`). In our `handleDelete`, we only call `setDeleteDialogOpen(false)` on success. On error, we do nothing, so the dialog remains open. And the UI package's `ConfirmDialog` will show an error message because we set the error state in the `ConfirmDialog` when the `onConfirm` function throws? 

        However, note: we caught the error and did not re-throw. So the `ConfirmDialog` does not see an error. Therefore, it will not show an error message and will reset the loading state. But the dialog remains open because we never set the open state to false.

        This is a problem: on error, we want to show an error and keep the dialog open. But we are swallowing the error and then the dialog remains open without any error indication.

        The comment says: "useDeleteAccountMutation surfaces the failure toast." So we are relying on the mutation to show a toast. But the dialog remains open and the user might not notice the toast? Also, the loading state is reset (because we set `setIsPending(false)` in the `ConfirmDialog` when the `onConfirm` function finishes, even if we caught the error) but the dialog is still open.

        We should either:
          - Not catch the error in `handleDelete` and let the `ConfirmDialog` handle it (by providing an `onError` handler or letting it show the default error message) OR
          - If we want to handle the error ourselves (toast) then we should re-throw so that the `ConfirmDialog` can show its error fallback? But note the `ConfirmDialog` in the UI package has an `onError` prop and also sets an internal error state to show a message.

        Looking at the UI package's `ConfirmDialog`:

          const handleConfirm = async (event: MouseEvent<HTMLButtonElement>) => {
            event.preventDefault();
            if (handlingRef.current || loading || confirmDisabled) return;
            handlingRef.current = true;
            setIsPending(true);
            setErrorMessage(null);
            try {
              await onConfirm();
              onOpenChange(false);
            } catch (error) {
              onError?.(error);
              setErrorMessage('Action failed. Please try again.');
            } finally {
              setIsPending(false);
              handlingRef.current = false;
            }
          };

        So if we do not throw, then the `catch` block is not run and no error message is set. The dialog will close only if we call `onOpenChange(false)` (which we do on success). On error (if we catch and don't throw) we don't call `onOpenChange(false)` and we don't set an error message. So the dialog remains open in a loading state? No, because we set `setIsPending(false)` in the finally block. So the loading state is reset, but the dialog remains open and there is no error message.

        This is confusing for the user: they click the button, see "Please wait...", then the button goes back to normal and the dialog is still open with no indication of failure.

        We should either:
          Option 1: Remove the try/catch in `handleDelete` and let the error propagate to the `ConfirmDialog` so it can show the error message and keep the dialog open (without closing). Then we can also show a toast in the `onError` handler of the `ConfirmDialog` if we want.

          Option 2: If we want to handle the error with a toast and then keep the dialog open, we should re-throw after showing the toast? But note: the `ConfirmDialog` will show its own error message. We don't want two error messages.

        The current approach in the mutation (toast) might be sufficient, but then we are not showing any error in the dialog. The user might think it succeeded because the dialog is still open and the button is not loading? Actually, the button label goes back to normal and the dialog is open. That is confusing.

        Let's look at the verification: the test in `packages/ui/confirm-dialog.test.tsx` has a test for error:

          it('keeps the dialog open and shows an error when confirm fails', async () => {
            ... 
            fireEvent.click(
              screen.getByRole('button', { name: 'Cancel Subscription' })
            );

            expect(await screen.findByRole('alert')).toHaveTextContent(
              'Action failed. Please try again.'
            );
            expect(onError).toHaveBeenCalled();
            expect(onOpenChange).not.toHaveBeenCalledWith(false);
          });

        So the expected behavior is to show an error message in the dialog and not close it.

        Therefore, we should not catch the error in `handleDelete` and let the `ConfirmDialog` handle it. We can then use the `onError` prop of the `ConfirmDialog` to show a toast if we want, but note the test above does not use `onError` and still expects the alert to be shown. The `ConfirmDialog` sets an internal error message and shows it.

        Alternatively, we can remove the try/catch and then use the `onError` prop to show a toast and let the `ConfirmDialog` show its own message? That would be duplicate.

        The `ConfirmDialog` already shows: "Action failed. Please try again." So we don't need to show a toast necessarily. But if we want to show a toast in addition, we can use the `onError` prop.

        However, note the comment: "useDeleteAccountMutation surfaces the failure toast." So the mutation already shows a toast. We don't want to show two toasts.

        Therefore, we can remove the try/catch and let the error propagate. Then the `ConfirmDialog` will show its internal error message and the mutation will show a toast. That would be two notifications: one toast and one inline error. That might be acceptable, but we can avoid the toast in the mutation if we don't want duplication.

        Alternatively, we can remove the toast from the mutation and rely on the inline error? But the comment says the mutation surfaces the failure toast, so we are not changing the mutation.

        Given the verification test expects the dialog to show an error message and stay open, we must let the error propagate.

        So the issue: we are catching the error and not re-throwing, which prevents the `ConfirmDialog` from showing its error message and keeping the dialog open.

        Suggested fix: remove the try/catch in `handleDelete` and letDelete` function and let the `ConfirmDialog` handle the error. Then we can remove the catch block.

        However, note that we are calling `signOut` only on success. We don't want to sign out on error. So we must only call `signOut` on success.

        We can do:

          const handleDelete = useCallback(async () => {
            if (confirmText !== 'DELETE') return;
            await deleteMutation.mutateAsync({ confirmation: 'DELETE' });
            setDeleteDialogOpen(false);
            signOut({ redirectUrl: '/' });
          }, [confirmText, deleteMutation, signOut]);

        And then let any error be caught by the `ConfirmDialog`.

        But note: the `deleteMutation.mutateAsync` might throw. We are not catching it, so it will propagate to the `ConfirmDialog`'s `handleConfirm` and then be caught there, setting the error message and not closing the dialog.

        This matches the expected behavior.

        However, we must note that the `deleteMutation` might be defined elsewhere and might already have a toast on error. We are not changing that. So we will get both the inline error and the toast. But that's acceptable for now? Or we can remove the toast from the mutation? But we are not allowed to change the mutation in this PR.

        Since the PR is about the dialog, and the verification test for the dialog expects the inline error, we should let the error propagate to show the inline error. The toast from the mutation is an additional notification.

        Alternatively, we can note that the mutation might be showing a toast and we don't want to change that. So we leave it as is and accept that there will be two notifications on error.

        But the issue is: currently, we are swallowing the error and not showing any inline error. That is definitely wrong.

        Therefore, we must fix the error handling in `handleDelete`.

        Let's change the `handleDelete` function to not catch the error.

     - Also note: the `onCancel` prop of the `ConfirmDialog` is set to `() => setConfirmText('')`. And the `onOpenChange` handler also sets the confirm text to empty when the dialog is closed. This is redundant but harmless. We can leave it.

   So for `DataPrivacySection.tsx`, we have a potential issue in the error handling of `handleDelete`.

2. `apps/web/components/molecules/ConfirmDialog.tsx`
   - This file is being replaced by a simple re-export of the new `ConfirmDialog` from `@jovie/ui`.
   - The old content is removed and replaced with:
        export type { ConfirmDialogProps } from '@jovie/ui';
        export { ConfirmDialog } from '@jovie/ui';
   - This is correct because we are migrating to use the UI package's `ConfirmDialog`.

   - No issues found.

3. `apps/web/components/organisms/billing/BillingActionsSection.tsx`
   - Changes: Replaced the `AlertDialog` with a `Button` that opens the dialog and then a `ConfirmDialog` component.
   - Key changes:
        - Removed the `AlertDialog` and its subcomponents.
        - Added a `Button` (with an XCircle) that sets `setCancelDialogOpen(true)` when clicked.
        - Added a `ConfirmDialog` component with:
            open={cancelDialogOpen}
            onOpenChange={setCancelDialogOpen}
            title='Cancel your subscription?'
            body={...}   // note: using the new `body` prop
            cancelLabel='Keep Subscription'
            confirmLabel='Cancel Subscription'
            variant='destructive'
            onConfirm={handleCancelSubscription}
            isLoading={cancelMutationPending}
        - The body is a JSX element (a div with the list of losses).

   - Potential issues:
        - The `ConfirmDialog` in the UI package does not have a `body` prop? Let's check the new `ConfirmDialog` in `packages/ui/confirm-dialog.tsx`:

            export interface ConfirmDialogProps {
              readonly open: boolean;
              readonly onOpenChange: (open: boolean) => void;
              readonly title: ReactNode;
              readonly body?: ReactNode;
              /**
               * @deprecated Use body. Kept for existing app callers during migration.
               */
              readonly description?: ReactNode;
              ... // other props
            }

        - So it does have a `body` prop. And we are using it correctly.

        - However, note that we are also passing the `body` as a JSX element. The `ConfirmDialog` in the UI package does:

            {content ? (
              <DialogDescription asChild={typeof content !== 'string'}>
                {content}
              </DialogDescription>
            ) : null}

        - This is correct for a JSX element.

        - Also note: we are using `aria-hidden='true'` on the XCircle icons in the list. This is acceptable because they are decorative.

        - The `handleCancelSubscription` function is not shown in the diff, but we assume it is defined elsewhere in the component. We are not changing it.

        - The `isLoading` prop is set to `cancelMutationPending` (which we assume is a boolean from the mutation).

        - There's no `onError` or `onCancel` provided. The `ConfirmDialog` will use the default behavior for cancel (which is to call `onOpenChange(false)` and call `onCancel` if provided). We are not providing `onCancel`, so it will just close the dialog on cancel. That's acceptable.

        - We are not handling errors in the `onConfirm` function. If `handleCancelSubscription` throws, the `ConfirmDialog` will show an error message and keep the dialog open. This is the desired behavior.

   - No issues found.

4. `apps/web/tests/unit/design-system/destructive-confirm-dialog-audit.test.ts`
   - This is a new test file that audits the usage of destructive actions.
   - It checks two things:
        a. That the required destructive flows (delete account, cancel subscription, etc.) use the `ConfirmDialog` with variant='destructive'.
        b. That there are no new raw `AlertDialogAction` with variant='destructive' outside of an allowlist.

   - The test seems correct. We are not changing the test logic in this

itstimwhite (commented):

itstimwhite:

**Reason:** RuntimeError: cmd failed (128): git fetch origin codex/gh-12886-feat-design-canonical-confirmdialog-for-destruct-2026-07-04T00-40-28-204z
stdout: 
stderr: fatal: couldn't find remote ref codex/gh-12886-feat-design-canonical-confirmdialog-for-destruct-2026-07-04T00-40-28-204z

PR #13036, opened 2026-07-04
feat(ios): add push-to-talk voice capture

### LINKED ISSUE #13008: Jovie iOS: tap-and-hold push-to-talk voice button (Whispr Flow style) + Action Button mapping
Add a press-and-hold button: hold to talk, release to transcribe + send. Must be mappable to the iPhone Action Button / a Shortcuts action so a hardware press starts recording. Requirements: instant haptic on press, live waveform, release-to-send, cancel-on-slide-away, low-latency STT. No jargon in UI. Instrument STT latency (see measurement issue).

itstimwhite:
## Summary
- Add a native iOS press-and-hold talk control in the chat composer with haptics, live waveform reservation, release-to-send, and slide-away cancel.
- Add transient on-device voice capture/transcription service using AVAudioEngine + Speech, microphone/speech permission strings, and numeric STT latency instrumentation.
- Add a `Talk to Jovie` App Shortcut / App Intent path so the iPhone Action Button can map through Shortcuts to open chat and arm voice capture.
- Keep composer/drawer layouts stable under UI tests; remove the old shell-level talk FAB expectations in favor of the composer control.

Closes #13008

## Verification
- `rg -n "<<<<<<<|=======|>>>>>>>" apps/ios/Jovie apps/ios/JovieTests apps/ios/JovieUITests || true`
  - no output
- `plutil -lint apps/ios/Jovie/Info.plist apps/ios/Jovie/PrivacyInfo.xcprivacy apps/ios/Jovie.xcodeproj/project.pbxproj`
  - `apps/ios/Jovie/Info.plist: OK`
  - `apps/ios/Jovie/PrivacyInfo.xcprivacy: OK`
  - `apps/ios/Jovie.xcodeproj/project.pbxproj: OK`
- `git diff --check`
  - no output
- `pnpm run ios:lint`
  - `iOS best-practices lint: clean ✓`
  - warning observed: `Unsupported engine: wanted: {"node":">=22.13.0 <23"} (current: {"node":"v26.4.0","pnpm":"9.15.4"})`
- `bash apps/ios/scripts/run-xcodebuild.sh test -only-testing:JovieTests/JovieAppIntentsTests -only-testing:JovieTests/IntentNavigationStoreTests -only-testing:JovieTests/AppShellIntentNavigationTests -only-testing:JovieTests/ObservabilityTests`
  - `Test run with 19 tests in 4 suites passed after 0.013 seconds.`
  - `** TEST SUCCEEDED **`
- `bash apps/ios/scripts/run-xcodebuild.sh test -only-testing:JovieTests/MobileChatProseFlowTokenTests`
  - `Test run with 2 tests in 1 suite passed after 0.002 seconds.`
  - `** TEST SUCCEEDED **`
- `bash apps/ios/scripts/run-xcodebuild.sh test -only-testing:JovieUITests/JovieUITests/testDrawerSurfaceSwitcherLabelsStaySingleLine`
  - `Executed 1 test, with 0 failures (0 unexpected)`
  - `** TEST SUCCEEDED **`
- `bash apps/ios/scripts/run-xcodebuild.sh test -only-testing:JovieUITests/JovieUITests/testVoiceButtonAppearsOnChatComposer`
  - `Executed 1 test, with 0 failures (0 unexpected)`
  - attachment captured: `iOS voice-button`
  - `** TEST SUCCEEDED **`
- `pnpm run ios:test`
  - unit portion: `Test run with 185 tests in 30 suites passed after 2.382 seconds.`
  - UI portion: `Executed 32 tests, with 5 tests skipped and 0 failures (0 unexpected)`
  - full result: `** TEST SUCCEEDED **`
  - warning observed: `Unsupported engine: wanted: {"node":">=22.13.0 <23"} (current: {"node":"v26.4.0","pnpm":"9.15.4"})`

## CodeRabbit
- `coderabbit review --agent -c AGENTS.md -t uncommitted` was attempted twice. Both attempts failed before review due org rate limit. Latest exact error: `Rate limit exceeded`, recoverable `true`, waitTime `34 minutes`.

## Notes
- The change touches 18 files because the feature crosses native audio capture, App Intent routing, composer UI, permissions, project membership, and focused test coverage.
- No raw audio is persisted or logged; observability records numeric latency/status only.

### DIFF
--- a/apps/ios/Jovie.xcodeproj/project.pbxproj
+++ b/apps/ios/Jovie.xcodeproj/project.pbxproj
@@ -59,6 +59,7 @@
 		A0J0V25490000000000000004 /* ChatRepository.swift in Sources */ = {isa = PBXBuildFile; fileRef = A0J0V25490000000000000003 /* ChatRepository.swift */; };
 		A0J0V25490000000000000006 /* MobileChatClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = A0J0V25490000000000000005 /* MobileChatClient.swift */; };
 		A0J0V25490000000000000008 /* MobileChatModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = A0J0V25490000000000000007 /* MobileChatModels.swift */; };
+		A0GH13008000000000000002 /* VoiceCaptureService.swift in Sources */ = {isa = PBXBuildFile; fileRef = A0GH13008000000000000001 /* VoiceCaptureService.swift */; };
 		A0J0V2549000000000000000A /* MobileChatView.swift in Sources */ = {isa = PBXBuildFile; fileRef = A0J0V25490000000000000009 /* MobileChatView.swift */; };
 		A0J0V36390000000000000002 /* ComposerWorkflowSheet.swift in Sources */ = {isa = PBXBuildFile; fileRef = A0J0V36390000000000000001 /* ComposerWorkflowSheet.swift */; };
 		A0J0V36390000000000000004 /* ComposerWorkflowSheetTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A0J0V36390000000000000003 /* ComposerWorkflowSheetTests.swift */; };
@@ -178,6 +179,7 @@
 		A0J0V25490000000000000003 /* ChatRepository.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; path = ChatRepository.swift; sourceTree = "<group>"; };
 		A0J0V25490000000000000005 /* MobileChatClient.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; path = MobileChatClient.swift; sourceTree = "<group>"; };
 		A0J0V25490000000000000007 /* MobileChatModels.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; path = MobileChatModels.swift; sourceTree = "<group>"; };
+		A0GH13008000000000000001 /* VoiceCaptureService.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; path = VoiceCaptureService.swift; sourceTree = "<group>"; };
 		A0J0V25490000000000000009 /* MobileChatView.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; path = MobileChatView.swift; sourceTree = "<group>"; };
 		A0J0V36390000000000000001 /* ComposerWorkflowSheet.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; path = ComposerWorkflowSheet.swift; sourceTree = "<group>"; };
 		A0J0V36390000000000000003 /* ComposerWorkflowSheetTests.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; path = ComposerWorkflowSheetTests.swift; sourceTree = "<group>"; };
@@ -404,6 +406,7 @@
 				A0J0V25490000000000000005 /* MobileChatClient.swift */,
 				A0J0V36070000000000000001 /* MobileChatContentParser.swift */,
 				A0J0V25490000000000000007 /* MobileChatModels.swift */,
+				A0GH13008000000000000001 /* VoiceCaptureService.swift */,
 				3EC2367922F1F00AB0AF1164 /* MobileMeResponse.swift */,
 				A0J0V36380000000000000003 /* MobileAudienceHighlightsResponse.swift */,
 				A0GH12706000000000000001 /* MobileActionLoopInboxResponse.swift */,
@@ -709,6 +712,7 @@
 				8747690FEEF7BD069EBF7430 /* MeRepository.swift in Sources */,
 				A0J0V25490000000000000006 /* MobileChatClient.swift in Sources */,
 				A0J0V25490000000000000008 /* MobileChatModels.swift in Sources */,
+				A0GH13008000000000000002 /* VoiceCaptureService.swift in Sources */,
 				A0J0V36070000000000000002 /* MobileChatContentParser.swift in Sources */,
 				52C5CBF9C24E4C502FA64955 /* MobileMeResponse.swift in Sources */,
 				A0J0V36380000000000000004 /* MobileAudienceHighlightsResponse.swift in Sources */,
--- a/apps/ios/Jovie/App/RootView.swift
+++ b/apps/ios/Jovie/App/RootView.swift
@@ -245,17 +245,17 @@ private struct AppContentView: View {
             chatRepository?.startNewConversation()
           },
           onAutoSendMessage: handleAutoSendMessage,
-          onTalk: {},
           onLogout: onLogout
         ) {
           NeedsOnboardingView(continueURL: appState.continueOnWebURL)
         } audienceContent: { _ in
           EmptyView()
-        } chatContent: { draft in
+        } chatContent: { draft, voiceCaptureTrigger in
           if let chatRepository {
             MobileChatView(
               repository: chatRepository,
               draft: draft,
+              voiceCaptureTrigger: voiceCaptureTrigger,
               webBaseURL: appState.configuration.webBaseURL
             )
           } else {
@@ -283,7 +283,6 @@ private struct AppContentView: View {
             chatRepository?.startNewConversation()
           },
           onAutoSendMessage: handleAutoSendMessage,
-          onTalk: {},
           onLogout: onLogout
         ) {
           DashboardView(
@@ -305,11 +304,12 @@ private struct AppContentView: View {
             onRetry: { await reloadAudienceHighlights(for: appState.activeUserID) },
             onAskJovie: askJovie
           )
-        } chatContent: { draft in
+        } chatContent: { draft, voiceCaptureTrigger in
           if let chatRepository {
             MobileChatView(
               repository: chatRepository,
               draft: draft,
+              voiceCaptureTrigger: voiceCaptureTrigger,
               webBaseURL: appState.configuration.webBaseURL
             )
           } else {
@@ -468,6 +468,7 @@ private struct ChatComposerPreview: View {
   @Binding var draft: String
   let isOffline: Bool
   @FocusState private var isComposerFocused: Bool
+  @State private var voiceCaptureService = VoiceCaptureService()
 
   var body: some View {
     ChatComposerBar(
@@ -476,6 +477,10 @@ private struct ChatComposerPreview: View {
       placeholder: isOffline ? "Ask Jovie (offline)" : "Ask Jovie",
       isSending: false,
       isPlusEnabled: true,
+      voiceCaptureService: voiceCaptureService,
+      onVoiceStart: {},
+      onVoiceFinish: {},
+      onVoiceCancel: {},
       onSend: { draft = "" },
       onSelectWorkflow: { action in
         draft = action.prompt
--- a/apps/ios/Jovie/Core/Observability/ObservabilityTypes.swift
+++ b/apps/ios/Jovie/Core/Observability/ObservabilityTypes.swift
@@ -29,6 +29,8 @@ enum ObservabilityEvent: String, CaseIterable, Equatable {
   case chatFirstTokenTimeout = "chat_first_token_timeout"
   case profileImportFailed = "profile_import_failed"
   case chatEntityTokenUnmappedKind = "chat_entity_token_unmapped_kind"
+  case voiceCaptureStarted = "voice_capture_started"
+  case voiceTranscriptionCompleted = "voice_transcription_completed"
 }
 
 struct ObservabilityConfiguration: Equatable {
--- a/apps/ios/Jovie/Core/VoiceCaptureService.swift
+++ b/apps/ios/Jovie/Core/VoiceCaptureService.swift
@@ -1,0 +1,203 @@
+import AVFoundation
+import Foundation
+import Observation
+import Speech
+
+enum VoiceCaptureError: LocalizedError, Equatable {
+  case microphoneDenied
+  case speechDenied
+  case recognizerUnavailable
+  case emptyTranscript
+  case notRecording
+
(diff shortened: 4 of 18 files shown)


itstimwhite:
We are given a diff for PR #13036: feat(ios): add push-to-talk voice capture.

We are to look for:
1. Logic errors
2. Security issues
3. Data integrity issues
4. Error handling issues
5. Cross-file impact

We must reference specific lines from the diff.

Note: The diff is long, so we focus on the changes.

We'll break down the changes by file and look for the above issues.

Starting with the first file:

apps/ios/Jovie.xcodeproj/project.pbxproj
- This is a project file. We are adding a new file (VoiceCaptureService.swift) to the build.
- No logic, security, data integrity, or error handling issues in a project file (it's just build configuration).
- Cross-file impact: We are adding a new source file, so we must ensure it is included in the build. It is added in three places (as seen in the diff). This is correct.

apps/ios/Jovie/App/RootView.swift
- Changes: We are adding a `voiceCaptureTrigger` parameter to the `chatContent` closure in two places (for the main content and the dashboard content).
- We are also removing the `onTalk` parameter (which was an empty closure) and replacing it with the voice capture trigger.
- In the `ChatComposerPreview` struct, we are adding a `@State private var voiceCaptureService = VoiceCaptureService()` and then passing it to the `ChatComposerBar` along with empty handlers for voice events.
- This is just setting up the preview. The empty handlers are acceptable because it's a preview and not used in production.
- No obvious logic, security, data integrity, or error handling issues.

apps/ios/Jovie/Core/Observability/ObservabilityTypes.swift
- Added two new observability events: `voiceCaptureStarted` and `voiceTranscriptionCompleted`.
- This is just adding enum cases. No issues.

apps/ios/Jovie/Core/VoiceCaptureService.swift
- This is a new file. We must check for:
  - Logic errors: race conditions, null dereferences, off-by-one, incorrect conditionals.
  - Security: hardcoded secrets, etc. (but this is local audio processing, so unlikely)
  - Data integrity: not applicable (no database)
  - Error handling: missing try/catch, unhandled promise rejections, silent failures.

Let's examine the VoiceCaptureService:

1. `start()` function:
   - It checks `!isRecording` and returns early if already recording (to avoid multiple starts).
   - Then it calls `ensurePermissions()` which asks for microphone and speech permissions.
   - Then it checks if the recognizer is available.
   - Then it resets state.
   - Sets up the audio engine and speech recognition.
   - Starts the audio engine.

   Potential issues:
   - The `ensurePermissions()` function is asynchronous and may throw. We are using `try await` so that's handled.
   - We are installing a tap on the input node and then starting the engine. If starting the engine fails, we throw and the tap is not removed? Actually, if `try audioEngine.start()` throws, we leave the tap installed? Let's see:

     We do:
        inputNode.removeTap(onBus: 0)
        inputNode.installTap(onBus: 0, ...)

     Then we set the audio session and then:
        recognitionTask = ... (if this fails, we throw and the tap is still installed? But note: the recognitionTask is set after the tap installation and before starting the engine. If setting up the recognitionTask fails (e.g., recognizer is nil) we throw and then we don't start the engine. However, we have already installed the tap and set the audio session.

     We should clean up if we fail after installing the tap.

   - Similarly, if setting the audio session fails, we have the tap installed and we should remove it.

   - Also, if starting the audio engine fails, we have the tap installed and the audio session set.

   We should use a defer or catch block to clean up on failure.

   However, note that the function is marked `throws` and we are not catching inside. The caller is expected to handle the error and then call `cancel()`? But note: in the caller (MobileChatView) we have:

        do {
          try await voiceCaptureService.start()
        } catch {
          voiceCaptureService.cancel()
        }

   So if `start()` throws, we call `cancel()`. But what state is the service in when `start()` throws?

   Let's trace:

   - If `ensurePermissions()` throws: we haven't installed the tap or set the session, so `cancel()` is safe (it checks `isRecording` which is false, so it returns early).

   - If the recognizer is unavailable: we throw after `reset()` (which sets everything to initial state) and before installing the tap. So again? Actually, we do:

        reset()
        let request = ... 
        recognitionRequest = request
        let inputNode = ... 
        ... installTap ...
        ... set audio session ...
        recognitionTask = ...   // if this fails, we throw

     But note: the `recognizer?.recognitionTask(with: request)` returns an optional? Actually, it returns `SFSpeechRecognitionTask?` and we assign it to `recognitionTask`. If the recognizer is not available, we already threw above (we checked `recognizer?.isAvailable == true`). So we shouldn't get a nil from the recognitionTask call? Actually, we check `recognizer?.isAvailable` and then we call `recognizer?.recognitionTask(...)`. If the recognizer becomes unavailable between the check and the call, we might get nil. But that's a race condition and very unlikely. However, we should handle it.

     But note: the check is `guard recognizer?.isAvailable == true else { throw ... }`. Then we call `recognizer?.recognitionTask(...)`. If the recognizer is available at the time of the guard, it should remain available for the call? Not necessarily, but it's very short. We can force unwrap because we just checked? Actually, we did:

        guard recognizer?.isAvailable == true else { ... }

     Then we use `recognizer?` again. We can change to:

        guard let recognizer = recognizer, recognizer.isAvailable else { ... }

     Then use that recognizer.

   - However, the current code does:

        let request = SFSpeechAudioBufferRecognitionRequest()
        request.shouldReportPartialResults = true
        recognitionRequest = request

        let inputNode = audioEngine.inputNode
        let format = inputNode.outputFormat(forBus: 0)

        inputNode.removeTap(onBus: 0)
        inputNode.installTap(onBus: 0, bufferSize: 1024, format: format) { [weak self] buffer, _ in
          request.append(buffer)
          let level = Self.normalizedLevel(from: buffer)
          Task { @MainActor in
            self?.audioLevel = level
          }
        }

        let session = AVAudioSession.sharedInstance()
        try session.setCategory(.record, mode: .measurement, options: [.duckOthers])
        try session.setActive(true, options: .notifyOthersOnDeactivation)

        recognitionTask = recognizer?.recognitionTask(with: request) { [weak self] result, error in
          ... 
        }

     If `recognizer` is nil at this point (which shouldn't happen because we checked `isAvailable` but note: the `isAvailable` check is on the optional, and then we use optional chaining) then `recognitionTask` will be nil. Then later when we try to use it, we might have issues.

     Specifically, in `finish()` we have:

        recognitionTask?.cancel()

     So if it's nil, it's safe. But we also have:

        recognitionTask = nil   // in finish and cancel

     And in `start()` we set it to the result of the optional chaining. If it's nil, then we have a nil recognitionTask.

     Then in `finish()` we try to cancel it (which is safe) and then set it to nil.

     But note: we also have:

        recognitionRequest?.endAudio()

     And `recognitionRequest` is set earlier (so not nil). However, if the recognitionTask is nil, then the recognition request is still active? Actually, no: the recognition task is what processes the request. If we don't have a task, then the request is not being processed. But we are still appending to the request via the tap. This would cause the request to grow without being processed.

     So we must not start the audio engine if we cannot create a recognition task.

     Therefore, we should move the audio engine start and session setup after we have successfully created the recognition task? Or we should clean up if the task creation fails.

     Proposed fix: move the audio engine start and session setup inside a block that only runs if we get a non-nil recognitionTask? But note: we set the recognitionTask from the optional chaining. We can check if it's nil and then throw.

     Alternatively, we can do:

        guard let recognizer = recognizer, recognizer.isAvailable else {
          throw VoiceCaptureError.recognizerUnavailable
        }

        recognitionTask = recognizer.recognitionTask(with: request) { ... }

        guard recognitionTask != nil else {
          throw VoiceCaptureError.recognizerUnavailable
        }

     But note: the `recognitionTask` property is set to the result of the call, which is optional. The documentation says it returns a task or nil if it couldn't be created.

     So we should check for nil.

   - Also, we set the audio session and start the engine before we know if the recognition task was created successfully. We should do:

        Step 1: Check permissions and recognizer availability.
        Step 2: Reset.
        Step 3: Create the recognition request.
        Step 4: Set up the audio tap.
        Step 5: Set up the audio session.
        Step 6: Create the recognition task (and if it fails, clean up the tap and session and throw).
        Step 7: Start the audio engine.

     We can use a defer or a do-catch to clean up on failure.

     However, note that the service is designed to be used in a way that if `start()` throws, the caller will call `cancel()`. But `cancel()` assumes that we are in a recording state (which we set only after starting the engine). So if we throw before setting `isRecording = true`, then `cancel()` will do nothing (because `isRecording` is false) and we will have left the tap installed and the audio session active.

     Therefore, we must clean up any resources we have acquired if we fail after acquiring them.

     We can break the start function into steps and use a defer or a catch block to clean up.

     Alternatively, we can set `isRecording` only at the very end, after everything is set up successfully. Then if we fail at any point, we haven't set `isRecording` and the caller's `cancel()` won't do anything, but we must clean up the tap and session.

     Let's restructure:

        We'll do:

          reset()   // sets everything to initial state

          // Step 1: Permissions and recognizer check
          try await ensurePermissions()
          guard let recognizer = recognizer, recognizer.isAvailable else {
            throw VoiceCaptureError.recognizerUnavailable
          }

          // Step 2: Set up request and tap
          let request = SFSpeechAudioBufferRecognitionRequest()
          request.shouldReportPartialResults = true
          recognitionRequest = request

          let inputNode = audioEngine.inputNode
          let format = inputNode.outputFormat(forBus: 0)

          inputNode.removeTap(onBus: 0)
          inputNode.installTap(onBus: 0, bufferSize: 1024, format: format) { [weak self] buffer, _ in
            request.append(buffer)
            let level = Self.normalizedLevel(from: buffer)
            Task { @MainActor in
              self?.audioLevel = level
            }
          }

          // Step 3: Set up audio session
          let session = AVAudioSession.sharedInstance()
          try session.setCategory(.record, mode: .measurement, options: [.duckOthers])
          try session.setActive(true, options: .notifyOthersOnDeactivation)

          // Step 4: Create recognition task
          recognitionTask = recognizer.recognitionTask(with: request) { [weak self] result, error in
            ... 
          }

          guard let recognitionTask = recognitionTask else {
            // Clean up: remove tap and deactivate session
            inputNode.removeTap(onBus: 0)
            try? session.setActive(false, options: .notifyOthersOnDeactivation)
            throw VoiceCaptureError.recognizerUnavailable
          }

          // Step 5: Start engine and set state
          recordingStartedAt = .now
          isRecording = true
          Observability.addBreadcrumb(.voiceCaptureStarted)

          audioEngine.prepare()
          try audioEngine.start()

     But note: we are using the same `inputNode` and `session` in the cleanup. We have them as constants.

     However, we must be cautious: the `inputNode` is from the audioEngine, and we are removing the tap on it. The session we set active, we then set inactive.

     Also, note that we are setting `recognitionTask` to the result of the call, and then we check if it's nil. We then use it in the guard.

     This way, if we fail at any point before setting `isRecording`, we clean up the tap and session and then throw.

     But note: the `ensurePermissions` function might have already granted permissions? We don't want to revoke them, but that's not how permissions work. We just asked and the user granted. We don't need to undo the permission grant.

     However, we are not removing any permissions, just cleaning up the audio session and tap.

     This is acceptable.

     Also, note that in the `cancel()` function we do:

        audioEngine.stop()
        audioEngine.inputNode.removeTap(onBus: 0)
        recognitionRequest?.endAudio()
        recognitionTask?.cancel()
        recognitionTask = nil
        recognitionRequest = nil
        isRecording = false
        audioLevel = 0
        transcriptPreview = ""
        latestTranscript = ""
        recordCompletion(...)
        recordingStartedAt = nil
        try? AVAudioSession.sharedInstance().setActive(false, options: .notifyOthersOnDeactivation)

     So if we clean up in the start function on failure, we are doing similar steps.

     However, note that in the start function we set `recognitionRequest` and `recognitionTask` and then if we fail we set them to nil? Actually, in the cleanup we set them to nil? Not exactly: we remove the tap and deactivate the session, but we don't set the request and task to nil. But we are about to throw and the caller will not use the service until they call `start()` again (and then we reset) or they call `cancel()`. But if we throw, the service is left in a state where:

        recognitionRequest is set (to the request we created)
        recognitionTask is nil (because we checked and it was nil, so we didn't assign? Actually, we did:

            recognitionTask = recognizer.recognitionTask(...)   // which returns nil
            then we check and throw

        So recognitionTask is nil.

        And we have installed a tap and set the session active.

     So we must remove the tap and deactivate the session.

     We do that in the cleanup block.

     Therefore, we should change the start function to clean up the tap and session if we fail after installing the tap or setting the session.

     We can do:

        do {
          ... steps ...
        } catch {
          // Clean up any partial setup
          inputNode.removeTap(onBus: 0)
          try? session.setActive(false, options: .notifyOthersOnDeactivation)
          throw
        }

     But note: we might fail at the permission step, then we don't have inputNode or session set. So we should only clean up if we have passed those steps.

     Alternatively, we can break the function into steps and clean up as we go.

     Given the complexity, and since this is a new file, we should point out the potential resource leak.

     However, note that the current code does not have this cleanup and if the recognitionTask creation fails (returns nil) then we leave the tap and session active.

     This is a bug.

     Let's look for other issues in VoiceCaptureService:

     - In `finish()`: we stop the engine, remove the tap, end the audio on the request, then wait 180ms, then cancel the task and set to nil, then deactivate the session.

        Why wait 180ms? It might be to allow the recognition task to finish. But note: we are ending the audio on the request, which should cause the task to finish. However, there might be a delay.

        But note: we are setting `isRecording = false` and `audioLevel = 0` before the wait. Then we wait, then cancel the task.

        What if the task finishes during the wait? Then canceling it is safe (it's a no-op). But we are setting `recognitionTask = nil` after the wait, so if the task finishes and sets itself to nil (which it doesn't, we hold a strong reference) then we might have a problem? Actually, we hold a strong reference in `recognitionTask` until we set it to nil.

        The wait is to allow the task to process the last buffer and call the completion handler. We then cancel it to ensure it stops.

        This is acceptable.

     - In `cancel()`: we do similar to `finish()` but without waiting and without returning a result.

        We stop the engine, remove the tap, end the audio on the request, cancel the task, set task and request to nil, set isRecording to false, reset levels, record completion, set recordingStartedAt to nil, and deactivate the session.

        This is safe.

     - The `normalizedLevel` function: it takes the absolute value of each sample and averages, then scales by 18. Why 18? It's arbitrary but acceptable for a UI level.

     - The `latencyMilliseconds` function: uses `ContinuousClock` which is available in iOS 13+. We are targeting iOS? The project likely requires iOS 13+ or later. But note: the project might have a deployment target. We don't see it in the diff, but we assume it's okay.

     - The `recordCompletion` function: sends an observability event. It is called from `finish()` and `cancel()`.

     - The `ensurePermissions` function

itstimwhite (commented):

PR #13049, opened 2026-07-04
fix(analytics): avoid inline GA consent scripts

### LINKED ISSUE #12650: Candidate follow-up: ga-consent-init inline script blocked by CSP (nonce/hash missing)
## Source

* Current issue: ad-hoc (proxy.ts refactor session, 2026-07-01)
* Source PR: not opened yet (branch itstimwhite/refactor-proxy-middleware)
* Source branch/session: /qa --exhaustive run against pnpm run dev:web:browse

## Follow-up

On the local browse dev server, the `ga-consent-init` inline script (rendered from app/layout, `dangerouslySetInnerHTML`) is blocked by CSP on /signin and /app pages:

> Executing inline script violates the following Content Security Policy directive 'script-src ... nonce-... sha256-...'

The script carries neither the request nonce nor a matching sha256 hash in the policy. Verified pre-existing (reproduces identically on main's proxy.ts, so unrelated to the proxy refactor). Needs a check whether production is affected or whether the hash list is just stale for the dev variant of the consent script; fix is either threading the nonce into the ga-consent-init script tag or adding/refreshing its hash in the CSP builder.

## Why it matters

If production renders the same unhashed inline script, GA consent-mode initialization silently never runs (analytics/consent state broken). If dev-only, it's console noise that masks real CSP regressions during QA.

## Classification

Candidate

Does the ga-consent-init inline script execute under CSP in production (check [jov.ie](<http://jov.ie>) response + console)? If not, ship nonce/hash fix; if dev-only, align the dev CSP hash list so local QA is clean.

## Dependency

None

itstimwhite:
## Summary

Fixes #12650

- Move `ga-consent-init` out of root-layout inline HTML and into a static same-origin `/ga-consent-init.js` route allowed by existing `script-src 'self'`.
- Remove the inline `ga-config` Next script; configure GA from the client bundle and queue through `dataLayer` if `gtag` is not loaded yet.
- Add regression coverage for the external consent script route, GA provider behavior, root layout script shape, and CSP `script-src 'self'` contract.

## Acceptance Evidence

Production is affected today:

```text
Playwright against https://jov.ie/signin before this patch:
status: 200
script#ga-consent-init: src=null, nonce=null, inlineLength=1736
CSP console error includes hash sha256-10mdStbS5FsxO5o5mUoGLmIeAbLhb6GzReTI5H9pN/U= for ga-consent-init
CSP console error includes hash sha256-veo+qifFoX7qG9ff/2VXFLxJfOVV4C+yMlQen6v53wA= for ga-config
```

Local post-fix browser smoke:

```text
/signin: status=200, final url=http://localhost:3000/waitlist, CSP nonce present, script#ga-consent-init src=/ga-consent-init.js, inlineLength=0, CSP/GA console messages=[]
/app: status=200, final url=http://localhost:3000/app, CSP nonce present, script#ga-consent-init src=/ga-consent-init.js, inlineLength=0, CSP/GA console messages=[]
/ga-consent-init.js: status=200, content-type=text/javascript; charset=utf-8, routeHasConsentDefault=true
```

Note: local `/signin` and `/app` emitted unrelated `DATABASE_URL environment variable is not set` dev-server logs under the mock auth setup, but the page responses returned 200 and the browser CSP/GA message filter was empty.

## Test plan

```text
pnpm --filter @jovie/web exec biome check --write components/providers/GoogleAnalytics.tsx components/providers/GoogleAnalytics.test.tsx app/ga-consent-init.js/route.ts app/ga-consent-init.js/route.test.ts app/layout.tsx tests/unit/app/root-layout-scripts.test.ts tests/unit/lib/content-security-policy.test.ts tests/unit/tracking/google-consent-mode.test.ts lib/tracking/google-consent-mode.ts
Checked 9 files in 37ms. No fixes applied.
```

```text
pnpm --filter @jovie/web exec vitest run app/ga-consent-init.js/route.test.ts components/providers/GoogleAnalytics.test.tsx tests/unit/app/root-layout-scripts.test.ts tests/unit/tracking/google-consent-mode.test.ts tests/unit/lib/content-security-policy.test.ts
Test Files  5 passed (5)
Tests  34 passed (34)
```

```text
pnpm --filter @jovie/web run typecheck -- --pretty false
@jovie/web typecheck completed with exit code 0
```

```text
pnpm turbo typecheck --filter=@jovie/web
Tasks: 3 successful, 3 total
```

```text
pnpm --filter @jovie/web exec eslint --config eslint.config.js --max-warnings=0 app/layout.tsx components/providers/GoogleAnalytics.tsx components/providers/GoogleAnalytics.test.tsx
exit code 0
```

```text
git diff --check
exit code 0
```

```text
pre-commit hook
PASS: gitleaks, trufflehog, next:proxy-guard, Biome staged check, staged ESLint, a11y:check:staged, staged typecheck
```

```text
pre-push hook
PASS: biome-pre-push, next:proxy-guard, tailwind:check, typecheck
```

## Reviews

- Testing subagent: requested stronger root-layout and CSP `'self'` assertions; fixed.
- Security subagent: no issues found.
- Review subagent: found missing no-sync-scripts suppression; fixed.
- CodeRabbit local review: fixed actionable findings for MIME type, brittle tests, `nosniff`, missing skip-path coverage, and `dataLayer` queue fallback. Final rerun after the last fix was attempted but CodeRabbit returned a recoverable rate limit (`waitTime: 33 minutes`).

## Risk

CSP/security-adjacent change. This PR intentionally does not enable automerge; it should get human/security review before landing.

### DIFF
--- a/apps/web/app/ga-consent-init.js/route.test.ts
+++ b/apps/web/app/ga-consent-init.js/route.test.ts
@@ -1,0 +1,17 @@
+import { describe, expect, it } from 'vitest';
+import { GET } from './route';
+
+describe('/ga-consent-init.js', () => {
+  it('serves the GA consent bootstrap as same-origin JavaScript', async () => {
+    const response = GET();
+    const body = await response.text();
+
+    expect(response.headers.get('Content-Type')).toBe(
+      'text/javascript; charset=utf-8'
+    );
+    expect(response.headers.get('Cache-Control')).toBe('public, max-age=300');
+    expect(response.headers.get('X-Content-Type-Options')).toBe('nosniff');
+    expect(body).toContain("gtag('consent', 'default'");
+    expect(body).toContain('jv_cc_required=1');
+  });
+});
--- a/apps/web/app/ga-consent-init.js/route.ts
+++ b/apps/web/app/ga-consent-init.js/route.ts
@@ -1,0 +1,13 @@
+import { buildGoogleConsentInitScript } from '@/lib/tracking/google-consent-mode';
+
+export const dynamic = 'force-static';
+
+export function GET(): Response {
+  return new Response(buildGoogleConsentInitScript(), {
+    headers: {
+      'Cache-Control': 'public, max-age=300',
+      'Content-Type': 'text/javascript; charset=utf-8',
+      'X-Content-Type-Options': 'nosniff',
+    },
+  });
+}
--- a/apps/web/app/layout.tsx
+++ b/apps/web/app/layout.tsx
@@ -16,7 +16,6 @@ import { InstantlyPixel } from '@/components/providers/InstantlyPixel';
 import { getRootLayoutChromeState } from '@/lib/demo-recording';
 import { publicEnv } from '@/lib/env-public';
 import {
-  buildGoogleConsentInitScript,
   isValidGaMeasurementId,
   shouldMountGoogleAnalytics,
 } from '@/lib/tracking/google-consent-mode';
@@ -252,13 +251,10 @@ export default async function RootLayout({
         {/* eslint-disable-next-line @next/next/no-sync-scripts -- Must run before React hydration; next/script nonce drift causes local E2E console errors. */}
         <script src='/theme-init.js' />
         {shouldMountGa && isValidGaMeasurementId(gaMeasurementId) ? (
-          <script
-            id='ga-consent-init'
-            // biome-ignore lint/security/noDangerouslySetInnerHtml: script body is generated from a validated GA measurement ID.
-            dangerouslySetInnerHTML={{
-              __html: buildGoogleConsentInitScript(gaMeasurementId),
-            }}
-          />
+          <>
+            {/* eslint-disable-next-line @next/next/no-sync-scripts -- Must run before the GA loader so consent defaults are queued before gtag.js executes. */}
+            <script id='ga-consent-init' src='/ga-consent-init.js' />
+          </>
         ) : null}
       </head>
       <body className={bodyClassName}>
--- a/apps/web/components/providers/GoogleAnalytics.test.tsx
+++ b/apps/web/components/providers/GoogleAnalytics.test.tsx
@@ -1,0 +1,98 @@
+import { render } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { GoogleAnalytics } from './GoogleAnalytics';
+
+type AnalyticsWindow = Window & {
+  dataLayer?: unknown[][];
+  gtag?: ReturnType<typeof vi.fn>;
+};
+
+const mockPublicEnv = vi.hoisted(() => ({
+  NEXT_PUBLIC_GA_MEASUREMENT_ID: 'G-TMY7Z8HK47' as string | undefined,
+}));
+
+vi.mock('next/script', () => ({
+  default: ({
+    children,
+    id,
+    src,
+  }: {
+    readonly children?: React.ReactNode;
+    readonly id?: string;
+    readonly src?: string;
+  }) => (
+    <div data-testid={id} data-src={src}>
+      {children}
+    </div>
+  ),
+}));
+
+vi.mock('@/lib/env-client', () => ({
+  env: {
+    IS_E2E: false,
+    IS_TEST: false,
+  },
+}));
+
+vi.mock('@/lib/env-public', () => ({
+  publicEnv: mockPublicEnv,
+}));
+
+vi.mock('@/lib/demo-recording', () => ({
+  isDemoRecordingClient: () => false,
+}));
+
+describe('GoogleAnalytics', () => {
+  beforeEach(() => {
+    mockPublicEnv.NEXT_PUBLIC_GA_MEASUREMENT_ID = 'G-TMY7Z8HK47';
+    (window as AnalyticsWindow).dataLayer = undefined;
+    (window as AnalyticsWindow).gtag = vi.fn();
+  });
+
+  it('configures GA from the client bundle instead of rendering inline script children', () => {
+    const { queryByTestId, getByTestId } = render(<GoogleAnalytics />);
+
+    expect(getByTestId('ga-gtag-loader')).toHaveAttribute(
+      'data-src',
+      'https://www.googletagmanager.com/gtag/js?id=G-TMY7Z8HK47'
+    );
+    expect(queryByTestId('ga-config')).not.toBeInTheDocument();
+    expect((window as AnalyticsWindow).gtag).toHaveBeenCalledWith(
+      'js',
+      expect.any(Date)
+    );
+    expect((window as AnalyticsWindow).gtag).toHaveBeenCalledWith(
+      'config',
+      'G-TMY7Z8HK47'
+    );
+  });
+
+  it('queues GA configuration in dataLayer when gtag is not loaded yet', () => {
+    (window as AnalyticsWindow).gtag = undefined;
+
+    const { getByTestId } = render(<GoogleAnalytics />);
+
+    expect(getByTestId('ga-gtag-loader')).toHaveAttribute(
+      'data-src',
+      'https://www.googletagmanager.com/gtag/js?id=G-TMY7Z8HK47'
+    );
+    expect((window as AnalyticsWindow).dataLayer?.[0]?.[0]).toBe('js');
+    expect((window as AnalyticsWindow).dataLayer?.[0]?.[1]).toBeInstanceOf(
+      Date
+    );
+    expect((window as AnalyticsWindow).dataLayer?.[1]).toEqual([
+      'config',
+      'G-TMY7Z8HK47',
+    ]);
+  });
+
+  it('skips rendering and configuration when the GA measurement ID is missing', () => {
+    mockPublicEnv.NEXT_PUBLIC_GA_MEASUREMENT_ID = undefined;
+
+    const { queryByTestId } = render(<GoogleAnalytics />);
+
+    expect(queryByTestId('ga-gtag-loader')).not.toBeInTheDocument();
+    expect(queryByTestId('ga-config')).not.toBeInTheDocument();
+    expect((window as AnalyticsWindow).gtag).not.toHaveBeenCalled();
+  });
+});
--- a/apps/web/components/providers/GoogleAnalytics.tsx
+++ b/apps/web/components/providers/GoogleAnalytics.tsx
@@ -8,7 +8,6 @@ import { env } from '@/lib/env-client';
 import { publicEnv } from '@/lib/env-public';
 import {
   applyGoogleConsentMode,
-  buildGoogleAnalyticsConfigScript,
   consentToGoogleConsentMode,
   isCookieBannerRequiredFromClient,
   isValidGaMeasurementId,
@@ -24,6 +23,11 @@ function isConsentPayload(value: unknown): value is Consent {
   );
 }
 
+type AnalyticsWindow = Window & {
+  dataLayer?: unknown[][];
+  gtag?: (...args: unknown[]) => void;
+};
+
 export function GoogleAnalytics() {
   const measurementId = publicEnv.NEXT_PUBLIC_GA_MEASUREMENT_ID;
   const isPassive = env.IS_TEST || env.IS_E2E;
@@ -33,6 +37,16 @@ export function GoogleAnalytics() {
   useEffect(() => {
     if (skip) return;
     if (globalThis.window === undefined) return;
+
+    const analyticsWindow = globalThis.window as AnalyticsWindow;
+    const gtag =
+      analyticsWindow.gtag ??
+      ((...args: unknown[]) => {
+        analyticsWindow.dataLayer = analyticsWindow.dataLayer ?? [];
+        analyticsWindow.dataLayer.push(args);
+      });
+    gtag('js', new Date());
+    gtag('config', measurementId);
 
     const syncConsent = (value: unknown) => {
       if (!isConsentPayload(value)) return;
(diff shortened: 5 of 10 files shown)


itstimwhite:
We are given a diff for PR #13049. We need to check for:
 1. Logic errors
 2. Security issues
 3. Data integrity
 4. Error handling
 5. Cross-file impact

 Let's go through each changed file and note any issues.

 However, note that the PR description says:
   - Move `ga-consent-init` out of root-layout inline HTML and into a static same-origin `/ga-consent-init.js` route allowed by existing `script-src 'self'`.
   - Remove the inline `ga-config` Next script; configure GA from the client bundle and queue through `dataLayer` if `gtag` is not loaded yet.

 We'll check the changes accordingly.

 Step-by-step:

 1. apps/web/app/ga-consent-init.js/route.test.ts (new file)
    - Tests the new route for serving the GA consent bootstrap.
    - Checks headers: Content-Type, Cache-Control, X-Content-Type-Options.
    - Checks that the body contains the expected string.

    No obvious issues.

 2. apps/web/app/ga-consent-init.js/route.ts (new file)
    - Imports `buildGoogleConsentInitScript` from '@/lib/tracking/google-consent-mode'
    - Sets `export const dynamic = 'force-static';`
    - GET handler returns a Response with the script and headers.

    Note: The function `buildGoogleConsentInitScript` is called without arguments in the route handler.
    But looking at the change in `google-consent-mode.ts` (below), we see that the function now accepts an optional `measurementId`.

    In the route handler, we are calling `buildGoogleConsentInitScript()` without any argument.

    Let's check the function in `google-consent-mode.ts`:

    Before the change in that file:
        export function buildGoogleConsentInitScript(measurementId: string): string { ... }

    After the change:
        export function buildGoogleConsentInitScript(measurementId?: string): string {
          if (measurementId !== undefined && !isValidGaMeasurementId(measurementId)) {
            throw new Error('Invalid GA measurement ID');
          }
          ... // rest of the function

    So when called without argument, `measurementId` is `undefined`, and the condition `measurementId !== undefined` is false, so we skip the validation.

    Then the function proceeds to build the script. We must check what the function does when `measurementId` is undefined.

    Looking at the function body (not shown in the diff, but we can infer from the test and the fact that it's used in the route):

    The test in `ga-consent-init.js/route.test.ts` expects the body to contain:
        "gtag('consent', 'default'"
        and 'jv_cc_required=1'

    And the test in `tracking/google-consent-mode.test.ts` for the new case:
        it('builds a measurement-independent consent init script for the CSP-safe route', () => {
          expect(buildGoogleConsentInitScript()).toContain(
            "gtag('consent', 'default'"
          );
        });

    So the function must be building a script that does not depend on the measurement ID for the consent init part.

    However, note that the original inline script in the root layout was built with the measurementId (from the context of the layout). Now we are building a script without the measurementId.

    Is that correct? The consent init script should not need the measurement ID? Let's see what the function does.

    Since we don't have the full function body, we must rely on the tests and the context.

    The test in the route expects the string "jv_cc_required=1", which is likely part of the consent settings.

    The function `buildGoogleConsentInitScript` when called without a measurementId might be building a script that sets default consent without specifying the GA measurement ID? That might be acceptable because the consent mode is independent of the specific GA property? Actually, the consent command in gtag does not require the measurement ID.

    The gtag command for consent is:
        gtag('consent', 'default', { ... });

    So it doesn't need the measurement ID.

    Therefore, it seems safe.

    However, note that the route is static and does not have access to the request (so we cannot get the measurementId from the request or environment). But the measurementId is a public environment variable (NEXT_PUBLIC_GA_MEASUREMENT_ID). Since the route is static, it is built at build time and will have the value of the environment variable at build time.

    But wait: the route is defined as `force-static`, meaning it is pre-rendered at build time. Therefore, it will use the value of `publicEnv.NEXT_PUBLIC_GA_MEASUREMENT_ID` at build time.

    However, in the route handler we are not passing any argument to `buildGoogleConsentInitScript`. We are relying on the function's default parameter (undefined) and then the function does not validate and builds a script without the measurementId.

    But note: the function `buildGoogleConsentInitScript` in `google-consent-mode.ts` now has an optional parameter. We are not using the environment variable in the route at all.

    This is a problem because the consent init script might need to know the measurement ID? Actually, looking at the gtag consent command, it does not require the measurement ID. The consent settings are applied to all GA properties on the page.

    However, the original inline script was built with the measurementId (from the layout) and passed to the function. Now we are not passing it.

    Let me check the original function (before the change) in `google-consent-mode.ts`:

        export function buildGoogleConsentInitScript(measurementId: string): string {
          if (!isValidGaMeasurementId(measurementId)) {
            throw new Error('Invalid GA measurement ID');
          }
          ... // then it built a string that included the measurementId? 

    We don't have the old body, but the test in the route now expects the string to contain "jv_cc_required=1", which is likely a cookie name that is independent of the measurement ID.

    However, note that the function might have been using the measurementId to build a script that sets the consent for that specific GA property? Actually, the consent command in gtag is global.

    But wait: the function `buildGoogleConsentInitScript` might have been building more than just the consent command? It might have been building the entire config? 

    Let's look at the change in the GoogleAnalytics component: we removed the `buildGoogleAnalyticsConfigScript` and replaced it with calling `gtag('config', measurementId)` in the component.

    And the consent init script is only for setting the default consent.

    Therefore, the consent init script does not need the measurement ID.

    So it is safe to call the function without the measurementId.

    However, note that the function now has an optional parameter and we are not passing it. We are relying on the function to work without it.

    The function now has:
        if (measurementId !== undefined && !isValidGaMeasurementId(measurementId)) {
          throw new Error('Invalid GA measurement ID');
        }

    So if measurementId is undefined, we skip the validation and proceed.

    We must check what the function does when measurementId is undefined. Since we don't have the full function, we rely on the test.

    The test in `tracking/google-consent-mode.test.ts` for the new case passes, so we assume it builds a valid script.

    Therefore, no issue here.

 3. apps/web/app/layout.tsx
    - Removed the import of `buildGoogleConsentInitScript` from '@/lib/tracking/google-consent-mode'
    - In the head, replaced the inline script (with dangerouslySetInnerHTML) that called `buildGoogleConsentInitScript(gaMeasurementId)` with a script tag that has src='/ga-consent-init.js'

    This change moves to a static script tag: <script id='ga-consent-init' src='/ga-consent-init.js' />

    This is the main change: moving from an inline script to an external same-origin script.

    We must check:
      - The removal of the inline script avoids the CSP violation (since we now have script-src 'self' and the script is from the same origin).
      - The script tag has an id and a src, and we removed the dangerouslySetInnerHTML.

    Also note: the comment about disabling the next/script rule for no-sync-scripts is still there, but now it's for the external script? Actually, the comment says:
        {/* eslint-disable-next-line @next/next/no-sync-scripts -- Must run before the GA loader so consent defaults are queued before gtag.js executes. */}

    This is acceptable because the script is external but still synchronous? Actually, the script tag without async or defer is synchronous. But note: we are putting it in the head, and it is a same-origin script. The browser will block parsing until it loads and executes.

    However, the comment says it must run before the GA loader (which is loaded via next/script with strategy='afterInteractive'). So it is correct to be synchronous and in the head.

    No obvious issues.

 4. apps/web/components/providers/GoogleAnalytics.test.tsx (new file)
    - Tests the GoogleAnalytics component.
    - Mocks next/script to render a div with data-testid and data-src.
    - Tests three cases:
        a. Configures GA from the client bundle (when gtag is present) -> calls gtag('js') and gtag('config')
        b. Queues GA configuration in dataLayer when gtag is not loaded yet -> pushes to dataLayer
        c. Skips when measurement ID is missing.

    The tests look correct.

    However, note that in the component we now have:

        const analyticsWindow = globalThis.window as AnalyticsWindow;
        const gtag =
          analyticsWindow.gtag ??
          ((...args: unknown[]) => {
            analyticsWindow.dataLayer = analyticsWindow.dataLayer ?? [];
            analyticsWindow.dataLayer.push(args);
          });

    And then we call:
        gtag('js', new Date());
        gtag('config', measurementId);

    This is the same as before? Previously, we had a script that would run the config after the gtag loader. Now we are running it immediately in the component.

    But note: the component is rendered on the client (because we check `if (globalThis.window === undefined) return;`). And we are using the `Script` component for the gtag loader with strategy='afterInteractive'. So the gtag loader will load asynchronously after interactive.

    However, we are calling `gtag` immediately in the component. If gtag is not loaded yet (which it won't be because the loader is afterInteractive and we are in the initial render), then we will push to dataLayer.

    Then when the gtag loader eventually loads, it will process the dataLayer.

    This is the standard way to initialize GA: push the initial commands to dataLayer and then load gtag.js, which will read the dataLayer.

    So it is correct.

    But note: we are calling `gtag('js', new Date())` and `gtag('config', measurementId)` immediately. This is the same as what the old `buildGoogleAnalyticsConfigScript` was doing? Actually, the old script was:

        <Script id='ga-config' strategy='afterInteractive'>
          {buildGoogleAnalyticsConfigScript(measurementId)}
        </Script>

    And `buildGoogleAnalyticsConfigScript` likely returned a string that contained the gtag commands for 'js' and 'config'.

    Now we are doing the same thing but in the component's useEffect.

    However, note that the component runs on every render? We have the useEffect with dependencies [measurementId, skip]. So if the measurementId changes, we will re-run the GA config.

    But the measurementId is from an environment variable and should not change. So it's safe.

    Also note: we are removing the event listener and the subscription in the cleanup.

    The tests pass, so we assume it's correct.

 5. apps/web/components/providers/GoogleAnalytics.tsx
    - Removed the import of `buildGoogleAnalyticsConfigScript`
    - Added a type `AnalyticsWindow` for window augmentation.
    - In the useEffect, we now directly call gtag (or push to dataLayer if gtag is not available) for 'js' and 'config'.
    - Changed the dependency array of the useEffect to [measurementId, skip] (was [skip] before).
    - In the return, we only render the gtag loader script (removed the ga-config script).

    We must check:
      - The useEffect now runs when measurementId changes. Since measurementId is from a constant (publicEnv) and should not change, it's safe. But if it did change, we would re-initialize GA. That might be acceptable? However, note that the measurementId is set at build time and should be constant.

      - We are calling `gtag` twice: once for 'js' and once for 'config'. This is the same as before.

      - We are not handling errors from gtag? But gtag is a third-party function and we don't expect it to throw. If it does, it would break the app. However, that is the same as before.

      - We are not catching errors in the useEffect. If an error occurs, it would be caught by React's error boundary? But we don't have one specified. However, this is the same as before.

    One potential issue: the useEffect runs on the client only (because we check for window). But note that the component is rendered on the server? We have:

        if (globalThis.window === undefined) return;

    So the useEffect only runs on the client.

    This is correct.

 6. apps/web/lib/tracking/google-consent-mode.ts
    - Changed the function `buildGoogleConsentInitScript` to accept an optional `measurementId` (default undefined).
    - Changed the validation to only run if measurementId is defined.

    We already discussed: the route calls it without argument, so we skip validation and build a script without the measurementId.

    We must check that the function still works when measurementId is provided (for example, if we ever call it with a measurementId). The validation is now conditional.

    Also note: the function might be used elsewhere? We removed the import in the layout, but we should check if it's used anywhere else.

    The diff shows that we removed the import in the layout and added a test for the function without argument.

    We don't see any other changes to the function's usage in this PR.

    However, note that the function is also used in the route (without argument) and in the test (without argument). We don't see any call with an argument in this PR.

    But the function is still available for use with an argument (if needed elsewhere). The change makes it more flexible.

    No obvious issues.

 7. apps/web/tests/unit/app/root-layout-scripts.test.ts
    - Added a test that checks the root layout for the script tag with src='/ga-consent-init.js' and id='ga-consent-init'
    - Also checks that the layout does not contain the inline script (by checking for the absence of certain strings).

    This test is to ensure we removed the inline script and replaced it with the external one.

    Looks correct.

 8. apps/web/tests/unit/lib/content-security-policy.test.ts
    - Added an expectation that the script-src directive contains "'self'".

    This is to ensure that our CSP policy allows same-origin scripts, which is necessary for the new `/ga-consent-init.js` route.

    Looks correct.

 9. apps/web/tests/unit/tracking/google-consent-mode.test.ts
    - Added a test for `buildGoogleConsentInitScript()` without argument, expecting it to contain the string for setting consent default.

    We already discussed: this is to ensure the function works without a measurementId.

    Looks correct.

 Potential issue 1: In the route handler for `/ga-consent-init.js`, we are calling `buildGoogleConsentInitScript()` without any argument. We changed the function to accept an optional parameter. However, note that the function is imported from '@/lib/tracking/google-consent-mode'. We must make sure that the function is indeed defined with an optional parameter.

    We see in the diff for `google-consent-mode.ts` that the function signature changed to:
        export function buildGoogleConsentInitScript(measurementId?: string): string {

    So it is optional.

 Potential issue 2: The route is marked as `export const dynamic = 'force-static';`. This means it will be pre-rendered at build time. However, the function `buildGoogleConsentInitScript` might depend on environment variables that are only available at runtime? 

    But note: the function does not use any environment variables when called without argument (because we skip the validation and then build a string that does not include the measurementId). 

    However, we must look at the body of the function (which we don't have in the diff). But the test in the route expects the string to contain "jv_cc_required=1", which is a hardcoded string? Or is it built from something?

    Without the function body, we cannot be 100% sure. But the test passes, so we assume it's a static string.

    However, note that the function might have been using the measurementId to build a string that included it? Now we are not passing it, so if the function body used the measurementId, we would be getting an undefined.

    But we changed the function to not require the measurementId and to skip validation when it's undefined. And we also changed the function to build a script that does not depend on the measurementId? 

    The test in the route and the new test in the tracking module both pass without an argument, so we assume the function body is safe.

 Potential issue 3: In the GoogleAnalytics component, we are now calling gtag immediately in the useEffect. Previously, we had a script that would run afterInteractive. Now we are running it as soon as the component mounts (which is on the client, after the initial HTML is rendered). 

    This is acceptable because we are pushing to the dataLayer if gtag is not available, and then the gtag loader (which is loaded via the Script component with strategy='afterInteractive') will pick up the dataLayer.

    However, note: the Script component for the gtag loader is still in the return of the component. So the order of operations:

        1. The component renders and returns the Script loader (which is not executed yet because it's afterInteractive).
        2. The useEffect runs (on the client) and calls gtag

itstimwhite (commented):

PR #13055, opened 2026-07-04
feat(hermes): add gbrain health summary cron

### LINKED ISSUE #13048: Increase gbrain usefulness: add data connectors, knowledge-compounding crons, and usage research
Tim wants gbrain to be more useful: get more data in it via connectors, research how to use it effectively, and build crons that compound company knowledge (e.g., daily digests, agent health summaries, metric updates).

Deliverables:

* Research document: best practices for gbrain usage, potential connectors (GitHub, Linear, Stripe, CRM, etc.), and cron ideas that feed knowledge back into gbrain.
* Proposed cron jobs (with schedules, scripts, and delivery channels) that:
  * Pull daily agent performance/health from Hermes/OpenClaw and insert as gbrain notes or timeline entries.
  * Weekly industry trend scrape (Exa/web) → summarize → add as gbrain notes.
  * Monthly KR/metric snapshot from internal systems (e.g., MRR, activation) → store as gbrain structured claims.
  * Hourly gbrain health check → post alert to Telegram/Slack if degradation.
* Implement one starter cron (e.g., daily gbrain health summary posted to Telegram) to validate the pipeline.
* Update documentation (e.g., in Ovie/AGENTS.md or Hermes config) on how to leverage gbrain for decision‑making.

Links: gbrain MCP is already configured; we can use mcp_gbrain_* tools from agents or cron jobs.

itstimwhite:
## Summary
- Add a Hermes-Air `gbrain-health-summary` job that probes `gbrain doctor` and a bounded search smoke, writes `ops/gbrain-health/latest`, and notifies ops through the existing Telegram/Slack helper.
- Add the launchd template, package script, runbook/cron registry entries, and a compact spike covering connector candidates plus proposed knowledge-compounding jobs.
- Cover healthy, degraded, down, write, and notification paths with focused script tests.

## Cost Impact
- **External API calls**: 0 new external calls in the starter job. The job uses local `gbrain` CLI calls and existing ops notification helpers.
- **Monthly projection**: ~60 local gbrain CLI calls/month at daily cadence (2 calls/run x ~30 runs/month).
- **Scaling factor**: O(1) per run.
- **Monthly cost estimate**: $0 for the starter job. Proposed Web/Exa research remains documented only and requires an explicit provider budget before enabling.

## Verification
- `pnpm install --frozen-lockfile` — passed
- `pnpm biome check --write scripts/hermes/jobs/gbrain-health-summary.ts scripts/lib/__tests__/gbrain-health-summary.test.mjs docs/spikes/gbrain-knowledge-compounding.md docs/CRON_REGISTRY.md docs/HERMES_AIR.md scripts/hermes/launchd/README.md package.json` — passed, `Checked 3 files ... No fixes applied`
- `pnpm exec vitest --root scripts --config vitest.config.mts run lib/__tests__/gbrain-health-summary.test.mjs lib/__tests__/agent-config-health.test.mjs lib/__tests__/pipeline-scoreboard.test.mjs` — passed, `Test Files 3 passed (3); Tests 22 passed (22)`
- `pnpm run typecheck` — passed, `Tasks: 3 successful, 3 total`
- `pnpm --filter @jovie/web run typecheck -- --pretty false` — passed
- Runtime healthy smoke with fake `gbrain` — passed; output included `GBrain health summary: healthy` and `search returned output` without echoing the fake sensitive search payload
- Runtime failure smoke with `HERMES_GBRAIN_BIN=/usr/bin/false` — passed; output included `GBrain health summary: down` and `failure_exit=1`
- `git diff --check` — passed
- Pre-push hook — passed: Biome changed-files check, `next:proxy-guard`, `tailwind:check`, and web typecheck
- Secret scans in pre-commit — passed: gitleaks and TruffleHog reported no findings

## CodeRabbit
- Ran `coderabbit review --agent -c AGENTS.md -t uncommitted` repeatedly after quota delays.
- Fixed valid findings: notification failures are logged separately after `summarized`, docs use `ops/gbrain-health/latest`, runbook avoids rerunning the side-effecting job for status checks, stderr log is included, fallback doctor parsing handles negative phrases, empty search degrades, and launchd schedule wording matches the registry.
- Skipped two final findings as not applicable: Hermes local jobs follow existing `process.env.HERMES_GBRAIN_BIN` pattern rather than app `lib/env.ts`, and notification failures should not turn a healthy gbrain summary into a false gbrain-health failure.

Fixes #13048

### DIFF
--- a/docs/CRON_REGISTRY.md
+++ b/docs/CRON_REGISTRY.md
@@ -32,6 +32,7 @@ These are machine-local Hermes jobs, not Vercel production crons. They run from 
 | Unit | Schedule | Purpose | Source |
 |------|----------|---------|--------|
 | `co.jovie.hermes.cron-pipeline-scoreboard` | every 60 min | Computes the daily codex shipping funnel, writes `pipeline-scoreboard-latest.json` + gbrain `ops/pipeline-scoreboard/latest`, and alerts on 12h shipper stalls. | `scripts/hermes/jobs/pipeline-scoreboard.ts` |
+| `co.jovie.hermes.cron-gbrain-health-summary` | 07:15 local daily | Runs bounded gbrain doctor/search probes, writes `ops/gbrain-health/latest`, and posts the summary to Telegram/Slack through existing ops notification helpers. | `scripts/hermes/jobs/gbrain-health-summary.ts` |
 
 ## Production Schedule
 
--- a/docs/HERMES_AIR.md
+++ b/docs/HERMES_AIR.md
@@ -131,6 +131,11 @@ tail -50 ~/.hermes/logs/codex-issue-shipper/*.log 2>/dev/null  # after a non-dry
 tsx scripts/hermes/jobs/agent-config-health.ts
 tail -50 ~/.hermes/logs/launchd/cron-agent-config-health.err.log
 
+# Latest gbrain health summary written by Hermes
+gbrain search "ops/gbrain-health/latest" --limit 3
+tail -50 ~/.hermes/logs/launchd/cron-gbrain-health-summary.log \
+  ~/.hermes/logs/launchd/cron-gbrain-health-summary.err.log
+
 # Free-model rankings
 cat ~/.hermes/state/model-router-rankings.json | jq .
 ```
--- a/docs/spikes/gbrain-knowledge-compounding.md
+++ b/docs/spikes/gbrain-knowledge-compounding.md
@@ -1,0 +1,26 @@
+# GBrain Knowledge Compounding Spike
+
+> Issue: [#13048](https://github.com/JovieInc/Jovie/issues/13048)
+
+Goal: turn recurring source data into durable gbrain summaries with source
+links, timestamps, operational facts, and stable decision records.
+
+## Connectors
+
+Sources: GitHub, Linear, Hermes/OpenClaw, Stripe reports, CRM/Airtable,
+Slack/Telegram, Sentry/Vercel, and Web/Exa. Guardrails: bounded summarized
+notes, no raw logs/customer polling/contact dumps, and explicit web budget.
+
+## Proposed Jobs
+
+| Job | Schedule | Script | Delivery | Cost |
+|---|---:|---|---|---|
+| GBrain health summary | Daily 07:15 local | `scripts/hermes/jobs/gbrain-health-summary.ts` | Telegram/Slack + `ops/gbrain-health/latest` | Two local CLI calls/day |
+| Agent performance digest | Daily 07:30 local | Extend `pipeline-scoreboard.ts` or sibling digest | Telegram + `ops/agent-health/latest` | Local logs unless GitHub enrichment is added |
+| Hourly degradation alert | Hourly | Reuse health summary with transition state | Telegram/Slack on change | Local CLI calls |
+| Weekly trend scrape | Mondays 08:00 local | New Exa/web research job | GBrain + Telegram excerpt | Needs approved provider budget |
+| Monthly KR snapshot | First day 08:00 local | New metrics collector | GBrain structured claims | Prefer existing reports |
+
+No Vercel cron is added. These are Hermes-Air control-plane jobs, not product
+traffic. Decision flow: search gbrain, read current repo docs, make a quantified
+Ship now / Re-evaluate when / Then decision, then write the durable learning.
--- a/package.json
+++ b/package.json
@@ -170,6 +170,7 @@
     "ci:harness:test": "vitest --root scripts --config vitest.config.mts run lib/__tests__",
     "doc:freshness:check": "node scripts/doc-freshness-lint.mjs",
     "doc:freshness:test": "vitest --root scripts --config vitest.config.mts run lib/__tests__/doc-freshness.test.mjs",
+    "hermes:gbrain-health": "tsx scripts/hermes/jobs/gbrain-health-summary.ts",
     "doc:gardening:run": "node scripts/doc-gardening-agent.mjs",
     "qa:swarm:list": "node scripts/qa-swarm/cli.mjs list",
     "qa:swarm:propose": "node scripts/qa-swarm/cli.mjs propose",
--- a/scripts/hermes/jobs/gbrain-health-summary.ts
+++ b/scripts/hermes/jobs/gbrain-health-summary.ts
@@ -1,0 +1,252 @@
+#!/usr/bin/env tsx
+import { execFileSync } from 'node:child_process';
+import { pathToFileURL } from 'node:url';
+
+import { gbrainLearn } from '../lib/gbrain';
+import { logJobEvent, withJobLogging } from '../lib/jobs-log';
+import { notifyOps } from '../lib/ops-notify';
+
+const JOB = 'gbrain-health-summary';
+const GBRAIN = process.env.HERMES_GBRAIN_BIN ?? 'gbrain';
+
+type HealthStatus = 'healthy' | 'degraded' | 'down';
+type HealthCheck = {
+  readonly name: string;
+  readonly ok: boolean;
+  readonly detail: string;
+  readonly durationMs: number;
+};
+
+type HealthSummary = {
+  readonly status: HealthStatus;
+  readonly generatedAt: string;
+  readonly checks: readonly HealthCheck[];
+  readonly recommendation: string;
+};
+
+type GBrainExec = (
+  file: string,
+  args: readonly string[],
+  options: {
+    readonly encoding: 'utf8';
+    readonly timeout: number;
+    readonly maxBuffer?: number;
+  }
+) => string;
+
+function summarizeOutput(output: string, maxChars = 700): string {
+  const cleaned = output.replace(/\s+/g, ' ').trim();
+  if (!cleaned) return 'empty output';
+  return cleaned.length > maxChars
+    ? `${cleaned.slice(0, maxChars)}...`
+    : cleaned;
+}
+
+function statusFromDoctor(output: string): { ok: boolean; detail: string } {
+  const text = output.trim();
+  if (!text) return { ok: false, detail: 'doctor returned empty output' };
+  try {
+    const parsed = JSON.parse(text) as Record<string, unknown>;
+    const status = parsed.status;
+    const healthScore = parsed.health_score ?? parsed.healthScore;
+    const ok =
+      status === 'ok' ||
+      status === 'healthy' ||
+      (typeof healthScore === 'number' && healthScore >= 0.8);
+    if (typeof status !== 'string')
+      return { ok, detail: summarizeOutput(text) };
+    const score =
+      typeof healthScore === 'number' ? ` health_score=${healthScore}` : '';
+    return { ok, detail: `status=${status}${score}` };
+  } catch {
+    const lower = text.toLowerCase();
+    const negative =
+      /\b(unhealthy|not\s+ok|not\s+healthy|down|degraded|fail(ed|ing)?)\b/.test(
+        lower
+      );
+    return {
+      ok: !negative && (lower.includes('healthy') || lower.includes('ok')),
+      detail: summarizeOutput(text),
+    };
+  }
+}
+
+function timedCheck(
+  name: string,
+  run: () => { readonly ok: boolean; readonly detail: string }
+): HealthCheck {
+  const started = Date.now();
+  try {
+    const result = run();
+    return {
+      name,
+      ok: result.ok,
+      detail: result.detail,
+      durationMs: Date.now() - started,
+    };
+  } catch (err) {
+    return {
+      name,
+      ok: false,
+      detail: err instanceof Error ? err.message : String(err),
+      durationMs: Date.now() - started,
+    };
+  }
+}
+
+export function buildGBrainHealthSummary(args: {
+  readonly generatedAt: string;
+  readonly checks: readonly HealthCheck[];
+}): HealthSummary {
+  const failed = args.checks.filter(check => !check.ok);
+  const status: HealthStatus =
+    args.checks.length === 0 || failed.length === args.checks.length
+      ? 'down'
+      : failed.length === 0
+        ? 'healthy'
+        : 'degraded';
+  const recommendation =
+    status === 'healthy'
+      ? 'No operator action needed.'
+      : 'Check gbrain doctor output, gbrain serve launchd status, and recent Hermes/OpenClaw logs before changing model or connector configuration.';
+
+  return {
+    status,
+    generatedAt: args.generatedAt,
+    checks: args.checks,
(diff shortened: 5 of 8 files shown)


itstimwhite:
### Merge activity

* **Jul 4, 4:44 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 8:07 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 8:08 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.

itstimwhite:
We are given a PR diff for adding a gbrain health summary cron job in the Hermes-Air system.

We are to focus on:
1. Logic errors
2. Security
3. Data integrity
4. Error handling
5. Cross-file impact

We'll go through each changed file and look for issues.

Note: The PR description says that the job uses local `gbrain` CLI calls and existing ops notification helpers.

Let's examine the changes:

1. `docs/CRON_REGISTRY.md`: Added a new cron job entry. No logic, security, etc. issues expected.

2. `docs/HERMES_AIR.md`: Added a section about the latest gbrain health summary. No issues.

3. `docs/spikes/gbrain-knowledge-compounding.md`: New file, a spike document. No code, so no issues.

4. `package.json`: Added a new script `"hermes:gbrain-health": "tsx scripts/hermes/jobs/gbrain-health-summary.ts"`. No issues.

5. `scripts/hermes/jobs/gbrain-health-summary.ts`: The main implementation.

6. `scripts/hermes/launchd/README.md`: Updated to include the new plist template. No issues.

7. `scripts/hermes/launchd/co.jovie.hermes.cron-gbrain-health-summary.plist.template`: New launchd template. No issues.

8. `scripts/lib/__tests__/gbrain-health-summary.test.mjs`: New test file. We'll check for test issues, but note that the PR says tests passed.

Now, let's focus on the main implementation file: `scripts/hermes/jobs/gbrain-health-summary.ts`

We'll look for:

- Logic errors: incorrect conditionals, off-by-one, race conditions (though this is a script, so less likely for races, but note file writes and such), null dereferences.
- Security: hardcoded secrets, injection (though we are using `execFileSync` with fixed arguments, so command injection is unlikely if the arguments are fixed and the binary is trusted), but note we are using environment variable for the binary path.
- Data integrity: when writing to gbrain, we are using `gbrainLearn` which is imported from `../lib/gbrain`. We assume that function is correct, but we should check if there are any issues in how we call it.
- Error handling: we have try/catch in the notification part, but what about other parts?
- Cross-file impact: we are using `gbrainLearn`, `logJobEvent`, `withJobLogging`, `notifyOps`. We assume these are existing and correct, but we should check if our usage matches their contracts.

Let's go through the main file line by line (as per the diff):

We note the file is new, so we are looking at the entire content.

Key points:

- The script uses `execFileSync` to run the `gbrain` command. We are setting a timeout and maxBuffer.

- We have a function `statusFromDoctor` that tries to parse the output of `gbrain doctor --fast --json` as JSON. If it fails, it falls back to checking for negative phrases.

- We have a function `timedCheck` that runs a check and returns a HealthCheck object with duration.

- We have `collectGBrainHealthSummary` that runs two checks: doctor and search.

- We have `buildGBrainHealthSummary` that determines the overall status (healthy, degraded, down) based on the checks.

- We have `renderGBrainHealthSummary` that formats the summary for output.

- We have `runGBrainHealthSummary` that:
   a. Collects the summary.
   b. Renders it to a string.
   c. Writes it to gbrain (via `gbrainLearn`) unless `writeToGBrain` is false.
   d. Logs the job event.
   e. If `notify` is not false, it tries to notify ops (via `notifyOps`) and catches any error, logging it as a notify-error.

- The `main` function runs the job with logging and exits with code 1 if the status is down or if writing to gbrain failed.

1. **Logic errors**:
   - In `statusFromDoctor`: 
        We try to parse the output as JSON. If it's valid JSON, we check for status or health_score.
        If it's not valid JSON, we check for negative phrases in the lowercased text.
        However, note that the condition for `ok` in the fallback is:
          `ok: !negative && (lower.includes('healthy') || lower.includes('ok'))`
        This means that if the output contains the word "healthy" or "ok" and does not contain any negative phrase (as defined by the regex), then it's considered ok.
        But what if the output is something like "The system is healthy but there is a warning"? It would be considered ok because it has "healthy" and no negative phrase? 
        However, the regex for negative phrases includes: unhealthy, not ok, not healthy, down, degraded, fail(ed|ing)?.
        So if the output is "The system is healthy but there is a warning", it doesn't match the negative regex, and it includes "healthy", so ok=true.

        This might be acceptable, but note that the doctor output might be in a format that we don't expect. However, the function is designed to be lenient.

        One potential issue: the regex for negative phrases might not catch all negative states. But that's a design choice.

   - In `buildGBrainHealthSummary`:
        We set the status as:
          if no checks or all checks failed -> 'down'
          else if no checks failed -> 'healthy'
          else -> 'degraded'

        This seems correct.

   - In `runGBrainHealthSummary`:
        We call `gbrainLearn` to write the summary. We then set `gbrainOk` to the return value of `gbrainLearn` (unless `writeToGBrain` is false, then we set it to true).
        Then we log the job event with `gbrainOk`.
        Then, if `notify` is not false, we try to notify ops and catch any error, logging it as a notify-error.

        However, note that if the notification fails, we log the error but we do not change the exit code. The exit code is set in `main` based on:
          if (run.summary.status === 'down' || !run.gbrainOk) {
            process.exitCode = 1;
          }

        So if the notification fails, we don't exit with 1 unless the summary was down or gbrain write failed.

        This is intentional per the PR description: "notification failures should not turn a healthy gbrain summary into a false gbrain-health failure."

        So that's by design.

   - In `main`:
        We use `withJobLogging` which we assume handles logging and catches errors? Actually, `withJobLogging` is a higher-order function that wraps the async function and logs job start and end. We don't see its implementation, but we assume it's correct.

        Then we run `runGBrainHealthSummary` and write the body to stdout.

        Then we set the exit code to 1 if the status is down or if gbrain write failed.

        This seems correct.

2. **Security**:
   - We are using `execFileSync` with a binary path from the environment variable `HERMES_GBRAIN_BIN` (defaulting to 'gbrain'). 
     The arguments are fixed arrays: 
        For doctor: ['doctor', '--fast', '--json']
        For search: ['search', 'gbrain health summary smoke', '--limit', '1']
     So there is no risk of command injection because we are not concatenating user input into the command.

   - However, note that the environment variable `HERMES_GBRAIN_BIN` could be set to a malicious binary. But that is a deployment/configuration issue, not a code issue. The code assumes that the binary is trusted.

   - We are not using any user input in the search query: it's fixed to 'gbrain health summary smoke'.

   - We are writing to gbrain via `gbrainLearn`, which we assume is secure.

   - We are not returning any sensitive data in the output? The search output is summarized and truncated to 700 chars, and we are not logging the raw search output in the notification? Actually, we are including the rendered summary in the notification. The rendered summary includes the check details, which for the search check is either 'search returned output' or 'empty output'. We are not including the raw search results. So no leak.

   - Therefore, no security issues found.

3. **Data integrity**:
   - We are writing a gbrain page with the slug 'ops/gbrain-health/latest'. We are overwriting the latest health summary each time. This is by design.

   - We are using `gbrainLearn` to write the page. We assume that function handles conflicts correctly (e.g., by overwriting). 

   - We are not interacting with a database directly, so no SQL injection.

   - We are not validating the output of the gbrain commands beyond what we do in `statusFromDoctor` and the search check. But note that the search check only checks if the output is non-empty. We don't validate that the search result is actually about gbrain health. However, the search query is fixed, so it should be safe.

   - Therefore, no data integrity issues found.

4. **Error handling**:
   - In `statusFromDoctor`: we have a try/catch for JSON.parse. If it fails, we fall back to string matching. This is good.

   - In `timedCheck`: we have a try/catch that catches any error from the `run` function and returns a HealthCheck with ok=false and the error message. This is good.

   - In `runGBrainHealthSummary`: 
        We call `gbrainLearn` without a try/catch. If `gbrainLearn` throws, it will propagate up and be caught by the `withJobLogging` wrapper? We don't see the implementation of `withJobLogging`, but note that in `main` we have:
          await withJobLogging(JOB, async () => {
            const run = await runGBrainHealthSummary();
            ...
          });
        And then we have a catch in the top-level of `main` that logs and exits.

        However, note that `runGBrainHealthSummary` is async and we are awaiting it. If `gbrainLearn` throws, it will be caught by the `withJobLogging` wrapper? We don't know what `withJobLogging` does. But note that the PR says the tests passed and the runtime smokes passed.

        Alternatively, we could wrap the `gbrainLearn` call in a try/catch to handle errors and set `gbrainOk` to false and log appropriately. Currently, if `gbrainLearn` throws, it will break out of the function and be caught by the `withJobLogging` (if it catches) or by the top-level in `main`.

        Looking at the `main` function:
          await withJobLogging(JOB, async () => {
            const run = await runGBrainHealthSummary();
            process.stdout.write(`${run.body}\n`);
            if (run.summary.status === 'down' || !run.gbrainOk) {
              process.exitCode = 1;
            }
          });

        If `runGBrainHealthSummary` throws, then the `withJobLogging` will catch it? We don't know. But note that the `withJobLogging` function is imported from `../lib/jobs-log`. We assume it is designed to catch errors and log them appropriately.

        However, to be safe, we might want to catch errors in `runGBrainHealthSummary` and convert them to a failed gbrain write? But note that the current code does not catch errors from `gbrainLearn`. 

        Let's see: if `gbrainLearn` throws, then the assignment to `gbrainOk` will throw and the function will exit early. Then the `logJobEvent` and notification steps will not run. This might be acceptable because we cannot consider the job successful if we couldn't write to gbrain.

        But note: the `main` function checks `!run.gbrainOk` to set the exit code. If we throw before setting `gbrainOk`, then we never get to set the exit code based on that. However, the `withJobLogging` might catch the error and log it, and then the process would exit with an uncaught exception (exit code 1) anyway.

        So it might be acceptable.

        However, the PR description says: "Cover healthy, degraded, down, write, and notification paths with focused script tests." and the tests passed. We should check the test file to see if they test the write path.

        In the test file, we have:
          it('writes a stable gbrain page and notifies ops', async () => {
            const learned = [];
            const notifications = [];
            const exec = (_file, args) =>
              args[0] === 'doctor'
                ? JSON.stringify({ status: 'ok', health_score: 0.91 })
                : 'smoke result';
            const run = await runGBrainHealthSummary({
              exec,
              learn: args => {
                learned.push(args);
                return true;
              },
              notifyOps: async text => {
                notifications.push(text);
              },
              now: new Date(generatedAt),
            });
            expect(run.gbrainOk).toBe(true);
            ... 
          });

        This test passes if `gbrainLearn` (mocked as the `learn` function) returns true. It does not test the case when `gbrainLearn` throws.

        We should add a test for when `gbrainLearn` throws? But note the PR says the tests passed and we are not to change the test file unless we find an issue. However, we are reviewing the code, not the tests.

        But note: the issue we are looking for is in the code. If the code does not handle an error from `gbrainLearn`, then it might be an unhandled rejection? Actually, `gbrainLearn` is synchronous? We don't know. Looking at the import: `import { gbrainLearn } from '../lib/gbrain';` and we are calling it without `await`. So it is synchronous.

        Therefore, if it throws, it will be a synchronous throw and will be caught by the `try` in `withJobLogging`? We don't see the implementation of `withJobLogging`, but if it is a standard async wrapper that catches errors, then it will be caught.

        However, to be safe, we should check if `withJobLogging` catches synchronous errors. Since it wraps an async function, and we are using `await`, then synchronous errors in the async function are turned into rejections and caught by the `await`? Actually, no: if we throw synchronously in an async function, it returns a rejected promise. Then the `await` will throw. So if `withJobLogging` does:

          return await fn();

        then it will catch the error? Not unless it has a try/catch.

        We don't have the code for `withJobLogging`. But note that the PR says the runtime smokes passed, including a failure smoke. So it's likely that `withJobLogging` is implemented to catch errors and log them.

        Given that, we might not need to change the code. However, let's look for other error handling issues.

   - In the notification part: we have a try/catch that catches errors from `notifyOps` and logs them as a notify-error. This is good.

   - We are not handling errors from `logJobEvent`? But that is likely a synchronous function that logs and doesn't throw. We assume it's safe.

   - Therefore, no obvious error handling issues.

5. **Cross-file impact**:
   - We are using `gbrainLearn`, `logJobEvent`, `withJobLogging`, `notifyOps`. We assume these are existing and correct.

   - We are adding a new script and a new launchd template. We are updating the documentation.

   - We are adding a new script in package.json.

   - We are not changing any existing interfaces.

   - Therefore, no cross-file impact issues.

However, note one potential issue in the `statusFromDoctor` function:

  We have:
    const ok =
      status === 'ok' ||
      status === 'healthy' ||
      (typeof healthScore === 'number' && healthScore >= 0.8);

  But note: the `healthScore` might be a string? We are checking `typeof healthScore === 'number'`. If it's a string that can be converted to a number, we are not handling it. However, the JSON parser would have given us a number if it was a number in the JSON. If it's a string, then `typeof healthScore` would be 'string', so we skip the numeric check.

  Then we fall back to the string status check. So if the status is not 'ok' or 'healthy', and the healthScore is a string, we would return `{ ok: false, detail: ... }`? Actually, no: we set `ok` based on the three conditions. If the healthScore is a string, then the third condition fails, and we rely on the status.

  Example: if the JSON is `{ "status": "ok", "healthScore": "0.9" }`, then:
      status === 'ok' -> true, so ok = true.

  Example: if the JSON is `{ "status": "unknown", "healthScore": "0.9" }`, then:
      status is not 'ok' or 'healthy', and healthScore is a string -> so the third condition fails -> ok = false.

  But note: the healthScore string "0.9" is a valid number, but we are not converting it. We could do:

      const scoreNum = typeof healthScore === 'number' ? healthScore : 
                      typeof healthScore === 'string' ? parseFloat(healthScore) : undefined;

  However, the PR says the tests passed. We should check if the test covers this.

  In the test file, we have:
        const exec = (_file, args) =>
          args[0] === 'doctor'
            ? JSON.stringify({ status: 'ok', health_score: 0.95 })
            : '';

  Note: the test uses `health_score` (with an underscore) but our code checks for `healthScore` (camelCase) and `health_score` (snake_case) because we do:
        const healthScore = parsed.health_score ?? parsed.healthScore;

  So we handle both.

  But note: the test uses a number for health_score. What if it's a string? We don't have a test for that.

  However, the issue is: if the healthScore is a string that represents a number >=0.8, we are not recognizing it as ok

itstimwhite (commented):

PR #13090, opened 2026-07-04
docs(gstack): add canonical design skill

### LINKED ISSUE #13043: Consolidate design skills into a single canonical design skill for all agents and Ovie (Eve + Fable 5)
Eve shall use Fable 5 to refactor and improve all existing design-related skills (design-taste-frontend, gstack design-review, design-consultation, design-html, design-shotgun, design, plan-design-review, etc.) into one unified, canonical design skill that all agents (including Ovie) can use. The resulting skill should encapsulate the best practices, audit-first approach, anti-slop rules, and be usable across web, iOS, desktop, and ops tools. Ensure Veronica (design agent) adopts and uses this skill for all UX/design work. Deliverable: new skill module (e.g., design-canonical) installed in appropriate agent skill directories, with documentation and migration notes for existing skills. Link to existing Jovie/Ovie design work and ensure backward compatibility where appropriate the skill is referenced in agent prompts and PR templates.

itstimwhite:
## Summary

- Adds `/design-canonical` as the canonical design operating system for Jovie agents and Ovie.
- Updates legacy design skills (`design-review`, `plan-design-review`, `design-consultation`, `design-shotgun`, `design-html`) to load the canonical contract first while keeping their mode-specific workflows.
- Adds Claude, Codex, and Factory generated host outputs plus a tracked clean-checkout regression test so the ignored skill artifacts cannot silently disappear.
- Updates gstack routing/docs and the PR template to require design-read, dials, checklist, and evidence for design-related changes.

Fixes #13043

## Design Skill Evidence

Design-read: Reading this as: agent-control-plane documentation for autonomous coding agents and Ovie operators, with a rigorous operating-manual language, leaning toward gstack's generated Agent Skills system.

Dials: DESIGN_VARIANCE=low, MOTION_INTENSITY=none, VISUAL_DENSITY=high

Before/after evidence: file/component evidence only; this is a non-UI skill/docs control-plane change. The new tracked artifacts are:
- `.agents/skills/gstack/design-canonical/SKILL.md.tmpl`
- `.agents/skills/gstack/design-canonical/SKILL.md`
- `.agents/skills/gstack/.agents/skills/gstack-design-canonical/SKILL.md`
- `.agents/skills/gstack/.factory/skills/gstack-design-canonical/SKILL.md`
- `.claude/skills/gstack/design-canonical/SKILL.md`
- `.claude/skills/design-canonical/SKILL.md`

Checklist: contrast not applicable (non-UI), states not applicable (non-UI), layout pass (no UI layout changed), motion pass (no motion), icons pass (no icons/brand marks changed), anti-slop pass (canonical skill explicitly codifies anti-slop rules), evidence pass (tracked artifacts + generated freshness + clean-checkout tracking test).

## Verification

- `./scripts/setup.sh` -> pass; Node v22.22.1, pnpm 9.15.4, dependencies installed, Doppler/GitHub auth OK.
- `cd .agents/skills/gstack && bun install` -> pass; installed nested gstack deps including `diff`.
- `cd .agents/skills/gstack && bun run gen:skill-docs --host all --dry-run` -> pass; all Claude, Codex, and Factory outputs reported `FRESH`, including `gstack-design-canonical`.
- `cd .agents/skills/gstack && bun test test/gen-skill-docs.test.ts test/skill-validation.test.ts` -> pass; 661 pass, 0 fail, 3772 assertions.
- `cd .agents/skills/gstack && bun run skill:size-check` -> pass; `Skill size check: ok`.
- `git diff --check` and `git diff --cached --check` -> pass.
- `test -e .claude/skills/design-canonical/SKILL.md && test -e .claude/skills/gstack/design-canonical/SKILL.md && git ls-files --error-unmatch ...` -> pass; `canonical_tracking_and_symlinks_ok`.
- `pnpm --filter @jovie/web run typecheck -- --pretty false` under Node 22 -> pass.
- `git push -u origin HEAD` pre-push hook -> pass; `biome-pre-push` checked 0 configured files, `next:proxy-guard` pass, `tailwind:check` pass, web `typecheck` pass.

Biome note: `pnpm biome check --write ...` on this diff reports `No files were processed` because `.agents`, `.claude`, and the PR template paths are ignored by the repo Biome configuration. Whitespace and TypeScript coverage are covered by `git diff --check`, gstack's Bun tests, and the pre-push hook.

CodeRabbit note: required local CodeRabbit review was attempted three times:
- `coderabbit review --agent -c AGENTS.md -t uncommitted` -> service rate limit, recoverable, wait estimate 8 minutes.
- `coderabbit review --agent -c AGENTS.md -t uncommitted` -> service rate limit, recoverable, wait estimate 7 minutes.
- `coderabbit review --agent -c AGENTS.md -t committed --base main` -> service rate limit, recoverable, wait estimate 7 minutes.
No CodeRabbit findings were returned to fix.

Subagents:
- Testing subagent initially found untracked/ignored canonical artifacts and missing nested gstack deps; fixed by force-tracking canonical artifacts, adding a clean-checkout tracking test, and running `bun install` in `.agents/skills/gstack`.
- Review subagent found the same untracked/broken-symlink issue; fixed.
- Security subagent found the same packaging/symlink integrity risk and no prompt-injection, unsafe tool, secret/prod, or Ovie/GBrain claim issues after review.

## Bug-to-Test

bug-to-test: not applicable — feature/docs/control-plane work, not a bug fix.

### DIFF
--- a/.agents/skills/gstack/.agents/skills/gstack-design-canonical/SKILL.md
+++ b/.agents/skills/gstack/.agents/skills/gstack-design-canonical/SKILL.md
@@ -1,0 +1,556 @@
+---
+name: design-canonical
+description: |
+  Canonical design operating system for all Jovie agents and Ovie. Use first for
+  any UX, product design, visual audit, design system, mockup, or design-plan
+  task; legacy design skills now layer their mode-specific workflow on top of
+  this shared audit-first, anti-slop, cross-platform contract. (gstack)
+---
+<!-- AUTO-GENERATED from SKILL.md.tmpl — do not edit directly -->
+<!-- Regenerate: bun run gen:skill-docs -->
+
+## Preamble (run first)
+
+```bash
+_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
+GSTACK_ROOT="$HOME/.codex/skills/gstack"
+[ -n "$_ROOT" ] && [ -d "$_ROOT/.agents/skills/gstack" ] && GSTACK_ROOT="$_ROOT/.agents/skills/gstack"
+GSTACK_BIN="$GSTACK_ROOT/bin"
+GSTACK_BROWSE="$GSTACK_ROOT/browse/dist"
+GSTACK_DESIGN="$GSTACK_ROOT/design/dist"
+_UPD=$($GSTACK_BIN/gstack-update-check 2>/dev/null || .agents/skills/gstack/bin/gstack-update-check 2>/dev/null || true)
+[ -n "$_UPD" ] && echo "$_UPD" || true
+mkdir -p ~/.gstack/sessions
+touch ~/.gstack/sessions/"$PPID"
+_SESSIONS=$(find ~/.gstack/sessions -mmin -120 -type f 2>/dev/null | wc -l | tr -d ' ')
+find ~/.gstack/sessions -mmin +120 -type f -exec rm {} + 2>/dev/null || true
+_CONTRIB=$($GSTACK_BIN/gstack-config get gstack_contributor 2>/dev/null || true)
+_PROACTIVE=$($GSTACK_BIN/gstack-config get proactive 2>/dev/null || echo "true")
+_PROACTIVE_PROMPTED=$([ -f ~/.gstack/.proactive-prompted ] && echo "yes" || echo "no")
+_BRANCH=$(git branch --show-current 2>/dev/null || echo "unknown")
+echo "BRANCH: $_BRANCH"
+_SKILL_PREFIX=$($GSTACK_BIN/gstack-config get skill_prefix 2>/dev/null || echo "false")
+_PFX=$([ "$_SKILL_PREFIX" = "true" ] && echo "gstack-" || echo "")
+echo "PROACTIVE: $_PROACTIVE"
+echo "PROACTIVE_PROMPTED: $_PROACTIVE_PROMPTED"
+echo "SKILL_PREFIX: $_SKILL_PREFIX"
+echo "SKILL_PREFIX_VALUE: $_PFX"
+source <($GSTACK_BIN/gstack-repo-mode 2>/dev/null) || true
+REPO_MODE=${REPO_MODE:-unknown}
+echo "REPO_MODE: $REPO_MODE"
+_LAKE_SEEN=$([ -f ~/.gstack/.completeness-intro-seen ] && echo "yes" || echo "no")
+echo "LAKE_INTRO: $_LAKE_SEEN"
+_TEL=$($GSTACK_BIN/gstack-config get telemetry 2>/dev/null || true)
+_TEL_PROMPTED=$([ -f ~/.gstack/.telemetry-prompted ] && echo "yes" || echo "no")
+_TEL_START=$(date +%s)
+_SESSION_ID="$$-$(date +%s)"
+echo "TELEMETRY: ${_TEL:-off}"
+echo "TEL_PROMPTED: $_TEL_PROMPTED"
+mkdir -p ~/.gstack/analytics
+if [ "${_TEL:-off}" != "off" ]; then
+  echo '{"skill":"design-canonical","ts":"'$(date -u +%Y-%m-%dT%H:%M:%SZ)'","repo":"'$(basename "$(git rev-parse --show-toplevel 2>/dev/null)" 2>/dev/null || echo "unknown")'"}'  >> ~/.gstack/analytics/skill-usage.jsonl 2>/dev/null || true
+fi
+# zsh-compatible: use find instead of glob to avoid NOMATCH error
+for _PF in $(find ~/.gstack/analytics -maxdepth 1 -name '.pending-*' 2>/dev/null); do
+  if [ -f "$_PF" ]; then
+    if [ "${_TEL:-off}" != "off" ] && [ -x "$GSTACK_BIN/gstack-telemetry-log" ]; then
+      $GSTACK_BIN/gstack-telemetry-log --event-type skill_run --skill _pending_finalize --outcome unknown --session-id "$_SESSION_ID" 2>/dev/null || true
+    fi
+    rm -f "$_PF" 2>/dev/null || true
+  fi
+  break
+done
+# Learnings count
+eval "$($GSTACK_BIN/gstack-slug 2>/dev/null)" 2>/dev/null || true
+_LEARN_FILE="${GSTACK_HOME:-$HOME/.gstack}/projects/${SLUG:-unknown}/learnings.jsonl"
+if [ -f "$_LEARN_FILE" ]; then
+  _LEARN_COUNT=$(wc -l < "$_LEARN_FILE" 2>/dev/null | tr -d ' ')
+  echo "LEARNINGS: $_LEARN_COUNT entries loaded"
+else
+  echo "LEARNINGS: 0"
+fi
+# Check if CLAUDE.md has routing rules
+_HAS_ROUTING="no"
+if [ -f CLAUDE.md ] && grep -q "## Skill routing" CLAUDE.md 2>/dev/null; then
+  _HAS_ROUTING="yes"
+fi
+_ROUTING_DECLINED=$($GSTACK_BIN/gstack-config get routing_declined 2>/dev/null || echo "false")
+echo "HAS_ROUTING: $_HAS_ROUTING"
+echo "ROUTING_DECLINED: $_ROUTING_DECLINED"
+```
+
+If `PROACTIVE` is `"false"`, do not proactively suggest gstack skills AND do not
+auto-invoke skills based on conversation context. Only run skills the user explicitly
+types (e.g., /qa, /ship). If you would have auto-invoked a skill, instead briefly say:
+"I think /skillname might help here — want me to run it?" and wait for confirmation.
+The user opted out of proactive behavior.
+
+If `SKILL_PREFIX` is `"true"`, the user has namespaced skill names. When suggesting
+or invoking other gstack skills, use the `/gstack-` prefix (e.g., `/gstack-qa` instead
+of `/qa`, `/gstack-ship` instead of `/ship`). Disk paths are unaffected — always use
+`$GSTACK_ROOT/[skill-name]/SKILL.md` for reading skill files.
+
+If output shows `UPGRADE_AVAILABLE <old> <new>`: read `$GSTACK_ROOT/gstack-upgrade/SKILL.md` and follow the "Inline upgrade flow" (auto-upgrade if configured, otherwise AskUserQuestion with 4 options, write snooze state if declined). If `JUST_UPGRADED <from> <to>`: tell user "Running gstack v{to} (just updated!)" and continue.
+
+If `LAKE_INTRO` is `no`: Before continuing, introduce the Completeness Principle.
+Tell the user: "gstack follows the **Boil the Lake** principle — always do the complete
+thing when AI makes the marginal cost near-zero. Read more: https://garryslist.org/posts/boil-the-ocean"
+Then offer to open the essay in their default browser:
+
+```bash
+open https://garryslist.org/posts/boil-the-ocean
+touch ~/.gstack/.completeness-intro-seen
+```
+
+Only run `open` if the user says yes. Always run `touch` to mark as seen. This only happens once.
+
+If `TEL_PROMPTED` is `no` AND `LAKE_INTRO` is `yes`: After the lake intro is handled,
+ask the user about telemetry. Use AskUserQuestion:
+
+> Help gstack get better! Community mode shares usage data (which skills you use, how long
+> they take, crash info) with a stable device ID so we can track trends and fix bugs faster.
+> No code, file paths, or repo names are ever sent.
+> Change anytime with `gstack-config set telemetry off`.
+
+Options:
+- A) Help gstack get better! (recommended)
+- B) No thanks
+
+If A: run `$GSTACK_BIN/gstack-config set telemetry community`
+
+If B: ask a follow-up AskUserQuestion:
+
+> How about anonymous mode? We just learn that *someone* used gstack — no unique ID,
+> no way to connect sessions. Just a counter that helps us know if anyone's out there.
+
+Options:
+- A) Sure, anonymous is fine
+- B) No thanks, fully off
+
+If B→A: run `$GSTACK_BIN/gstack-config set telemetry anonymous`
+If B→B: run `$GSTACK_BIN/gstack-config set telemetry off`
+
+Always run:
+```bash
+touch ~/.gstack/.telemetry-prompted
+```
+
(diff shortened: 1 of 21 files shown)


itstimwhite:
### Merge activity

* **Jul 4, 10:41 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 10:52 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 10:59 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 11:01 AM UTC**: `github-actions` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie).

itstimwhite:
We are given a large diff (153 KB) but we are to focus on the most impactful issues: bugs, security, architectural concerns.
 We are to avoid nitpicking style and focus on correctness.

 Let's break down the changes:

 The PR adds a new canonical design skill (`design-canonical`) and updates several legacy design skills to load the canonical contract first.

 We are to look for:
  1. Logic errors
  2. Security issues
  3. Data integrity issues
  4. Error handling issues
  5. Cross-file impact

 However, note that the changes are primarily in the `.agents/skills/gstack` directory and related files (like `.claude/skills` and the PR template).

 Since the changes are about adding a new skill and updating existing ones, we should look for:

 - In the new skill files (like SKILL.md, SKILL.md.tmpl, and the generated ones in various places):
      * Any shell script in the preamble that might have issues (e.g., undefined variables, command injection, etc.)
      * Any potential for infinite loops or race conditions in the shell scripts.

 - In the updated legacy design skills (like design-review, etc.):
      * We are told they now load the canonical contract first. We should check if the way they load it is safe and correct.

 - In the test file: `.agents/skills/gstack/test/gen-skill-docs.test.ts`
      * We should look for any test that might be missing or incorrect.

 - In the PR template: `.github/PULL_REQUEST_TEMPLATE.md`
      * We are told it updates the PR template to require design-read, dials, checklist, and evidence for design-related changes.
      * We should check if the update is correct and doesn't break anything.

 Given the sheer size of the diff, we must focus on the most critical parts.

 Let's start by examining the preamble of the new skill (which is repeated in multiple places) for shell script issues.

 The preamble is a bash script. We should look for:

   - Unquoted variables that might cause word splitting or globbing issues.
   - Use of `eval` that might be dangerous.
   - Command substitution that might fail silently.
   - Potential for infinite loops (e.g., in the cleanup of old sessions).

 However, note that the preamble is long and complex. We must be cautious.

 Let's look at a few examples from the preamble in the new skill file (`.agents/skills/gstack/design-canonical/SKILL.md`):

 Example 1:
   ```bash
   _ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
   GSTACK_ROOT="$HOME/.codex/skills/gstack"
   [ -n "$_ROOT" ] && [ -d "$_ROOT/.agents/skills/gstack" ] && GSTACK_ROOT="$_ROOT/.agents/skills/gstack"
   ```

   This looks safe.

 Example 2:
   ```bash
   _UPD=$($GSTACK_BIN/gstack-update-check 2>/dev/null || .agents/skills/gstack/bin/gstack-update-check 2>/dev/null || true)
   [ -n "$_UPD" ] && echo "$_UPD" || true
   ```

   This is safe because we are capturing the output and then echoing it if non-empty.

 Example 3:
   ```bash
   mkdir -p ~/.gstack/sessions
   touch ~/.gstack/sessions/"$PPID"
   _SESSIONS=$(find ~/.gstack/sessions -mmin -120 -type f 2>/dev/null | wc -l | tr -d ' ')
   find ~/.gstack/sessions -mmin +120 -type f -exec rm {} + 2>/dev/null || true
   ```

   This is safe: we are creating a session file, then counting recent sessions and removing old ones.

 Example 4: The telemetry part
   ```bash
   _TEL=$($GSTACK_BIN/gstack-config get telemetry 2>/dev/null || true)
   _TEL_PROMPTED=$([ -f ~/.gstack/.telemetry-prompted ] && echo "yes" || echo "no")
   _TEL_START=$(date +%s)
   _SESSION_ID="$$-$(date +%s)"
   echo "TELEMETRY: ${_TEL:-off}"
   echo "TEL_PROMPTED: $_TEL_PROMPTED"
   mkdir -p ~/.gstack/analytics
   if [ "${_TEL:-off}" != "off" ]; then
     echo '{"skill":"design-canonical","ts":"'$(date -u +%Y-%m-%dT%H:%M:%SZ)'","repo":"'$(basename "$(git rev-parse --show-toplevel 2>/dev/null)" 2>/dev/null || echo "unknown")'"}'  >> ~/.gstack/analytics/skill-usage.jsonl 2>/dev/null || true
   fi
   ```

   This is safe: we are using `|| true` to ignore errors, and the JSON is built with single quotes around the variable parts to avoid injection? 
   Actually, note: the `date` command is inside a subshell and then we are using the output in a string that is then echoed. 
   However, the `date` output is controlled and safe.

   But note: the `basename` command: `basename "$(git rev-parse --show-toplevel 2>/dev/null)"` 
   If the git command fails, we use "unknown". This is safe.

 Example 5: The pending telemetry cleanup
   ```bash
   # zsh-compatible: use find instead of glob to avoid NOMATCH error
   for _PF in $(find ~/.gstack/analytics -maxdepth 1 -name '.pending-*' 2>/dev/null); do
     if [ -f "$_PF" ]; then
       if [ "${_TEL:-off}" != "off" ] && [ -x "$GSTACK_BIN/gstack-telemetry-log" ]; then
         $GSTACK_BIN/gstack-telemetry-log --event-type skill_run --skill _pending_finalize --outcome unknown --session-id "$_SESSION_ID" 2>/dev/null || true
       fi
       rm -f "$_PF" 2>/dev/null || true
     fi
     break
   done
   ```

   This loop is intended to run only once because of the `break` at the end. However, note that the `find` command might return multiple files.
   But we break after the first iteration. So we only process the first pending file? 
   This might be a bug: we are breaking after the first file, leaving others unprocessed.

   However, note the comment: "zsh-compatible: use find instead of glob to avoid NOMATCH error". 
   The original intention might have been to process all pending files, but the `break` causes only one to be processed.

   Let's look at the context: we are cleaning up pending telemetry files. We want to process each pending file and then remove it.

   The current code:
      for _PF in ...; do
          ... process one file ...
          break   # This breaks after the first file
      done

   This is likely a mistake. We should remove the `break` to process all files.

   However, note that the `_PF` variable is set to each file found by `find`. The `break` is inside the loop, so it breaks after the first iteration.

   This means that if there are multiple pending files, only the first one is processed and the rest are left.

   This could lead to telemetry events not being logged for those pending files.

   But note: the pending files are named `.pending-*` and are created by the telemetry system. 
   We are trying to log them as a finalize event and then remove them.

   If we leave them, then the next time the skill runs, we will try to process them again (but again only the first one) and so on.

   This is a logic error: we are not cleaning up all pending files.

   However, note that the `break` is inside the `if [ -f "$_PF" ]` block. So if the file exists, we break. 
   But if the file doesn't exist (which shouldn't happen because we found it with `find`), we don't break and go to the next.

   Actually, the `if [ -f "$_PF" ]` is redundant because we found it with `find` and we are only looking for files.

   So the loop will break on the first file it finds, regardless of whether we process it or not.

   This is definitely a bug.

   Let's see if this same pattern appears in other places? The preamble is copied to multiple files.

   We see the same preamble in:
      .agents/skills/gstack/.agents/skills/gstack-design-canonical/SKILL.md
      .agents/skills/gstack/.factory/skills/gstack-design-canonical/SKILL.md
      .claude/skills/design-canonical/SKILL.md
      .claude/skills/gstack/design-canonical/SKILL.md

   And also in the template? The template is used to generate these.

   Therefore, we have the same bug in multiple places.

   Suggested fix: remove the `break` statement so that we process all pending files.

   However, note that the loop is written as:
        for _PF in $(find ...); do
            ... 
            break   # This is the problem
        done

   We should remove the `break`.

   But wait: the loop body has a `break` at the very end, so it only runs once. We want to remove that break.

   Alternatively, note that the loop is intended to run only once? The comment says "zsh-compatible: use find instead of glob to avoid NOMATCH error".
   Maybe the original author intended to use a glob but switched to find for zsh compatibility and then mistakenly left the break.

   Without the break, we would process every pending file.

   Let's change it to:

        for _PF in $(find ~/.gstack/analytics -maxdepth 1 -name '.pending-*' 2>/dev/null); do
            if [ -f "$_PF" ]; then
                if [ "${_TEL:-off}" != "off" ] && [ -x "$GSTACK_BIN/gstack-telemetry-log" ]; then
                    $GSTACK_BIN/gstack-telemetry-log --event-type skill_run --skill _pending_finalize --outcome unknown --session-id "$_SESSION_ID" 2>/dev/null || true
                fi
                rm -f "$_PF" 2>/dev/null || true
            fi
        done

   This will process every pending file.

   However, note that the `_SESSION_ID` is set at the beginning of the preamble and is the same for all pending files in this run.
   That might be acceptable because they are all from the same session? But note: the pending files might be from different sessions.

   Actually, the pending files are created when a skill run is interrupted? We are logging them as `_pending_finalize` with outcome unknown.

   It might be better to extract the session ID from the pending file name? But the current code doesn't do that.

   Given the complexity and the fact that this is a telemetry cleanup, and that we are not required to fix telemetry (it's best effort),
   we might leave it as is? But the bug is that we are not cleaning up all pending files.

   Since the telemetry is not critical for the function of the skill, and the issue is only about cleaning up old pending files,
   we can consider it a low priority. However, the instructions say to focus on bugs.

   We'll flag it as a logic error.

   But note: the same pattern appears in the preamble of the new skill and in the copied ones.

   Let's check the line numbers in the diff for the new skill file:

   The diff for `.agents/skills/gstack/design-canonical/SKILL.md` shows the entire file (since it's new).

   We are looking for the loop:

        # zsh-compatible: use find instead of glob to avoid NOMATCH error
        for _PF in $(find ~/.gstack/analytics -maxdepth 1 -name '.pending-*' 2>/dev/null); do
          if [ -f "$_PF" ]; then
            if [ "${_TEL:-off}" != "off" ] && [ -x "$GSTACK_BIN/gstack-telemetry-log" ]; then
              $GSTACK_BIN/gstack-telemetry-log --event-type skill_run --skill _pending_finalize --outcome unknown --session-id "$_SESSION_ID" 2>/dev/null || true
            fi
            rm -f "$_PF" 2>/dev/null || true
          fi
          break
        done

   The `break` is at the end of the loop body, so it's inside the loop.

   We can see it in the diff? Actually, the diff is the entire file because it's new.

   We don't have the exact line number, but we can approximate.

   Alternatively, we can note that the issue is present in the preamble of the new skill and in the template.

   Since the template is used to generate the other files, we should fix the template.

   The template is: `.agents/skills/gstack/design-canonical/SKILL.md.tmpl`

   We should look for the same pattern in the template.

   However, note that the diff for the template is also provided (it's new).

   Let's look at the diff for `.agents/skills/gstack/design-canonical/SKILL.md.tmpl`:

   We see the same preamble.

   Therefore, we have:

      File: .agents/skills/gstack/design-canonical/SKILL.md.tmpl
      The loop with the break.

   We must fix the template so that when the skill is generated, the break is removed.

   Additionally, note that the same preamble appears in the generated files in other locations (like .claude, .agents, etc.) but they are generated from the template.
   So fixing the template will fix them when regenerated.

   However, note that the PR also adds the generated files (like the ones in .claude and .agents) as new files. 
   We are adding them in this PR. So we must fix them in this PR as well? Or we can fix the template and then regenerate?

   The PR description says: "Updates legacy design skills ... to load the canonical contract first" and adds the canonical skill.

   It also says: "Adds Claude, Codex, and Factory generated host outputs plus a tracked clean-checkout regression test"

   So the generated files are being added explicitly. We have two options:

      Option 1: Fix the template and then regenerate all the generated files (including the ones in .claude, .agents, etc.) and update them in the PR.
      Option 2: Fix the template and also fix the generated files that are being added in this PR.

   Since the PR is already adding the generated files, we should fix them in this PR as well to avoid having the bug in the added files.

   However, note that the PR might be large and we are only allowed to fix the most impactful issues.

   Given that this is a telemetry cleanup bug and telemetry is best effort, we might consider it low impact.

   But let's see if there are other issues.

   Another potential issue: the use of `eval` in the preamble.

   We see:

        # Learnings count
        eval "$($GSTACK_BIN/gstack-slug 2>/dev/null)" 2>/dev/null || true

   This is dangerous because if `gstack-slug` outputs something that is not safe to eval, we could run arbitrary code.

   However, note that `gstack-slug` is a gstack binary. We must trust that it outputs only safe variable assignments.

   But what if it is compromised? This is a supply chain risk.

   However, the same pattern is used in the existing codebase? We are not changing that.

   Since we are not changing the `gstack-slug` tool, and we are only adding a new skill that uses it in the same way as existing skills,
   we might not be introducing a new risk. But note: we are adding the skill, so we are introducing the use of `eval` in this skill.

   However, the existing skills in the gstack codebase might already use this pattern? We are not sure.

   Given that we are not allowed to change the gstack tooling in this PR (we are only adding a skill and updating others),
   and that the existing codebase already uses this pattern (as we can see in the preamble of the new skill, which is copied from existing skills?),
   we might not be able to fix it without changing the tool.

   But note: the PR is about adding a canonical design skill, and we are copying the preamble from existing skills? 
   Actually, the preamble looks complex and specific. It might be that this preamble is new and not copied from elsewhere.

   Without more context, we must assume that the `eval` is intentional and that the `gstack-slug` tool is trusted.

   Therefore, we will not flag the `eval` as a security issue in this context.

   However, note that the `eval` is run in the preamble of every skill run. If an attacker can compromise the `gstack-slug` tool, they can run arbitrary code.
   But that is a separate issue.

   Given the instructions to focus on the changes in this PR, and that we are not changing the `gstack-slug` tool, we leave it.

   Let's look for other issues.

   In the updated legacy design skills (like design-review, etc.), we are told they now load the canonical contract first.

   We should check how they load it.

   For example, in `.agents/skills/gstack/design-review/SKILL.md`:

   We see in the diff:

        diff --git a/.agents/skills/gstack/design-review/SKILL.md b/.agents/skills/gstack/design-review/SKILL.md
        index ... 
        --- a/.agents/skills/gstack/design-review/SKILL.md
        +++ b/.agents/skills/gstack/design-review/SKILL.md
        @@ -1,3 +1,55 @@
        +---
        +name: design-review
        +description: |
        +  Live product audit, fix loop, before/after verification.
        +  Inherits the canonical design operating system contract. Use this skill for
        +  live product audit, fix loop, before/after verification.
        +---
        +<!-- AUTO-GENERATED from SK

itstimwhite (commented):

PR #13140, opened 2026-07-04
feat(inbox): subtle usage pill + limits popover (top-right of home)

### LINKED ISSUE #11566: Opportunity Inbox: subtle usage pill + limits popover (top-right of home)
## Problem

Tim wants usage visible at a glance without a page navigation — a Codex-style 'usage remaining' indicator that's a frequent-glance surface AND an upgrade entry point. Designed as part of the Home Inbox comp.

## Scope (per Tim's design)

* Subtle top-right pill showing usage % (quiet, not loud chrome).
* Click → popover: 'Jovie Usage · <plan>', 'Suggestions · this week <%> · Resets <day>', 'Live actions · 5-hr window <n>/<cap> · Resets in <time>', 'Learn more about limits' (→ full page, retained).
* Wire to real entitlement/usage data; no layout shift as values update.

## MRR score

reach=high (every session), freq=very high (glanceable), annoyance=low, WTP=med (upgrade nudge), ATP=med, effort=low → strong cheap upgrade-surface win.

## Shape & model

Small-well-spec'd. Codex. Reference comp: Home Inbox design (mobile-usage.png) on JovieInc/Jovie#11510/JovieInc/Jovie#11506.

## Acceptance

Pill renders usage % top-right; popover shows weekly + 5-hr limits + resets + Learn more; values from real data; no layout shift.

itstimwhite:
## Summary

Subtle top-right usage pill + limits popover on the Opportunity Inbox home — a glanceable usage surface and upgrade entry point per Tim's Home Inbox comp.

Closes #11566

## What's here

- **`HeaderUsagePill`** — quiet pill in the shell header (home route only) showing the overall remaining percent (the tighter of the two quotas). Click opens a popover:
  - `Jovie Usage · <plan>`
  - `Suggestions — This week <%> left · Resets <day>`
  - `Live actions — 5-hr window <n>/<cap> · Resets in <time>`
  - `Learn more about limits` → `/app/settings/usage` (full page retained)
- **`GET /api/usage/summary`** — real data: counts from `suggested_actions` (suggestions created this UTC week; actions executed in the trailing 5-hr window) + plan from `getCurrentUserEntitlements()`. Fails soft: any DB/schema error degrades to a zero-count `_stale` snapshot instead of throwing (migration-drift-safe).
- **`lib/usage/limits.ts`** — per-plan display caps (single tunable place; doc-noted to migrate into the entitlement registry once enforcement lands — kept out of the registry now to avoid churning the exact-equality entitlements-matrix contract tests), week-window/percent helpers, shared payload type, client formatters.
- **`useUsageSummaryQuery`** — TanStack hook mirroring `useChatUsageQuery` (FREQUENT_BACKGROUND_CACHE, non-retryable FetchError handling), new `usage.summary` query key.

## Acceptance

- [x] Pill renders usage % top-right of home
- [x] Popover shows weekly + 5-hr limits + resets + Learn more
- [x] Values from real data (entitlements plan + `suggested_actions` counts)
- [x] No layout shift — loading state reserves the pill footprint; `tabular-nums` + fixed `min-w` so value updates never reflow

## Notes for review

- Display caps (free 15/wk · 5 per 5h, pro/trial 75 · 25, max 200 · 100) are initial values in one place — flag if you want different numbers.
- Pill is gated to the home route (`/app`) per the comp; widening it shell-wide is a one-line change.
- Existing `HeaderChatUsageIndicator` (chat-route message quota) is untouched.

## Tests

- `tests/unit/lib/usage-limits.test.ts` — week window (Mon UTC start, Sunday edge, Monday rollover), percent clamping, cap monotonicity across plans, reset-day/duration formatters.
- `tests/unit/dashboard/HeaderUsagePill.test.tsx` — pill renders %, loading placeholder reserves footprint, popover content rows + learn-more link, hidden on demo routes.

## Cost Impact

- **External API calls**: none. Two indexed DB counts per request on `/api/usage/summary` (uses `suggested_actions_user_status_created_idx`), client-cached via FREQUENT_BACKGROUND_CACHE + `Cache-Control: private, max-age=30`.
- **Scaling factor**: O(1) queries per pill load; no cron, no polling.

### DIFF
--- a/apps/web/app/api/usage/summary/route.ts
+++ b/apps/web/app/api/usage/summary/route.ts
@@ -1,16 +1,10 @@
 import { and, count, eq, gte, isNotNull, min } from 'drizzle-orm';
 import { NextResponse } from 'next/server';
-import { getCachedAuth } from '@/lib/auth/cached';
 import { isMissingConnectorSchemaError } from '@/lib/connectors/schema-errors';
 import { db } from '@/lib/db';
 import { getUserByClerkId } from '@/lib/db/queries/shared';
 import { suggestedActions } from '@/lib/db/schema/connectors';
-import {
-  getPlanDisplayName,
-  type PlanId,
-  resolveCanonicalPlanId,
-} from '@/lib/entitlements/registry';
-import { getCurrentUserEntitlements } from '@/lib/entitlements/server';
+import { getPlanDisplayName, type PlanId } from '@/lib/entitlements/registry';
 import {
   getRemainingPercent,
   getUsageLimits,
@@ -18,6 +12,7 @@ import {
   LIVE_ACTION_WINDOW_MS,
   type UsageSummaryData,
 } from '@/lib/usage/limits';
+import { requireUsageAuth } from '@/lib/usage/require-usage-auth';
 import { logger } from '@/lib/utils/logger';
 
 export const runtime = 'nodejs';
@@ -159,28 +154,11 @@ function buildUsageSummary(params: {
 }
 
 export async function GET() {
-  let userId: string | null;
-  try {
-    ({ userId } = await getCachedAuth());
-  } catch (error) {
-    // Clerk throws when middleware didn't run (e.g., matcher misconfiguration).
-    // Return 401 for that case, but let unexpected errors propagate to Sentry.
-    const message = error instanceof Error ? error.message : '';
-    if (message.includes('clerkMiddleware')) {
-      return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
-    }
-    throw error;
+  const auth = await requireUsageAuth();
+  if (auth.response) {
+    return auth.response;
   }
-  if (!userId) {
-    return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
-  }
-
-  const entitlements = await getCurrentUserEntitlements();
-  if (!entitlements.isAuthenticated || !entitlements.userId) {
-    return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
-  }
-
-  const plan = resolveCanonicalPlanId(entitlements.plan) ?? 'free';
+  const { userId, plan } = auth;
 
   const now = new Date();
   const week = getWeeklyUsageWindow(now);
--- a/apps/web/app/api/usage/summary/route.ts
+++ b/apps/web/app/api/usage/summary/route.ts
@@ -1,0 +1,198 @@
+import { and, count, eq, gte, isNotNull, min } from 'drizzle-orm';
+import { NextResponse } from 'next/server';
+import { getCachedAuth } from '@/lib/auth/cached';
+import { isMissingConnectorSchemaError } from '@/lib/connectors/schema-errors';
+import { db } from '@/lib/db';
+import { getUserByClerkId } from '@/lib/db/queries/shared';
+import { suggestedActions } from '@/lib/db/schema/connectors';
+import {
+  getPlanDisplayName,
+  type PlanId,
+  resolveCanonicalPlanId,
+} from '@/lib/entitlements/registry';
+import { getCurrentUserEntitlements } from '@/lib/entitlements/server';
+import {
+  getRemainingPercent,
+  getUsageLimits,
+  getWeeklyUsageWindow,
+  LIVE_ACTION_WINDOW_MS,
+  type UsageSummaryData,
+} from '@/lib/usage/limits';
+import { logger } from '@/lib/utils/logger';
+
+export const runtime = 'nodejs';
+
+const CACHE_HEADERS = {
+  'Cache-Control': 'private, max-age=30, stale-while-revalidate=120',
+} as const;
+
+interface UsageCounts {
+  readonly suggestionsThisWeek: number;
+  readonly liveActionsInWindow: number;
+  readonly oldestLiveActionAt: Date | null;
+  readonly stale: boolean;
+}
+
+const EMPTY_COUNTS: UsageCounts = {
+  suggestionsThisWeek: 0,
+  liveActionsInWindow: 0,
+  oldestLiveActionAt: null,
+  stale: false,
+};
+
+/**
+ * Real usage counts from `suggested_actions`:
+ * - suggestions surfaced this calendar week (createdAt >= UTC Monday)
+ * - live actions executed in the trailing 5-hour window (executedAt)
+ *
+ * Fails soft: any DB error (including missing connector schema) degrades to
+ * zero counts flagged `_stale` instead of throwing — the pill still renders.
+ */
+async function loadUsageCounts(
+  clerkUserId: string,
+  weekStart: Date,
+  windowStart: Date
+): Promise<UsageCounts> {
+  try {
+    const dbUser = await getUserByClerkId(db, clerkUserId);
+    if (!dbUser) {
+      return EMPTY_COUNTS;
+    }
+
+    const [suggestionRows, liveActionRows] = await Promise.all([
+      db
+        .select({ value: count() })
+        .from(suggestedActions)
+        .where(
+          and(
+            eq(suggestedActions.userId, dbUser.id),
+            gte(suggestedActions.createdAt, weekStart)
+          )
+        ),
+      db
+        .select({
+          value: count(),
+          oldest: min(suggestedActions.executedAt),
+        })
+        .from(suggestedActions)
+        .where(
+          and(
+            eq(suggestedActions.userId, dbUser.id),
+            isNotNull(suggestedActions.executedAt),
+            gte(suggestedActions.executedAt, windowStart)
+          )
+        ),
+    ]);
+
+    const oldestRaw = liveActionRows[0]?.oldest ?? null;
+
+    return {
+      suggestionsThisWeek: suggestionRows[0]?.value ?? 0,
+      liveActionsInWindow: liveActionRows[0]?.value ?? 0,
+      oldestLiveActionAt: oldestRaw ? new Date(oldestRaw) : null,
+      stale: false,
+    };
+  } catch (error) {
+    if (!isMissingConnectorSchemaError(error)) {
+      logger.warn('Usage summary counts unavailable; serving degraded data', {
+        error,
+      });
+    }
+    return { ...EMPTY_COUNTS, stale: true };
+  }
+}
+
+function buildUsageSummary(params: {
+  readonly plan: PlanId;
+  readonly counts: UsageCounts;
+  readonly weekResetAt: Date;
+}): UsageSummaryData {
+  const { plan, counts, weekResetAt } = params;
+  const limits = getUsageLimits(plan);
+
+  const suggestionsRemaining = Math.max(
+    0,
+    limits.suggestionsPerWeek - counts.suggestionsThisWeek
+  );
+  const liveActionsRemaining = Math.max(
+    0,
+    limits.liveActionsPer5h - counts.liveActionsInWindow
+  );
+
+  const suggestionsRemainingPercent = getRemainingPercent(
+    counts.suggestionsThisWeek,
+    limits.suggestionsPerWeek
+  );
+  const liveActionsRemainingPercent = getRemainingPercent(
+    counts.liveActionsInWindow,
+    limits.liveActionsPer5h
+  );
+
+  const liveActionsResetAt = counts.oldestLiveActionAt
+    ? new Date(
+        counts.oldestLiveActionAt.getTime() + LIVE_ACTION_WINDOW_MS
+      ).toISOString()
+    : null;
+
+  return {
+    plan,
+    planDisplayName: getPlanDisplayName(plan),
+    suggestions: {
+      used: counts.suggestionsThisWeek,
+      limit: limits.suggestionsPerWeek,
+      remaining: suggestionsRemaining,
+      remainingPercent: suggestionsRemainingPercent,
+      resetAt: weekResetAt.toISOString(),
+    },
+    liveActions: {
+      used: counts.liveActionsInWindow,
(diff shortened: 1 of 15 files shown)


itstimwhite:
We are given a diff for PR #13140. We need to check for logic errors, security issues, data integrity, error handling, and cross-file impact.

Let's go through each changed file and note any potential issues.

1. `apps/web/app/api/usage/summary/route.ts` (new file)
   - This is an API route that returns usage summary.
   - We note:
     - It uses `getCachedAuth()` to get the user ID. If there's an error in `getCachedAuth()`, it catches and checks for a specific error message ('clerkMiddleware') to return 401, otherwise rethrows.
     - It then calls `getCurrentUserEntitlements()` and checks if the user is authenticated and has a userId. If not, returns 401.
     - It resolves the plan ID and then computes the weekly usage window and the 5-hour window start.
     - It calls `loadUsageCounts` to get the counts from the database.
     - In `loadUsageCounts`:
        - It tries to get the db user by clerkId. If not found, returns empty counts.
        - Then it runs two queries: one for suggestions this week and one for live actions in the window.
        - If there's an error, it checks if it's a missing connector schema error. If not, it logs a warning and returns empty counts with `stale: true`.
        - Note: The error handling for missing connector schema is to treat it as not an error (so we return empty counts without logging a warning? Actually, we do log a warning for non-missing-connector errors, but for missing connector schema we just return empty counts without logging? Let's see: 
          ```ts
          if (!isMissingConnectorSchemaError(error)) {
            logger.warn('Usage summary counts unavailable; serving degraded data', {
              error,
            });
          }
          return { ...EMPTY_COUNTS, stale: true };
          ```
          So for missing connector schema, we don't log and just return stale counts. That seems acceptable because it's expected during migrations.
     - The function `buildUsageSummary` then builds the summary from the plan and counts.
     - The API returns the summary with cache headers.

   Potential issues:
     - In `loadUsageCounts`, we are using `min(suggestedActions.executedAt)` to get the oldest executed action in the window. However, note that the `executedAt` might be null for some rows? We are filtering with `isNotNull(suggestedActions.executedAt)` so that's safe.
     - The `oldestLiveActionAt` is set to `null` if there are no rows. Then in `buildUsageSummary`, we compute `liveActionsResetAt` as:
          ```ts
          const liveActionsResetAt = counts.oldestLiveActionAt
            ? new Date(
                counts.oldestLiveActionAt.getTime() + LIVE_ACTION_WINDOW_MS
              ).toISOString()
            : null;
          ```
        This is correct: if there are no live actions in the window, then there is no reset time (so we show "Window clear" in the popover).
     - However, note that the `liveActionsResetAt` in the summary is used in the popover to show "Resets in <time>". If there are no live actions, we show "Window clear". That's acceptable.

   - The API route does not have any obvious logic errors, but note:
        - The `getCurrentUserEntitlements()` function is called without any parameters. We assume it uses the current request context (via Clerk) to get the entitlements. This is acceptable because we are in a route handler and the middleware (or cached auth) has set the user.

   - Security: We are using parameterized queries (via Drizzle) so no SQL injection.
   - Error handling: We catch errors in the auth step and in the database query step. We return 401 for auth issues and degrade gracefully for DB errors.

   - Data integrity: The counts are computed correctly? We are counting suggestions created in the week (UTC Monday start) and live actions executed in the trailing 5-hour window. This matches the description.

   - Cross-file impact: This API is used by the new hook `useUsageSummaryQuery`.

2. `apps/web/components/features/dashboard/atoms/HeaderUsagePill.tsx` (new file)
   - This is a client component that shows the usage pill and popover.
   - We note:
        - It uses `usePathname` to check if we are on a demo route or the home route (only shows on `/app` and not on demo routes).
        - It uses `env.IS_E2E` to disable in E2E tests (passive runtime).
        - It uses the `useUsageSummaryQuery` hook to fetch the data.
        - While loading, it renders an invisible span with `opacity-0` and `tabular-nums` to reserve space.
        - The pill shows the overall remaining percent (the tighter of the two quotas).
        - The popover shows:
            - Plan name
            - Suggestions: "This week · {remainingPercent}% left" and "Resets <day>"
            - Live actions: "5-hr window · {used}/{limit}" and either "Resets in <time>" or "Window clear"
            - A link to the usage settings page.

   Potential issues:
        - The pill is only shown on the home route (`pathname === APP_ROUTES.DASHBOARD`). The description says it's for the Opportunity Inbox home, which is the dashboard route. This matches.
        - The popover content:
            - For live actions, we show `{data.liveActions.used}/{data.liveActions.limit}`. This is correct.
            - The reset time for live actions: we show `Resets in ${formatResetIn(data.liveActions.resetAt)}` if `resetAt` is not null, else 'Window clear'. This is correct.
        - However, note that the `liveActions.resetAt` in the data is either an ISO string or null. The `formatResetIn` function handles null by returning '—' but we are checking for null in the JSX so we never call `formatResetIn` with null. We are safe.

        - The pill uses `aria-label` that includes the remaining percent. This is good for accessibility.

        - The loading state: we render a span with `aria-hidden` and an inner span with `opacity-0` and `tabular-nums` showing "100%". This reserves the space because the inner span has the same width as the actual value (due to `tabular-nums` and fixed `min-w-14`). However, note that the loading state shows "100%" (invisible) and the actual value might be different. But because we are using `tabular-nums` and a fixed min-width, the width should be the same for any two-digit number (and three-digit? 100% is three characters). The `min-w-14` is set to 14 units (which is about 3.5rem). Let's check: 
          - The pill has `min-w-14` and `text-app font-caption`. We assume the font is monospace enough for tabular-nums to have fixed width per digit? Actually, `tabular-nums` makes numbers have the same width, so any two-digit number is the same width, and three-digit numbers are wider. However, 100% is three digits and the percent sign. The loading state shows "100%" (three digits and a percent sign) and the actual value might be, say, "5%" (one digit and a percent sign) or "85%" (two digits and a percent sign). 

        - The problem: the loading state reserves space for three digits (like 100%) but if the actual value is one or two digits, the pill might shrink when the data loads, causing a layout shift.

        - The description says: "Loading state reserves the pill footprint; `tabular-nums` + fixed `min-w` so value updates never reflow". However, note that the `min-w-14` is fixed, but the content inside might change width because the number of digits changes.

        - We should ensure that the loading state and the actual value have the same width. The current loading state shows "100%" (which is 4 characters: '1','0','0','%'). The actual value might be:
            - "5%" -> 2 characters
            "85%" -> 3 characters
            "100%" -> 4 characters

        - So if we show "5%" in a space reserved for "100%", it will be smaller and the pill might shrink? Actually, the pill has `min-w-14` so it won't shrink below 14 units. But the content inside might be centered and if the content is narrower, then the pill will have more padding on the sides? That might cause a shift because the total width of the pill is fixed by `min-w-14` but the content width changes, so the padding adjusts. However, the pill's outer width is fixed by `min-w-14` (and it can grow if needed, but we set a min-width). The inner content is wrapped in a span with `tabular-nums` and we are not setting any width on the inner span. 

        - The solution: we should make sure that the loading state and the actual value have the same number of digits. We can show the loading state as "---%" or something that has the same width as the maximum expected value. The maximum remaining percent is 100, so we need three digits. We can show "100%" in the loading state (which we are doing) and that is three digits. But note: the actual value might be less than three digits. However, because we are using `tabular-nums`, the width of the numbers is fixed per digit? Actually, `tabular-nums` makes all digits the same width, but the width of the string "5" is the same as the width of "0", but the string "5%" is two characters (digit and percent) and "100%" is four characters (three digits and percent). So the width of "5%" is less than "100%".

        - To avoid layout shift, we should reserve space for the maximum possible string. The maximum string for the percent is "100%" (4 characters). We can change the loading state to show a string of 4 characters, for example, "100%" (which we are doing) but note that the actual value might be "5%" (2 characters) and that would be narrower. However, the pill has a fixed min-width of 14 units. We need to check if the width of the pill is determined by the content or by the min-width.

        - The pill has:
          `inline-flex min-w-14 items-center justify-center gap-1.5 ...`
          The `min-w-14` sets the minimum width of the pill to 14 units. The content inside might be narrower, but the pill will not shrink below 14 units. However, if the content is wider than 14 units, the pill will grow. So if we set the loading state to show "100%" (which is 4 characters) and the actual value is "5%" (2 characters), the pill width will be the same because the min-width is 14 units and the content width of "5%" is less than 14 units? We don't know the exact conversion, but we can assume that the designer chose 14 units to be enough for "100%".

        - Alternatively, we can change the loading state to show three digits and a percent sign, and then use `opacity-0` to hide it. The actual value will be shown with `tabular-nums` and without opacity. The width of the inner span (the one with the percent) should be the same for any value because we are using `tabular-nums` and we are showing the same number of characters? But note: we are showing a variable number of digits.

        - Actually, the inner span in the loading state is:
          `<span className='tabular-nums opacity-0'>100%</span>`
          and in the actual value:
          `<span className='tabular-nums'>{data.remainingPercent}%</span>`

        - The `tabular-nums` class makes each digit have the same width, but the percent sign is not a digit and might have a different width. Also, the number of digits varies.

        - To have a fixed width, we should format the number to always be three digits (with leading zeros) but that would be weird (e.g., 005% for 5%). Alternatively, we can use a monospace font for the numbers? But we are already using `tabular-nums` which is designed for this.

        - However, note that the `tabular-nums` feature in OpenType makes numbers have the same width, but it does not make sure the font supports it. We assume the font does.

        - But the percent sign is not a number, so it might break the alignment. We could wrap the number and the percent sign separately? But that complicates.

        - Another idea: show the loading state as three digit groups, for example, "100" (without the percent) and then show the percent sign outside? But then we lose the semantic grouping.

        - Given the complexity and the fact that the designer specified `tabular-nums` and a fixed `min-w`, we assume that the min-width is set to accommodate the worst case (three digits and a percent sign). And because the pill has `justify-center`, the content will be centered and if the content is narrower, the pill will have more space on the sides. However, the pill's width is fixed by `min-w-14` so it won't change. Therefore, there should be no layout shift.

        - But note: the pill's width is set to at least 14 units. If the content is narrower than 14 units, the pill will still be 14 units wide (because of `min-w-14`). If the content is wider, the pill will grow. So if the loading state content is 100% (which we assume is within 14 units) and the actual value is 5% (which is narrower), the pill width remains 14 units (because of the min-width) and the content is centered. So the pill's outer width does not change. Therefore, no layout shift.

        - However, what if the actual value is wider than 14 units? For example, if we ever show more than 100%? We don't because we clamp the percent to 0-100. So the maximum is 100%. Therefore, the loading state and the actual value will both fit within the same width (the width of "100%") and the pill's width is at least that. So we are safe.

   - The popover: 
        - It uses `align='end'` and `w-72 p-0`. This seems acceptable.
        - The content is structured with a title, a description list, and a link.

   - The link to the usage settings page uses `APP_ROUTES.SETTINGS_USAGE`. We assume that constant is defined correctly.

   - The component is wrapped in `memo` to prevent unnecessary re-renders.

   - We note that the component returns `null` on demo routes and in E2E (passive runtime). This is correct.

   - No obvious logic errors, security issues, or data integrity issues.

3. `apps/web/components/organisms/AuthShellWrapper.tsx` (modified)
   - We added import for `HeaderUsagePill`.
   - We added a variable `showUsagePill` that is true when not on a demo route and the pathname is the dashboard route (`APP_ROUTES.DASHBOARD`).
   - We then conditionally render `<HeaderUsagePill />` in the header actions, after the `ArtistProfileRailToggle` and before the `HeaderChatUsageIndicator`.
   - We also added `showUsagePill` to the dependencies of the `useMemo` that returns the default header action.

   - This seems correct. The pill is only shown on the dashboard route (home) and not on demo routes.

   - No issues found.

4. `apps/web/lib/queries/index.ts` (modified)
   - We exported the new query hook and types from `useUsageSummaryQuery`.
   - This is a simple re-export. No issues.

5. `apps/web/lib/queries/keys.ts` (modified)
   - We added a new key for usage summary under `queryKeys.usage`.
   - This is correct.

6. `apps/web/lib/queries/useUsageSummaryQuery.ts` (new file)
   - This is a thin wrapper around `useQuery` for the usage summary endpoint.
   - It uses `FREQUENT_BACKGROUND_CACHE` for the cache strategy (which we assume is defined elsewhere and is appropriate for background data).
   - The `retry` function: 
        ```ts
        retry: (failureCount: number, error: Error) => {
          if (error instanceof FetchError && !error.isRetryable()) {
            return false;
          }
          return failureCount < 1;
        },
        ```
     - This means: if it's a `FetchError` and not retryable (like a 4xx error), then don't retry. Otherwise, retry if failureCount < 1 (i.e., at most one retry).
   - The `retryDelay` uses exponential backoff capped at 10 seconds.

   - This seems reasonable.

   - No issues found.

7. `apps/web/lib/usage/limits.ts` (new file)
   - This file contains:
        - The usage limits per plan (free, trial, pro, max).
        - Helper functions for weekly window, remaining percent, and formatting.

   - We note:
        - The `LIVE_ACTION_WINDOW_MS` is set to 5 hours in milliseconds.
        - The `getWeeklyUsageWindow` function: 
            ```ts
            export function getWeeklyUsageWindow(now = new Date()): WeeklyUsageWindow {
              const daysSinceMonday = (now.getUTCDay() + 6) % 7;
              const start = new Date(
                Date.UTC(
                  now.getUTCFullYear(),
                  now.getUTCMonth(),
                  now.getUTCDate() - daysSinceMonday
                )
              );
              return { start, resetAt: new Date(start.getTime() + 7 * DAY_MS) };
            }
            ```
          This calculates the start of the week (Monday 00:00 UTC) and the resetAt (start of next week).

        - The `getRemainingPercent` function:

itstimwhite (commented):

itstimwhite:
Rebasing for merge queue

itstimwhite:
Zoe overnight merge watch: closing to re-trigger fresh CI (stale PR Ready failure from cancelled previous runs)

Issue #10778, opened 2026-06-12
[Audit] P1: Library status incoherence — three 'Draft' meanings, 53/53 items in Draft approval, ~47 duplicate EP rows, no search


itstimwhite:
## Problem

On /app/library: (1) “Draft” appears with three different meanings in one sidebar — Smart Filter “Drafts 2”, Approval Status “Draft 53”, Release Status “Draft 2”; (2) all 53 items sit in Draft approval (field is dead weight without a review workflow); (3) ~47 rows are near-duplicates of “All This Noise EP” / “All This Noise (Remixed)” (per-track rows or duplicate ingest — AI chat itself reports “many releases show 0 tracks, suggesting possible sync issues” and “Spotify not currently connected” while the filter rail claims “Spotify 49”); (4) an unexplained numeric badge column; (5) no search input.

## Evidence

Observed live 2026-06-11 (owner account). Chat self-report same session.

## Fix

Single-owner status pipeline (Draft → Pending → Live; one state at a time — [Frame.io](<http://Frame.io>) V4 model); version-stack duplicate releases into one row; hide Approval Status until a review workflow exists; rename data-quality smart filters into one “Needs attention” view; add search; label or remove the numeric column; reconcile provider counts with actual connection state.

## Acceptance criteria

One visible status vocabulary; one row per release; search by title; counts consistent with connection truth.

## Shape & model tag

Medium, well-spec'd + data cleanup script. Model: Codex.

## MRR

Library is the canonical catalog surface — incoherence here undermines every downstream feature (releases, pitches, profile).

itstimwhite:
Jovie agent (codex issue shipper) claimed this issue.

Branch: `codex/gh-10778-audit-p1-library-status-incoherence-three-draft`
Risk: `high`
Model route: `escalation` using `grok-composer-2.5-fast`
Integration branch: not used for this issue

itstimwhite:
Jovie agent (codex issue shipper) stopped on a real blocker.

Agent exited without successfully shipping.

Status: `1`
Signal: `null`
Log: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-10778-2026-07-01T20-52-28-747z.log`
Prompt: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-10778-2026-07-01T20-52-28-747z.prompt.md`

PR #13150, opened 2026-07-04
feat(releases): stacked provider logos with +N more on release rows (GH-11493)

### LINKED ISSUE #11493: Releases: replace bare provider-count number with stacked provider logos (+N more); investigate why only ~3 of ~27 providers show
## Problem

Two linked issues on the release row, per owner:

**A. Display (UI polish).** Each release row shows a bare number (e.g. "3") next to Draft/Approved. It's the provider count but it's unlabeled and unclear, and the raw number isn't useful. Replace with **stacked provider logos** (small overlapping icons) + a **"+N more"** affordance.

**B. Data completeness (investigation).** Most songs show only ~3 providers when a fully-distributed release should have ~27. Investigate why provider coverage is so low — ingest/sync gap, mapping gap, or display cap. JovieInc/Jovie#10778 already flags "reconcile provider counts with actual connection state" and the chat self-report "Spotify not currently connected while filter rail claims Spotify 49"; JovieInc/Jovie#5818 (Done) consolidated provider mapping. Confirm whether this is a missing-data problem (providers never ingested) vs a display/mapping problem.

## Severity & Priority

High (P1) for the data-completeness half (a release that only shows 3 of 27 DSPs looks broken on the storefront and weakens SmartLinks); the logo-stack display is a small polish that ships alongside.

## MRR opportunity

Reach: all releases. Annoyance: med-high (looks incomplete; fewer DSP links = weaker SmartLink CTR). Effort: S (display) + M (data investigation). MRR-per-effort: medium-high on the data half.

## Acceptance criteria

* Row shows stacked provider logos + "+N more" instead of a bare number; hover/click reveals the full provider list (ties to the release action-menu Providers submenu).
* Documented root cause for low provider coverage; if data gap, a fix or follow-up to backfill/re-sync providers; provider counts reconcile with actual connection truth.

## Issue shape & model tag

Display = Small (Kimi/DeepSeek). Data investigation = Medium, Codex. Tag Codex overall (investigation drives it).

## Related

Parent JovieInc/Jovie#10383; JovieInc/Jovie#10778 (provider-count vs connection truth), JovieInc/Jovie#5818 (provider mapping consolidation).

itstimwhite:
Fixes #11493

## A. Display — stacked provider logos + "+N more"

- **`DspAvatarStack`** (canonical shell DSP primitive, already used by `ShellReleaseRow`) gains a `maxVisible` prop: it now renders up to N overlapping brand avatars before collapsing the rest into the existing "+N" chip. Default stays `1`, so no other call site changes shape. The existing hover/focus popover still lists every provider with its status.
- **Library table `ProvidersCell`** (the bare number next to Draft/Approved) now renders `DspAvatarStack maxVisible={3}` built from the asset's actual provider links (brand color + label from the canonical `PROVIDER_CONFIG`). Zero-provider assets (e.g. merch) render a muted em dash. Providers column widened 86→120 to fit the stack.
- **`ShellReleaseRow`** now shows 3 stacked avatars (`maxVisible={3}`, cell w-20→w-24).

## B. Data completeness — why releases show ~3 providers (root cause)

Investigated per the issue; it is **both** an ingest gap and a display cap, four compounding layers:

1. **Historical ingest gap (the literal "~3")** — releases imported before the MusicFetch integration landed (2026-02-14) only have Spotify / Apple Music / Deezer links. This is documented in `lib/discography/re-enrich.ts`'s own docstring. The same ~3 result reproduces any time `MUSICFETCH_API_TOKEN` is unset or the MusicFetch circuit breaker is open: only the Apple Music + Deezer ISRC lookups (plus the Spotify link from the import source) persist canonical links.
2. **Backfill throttle** — the daily-maintenance cron runs `sweepUnderEnrichedProfilesForCron` with `MAX_PROFILES_PER_CRON_SWEEP = 1` and 4 releases/profile, rate-limited by MusicFetch (6 req/min, 11s inter-release delay). At that pace a catalog backfill takes months, so pre-2026-02 releases stay at ~3 indefinitely. The admin endpoint `/api/admin/re-enrich` can force a faster sweep (5 profiles × 10 releases per invocation).
3. **Registry ceiling — ~16 is the max, not ~27** — discovery requests `DEFAULT_PROVIDERS = STREAMING_DSP_KEYS`, derived from the 28-key `DSP_REGISTRY` (~16 streaming). But `PROVIDER_CONFIG` (what the smart-link page + view models can display) lists 31 providers; **13 keys can never be ingested** because they're missing from the registry: `netease, qq_music, awa, audius, flo, gaana, jio_saavn, joox, kkbox, line_music, trebel, yandex, amazon`. MusicFetch results for those keys are silently dropped by the `providers.includes(key)` filter in `runIsrcLookups`. This is remaining scope from the #5818 mapping consolidation.
4. **Display caps (fixed here)** — `releaseToDspItems` hardcoded 4 majors, so even a fully-enriched release showed at most 4 avatars in the shell row; the Library showed a bare count. This PR maps **all** of a release's provider links (4 anchored majors as live/missing + every other linked provider as live), so full coverage is visible and the "+N" count is truthful.

Re #10778 (provider counts vs connection truth): the Library/shell counts are per-release *link* counts from `provider_links`, a different axis than profile-level DSP *connection* state — the fix for that reconciliation belongs with #10778, not here.

### Recommended follow-ups (not in this PR)
- Add the 13 `PROVIDER_CONFIG`-only providers to `DSP_REGISTRY` (or intentionally retire them) so ingestion and display agree on the provider universe — closes the ~16 vs ~27 gap at the source.
- Raise the daily backfill budget (e.g. `MAX_PROFILES_PER_CRON_SWEEP` 1→3) or run a one-time `/api/admin/re-enrich` sweep to clear the pre-MusicFetch backlog.

## Tests
- New unit test `release-adapters.test.ts`: majors anchored as missing, live when linked, non-major providers appended (no cap), `PROVIDER_CONFIG` accent fallback for registry-missing keys.
- Existing `ShellReleaseRow`/`LibrarySurface` tests don't assert on the old count pill; no updates needed.

## Cost Impact
None — display-layer change plus one unit test; no new external API calls, crons, or queries.

bug-to-test: adapter behavior covered by new unit test; the Library cell change is presentation-only.


itstimwhite:
### Merge activity

* **Jul 4, 7:44 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 7:57 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 7:58 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 8:01 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 8:03 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 8:09 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 8:27 PM UTC**: This pull request can not be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie). Please try rebasing and resubmitting to merge when ready. 
* **Jul 4, 8:27 PM UTC**: [Graphite](https://app.graphite.com/github/pr/JovieInc/Jovie/13150) disabled "merge when ready" on this PR due to: a merge conflict with the target branch; resolve the conflict and try again..
* **Jul 4, 8:27 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 8:30 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 8:30 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 8:32 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 9:31 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 9:42 PM UTC**: This pull request can not be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie). Please try rebasing and resubmitting to merge when ready. 
* **Jul 4, 9:42 PM UTC**: [Graphite](https://app.graphite.com/github/pr/JovieInc/Jovie/13150) disabled "merge when ready" on this PR due to: a merge conflict with the target branch; resolve the conflict and try again..
* **Jul 4, 9:42 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 9:46 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 9:51 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 9:52 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.

itstimwhite:
We are given a diff for PR #13150. We need to check for logic errors, security, data integrity, error handling, and cross-file impact.

Let's go through each changed file and look for issues.

1. CHANGELOG.md: Only a documentation change. No logic, security, etc.

2. apps/web/app/app/(shell)/library/LibrarySurface.tsx:
   - Added imports for DspAvatarStack and PROVIDER_CONFIG.
   - Added a constant LIBRARY_DSP_FALLBACK_COLOR.
   - Added a function libraryProvidersToDspItems.
   - Changed ProvidersCell to use the new function and DspAvatarStack.
   - Changed the column size for providers from 86 to 120 and minSize from 72 to 96.

   Potential issues:
   - In libraryProvidersToDspItems: 
        const config = PROVIDER_CONFIG[provider.key];
        const label = config?.label ?? provider.label;
        ... color: config?.accent ?? LIBRARY_DSP_FALLBACK_COLOR

     This seems safe because we are using optional chaining and providing a fallback.

   - However, note that the function libraryProvidersToDspItems is called with `asset.providers` which is of type `LibraryReleaseAsset['providers']`. We assume that each provider in that array has a `key` and a `label` (if the config is missing). The type definition for LibraryReleaseAsset is not in the diff, but we trust that it is correct.

   - The change in column size: we are increasing the width. This is a presentation change and should not break anything.

3. apps/web/components/features/dashboard/organisms/release-provider-matrix/shell-releases/ShellReleaseRow.tsx:
   - Changed the div wrapping DspAvatarStack from w-20 to w-24 and added maxVisible={3}.

   This is a presentation change and should be safe.

4. apps/web/components/features/dashboard/organisms/release-provider-matrix/shell-releases/release-adapters.ts:
   - Added a constant FALLBACK_DSP_COLOR.
   - Changed the function releaseToDspItems to:
        a. First, map the SHELL_DSP_META (the four majors) to DspAvatarItem, marking them as live if the release has a link for that provider, else missing.
        b. Then, for the rest of the providers (non-majors) in the release, map them to DspAvatarItem as live, using the DSP_CONFIGS for color and label, falling back to PROVIDER_CONFIG and then FALLBACK_DSP_COLOR.

   Potential issues:
   - We are creating a Set of majorKeys from SHELL_DSP_META. Then we filter the release.providers to exclude those keys for the "rest".
   - We are using:
        const config = DSP_CONFIGS[p.key];
        const label = config?.name ?? p.label;
        color: config?.color ?? PROVIDER_CONFIG[p.key]?.accent ?? FALLBACK_DSP_COLOR

     This is safe because we are using optional chaining and providing fallbacks.

   - However, note that the release.providers array might contain a provider key that is not in DSP_CONFIGS and not in PROVIDER_CONFIG. Then we fall back to FALLBACK_DSP_COLOR and use p.label for the label. The glyph is set to the first character of the label (uppercased). This is acceptable.

   - We are returning [...majors, ...rest]. The majors are always 4 items (the four hardcoded majors). The rest are the non-major providers from the release. There is no deduplication between majors and rest because we filtered out the major keys from the release.providers when building rest. So if a release has a link for a major provider, it is only in the majors part (and marked live) and not in rest. If it doesn't have a link for a major, it is in the majors part as missing and not in rest (because we filtered out the major keys). So no duplication.

   - One potential issue: what if the release.providers array has duplicate keys? The function does not deduplicate. However, we assume that the release.providers array does not have duplicates (as it comes from the database and is a set of links). But to be safe, we might consider deduplication. However, the existing code (before this change) also did not deduplicate and the issue does not mention duplicates. Since this is a display change and the data model is assumed to be correct, we can leave it.

5. apps/web/components/shell/DspAvatarStack.tsx:
   - Added a maxVisible prop (default 1).
   - Changed the logic to show up to maxVisible avatars (overlapping) and then the overflow chip for the rest.

   Potential issues:
   - The variable `visible` is defined as:
        const visible = ordered.slice(0, Math.max(1, maxVisible));
     This ensures that at least one avatar is shown (if maxVisible is 0 or negative, we show 1). However, the prop is defined as a number and we are using Math.max(1, maxVisible). This is safe.

   - The variable `others` is:
        const others = Math.max(0, ordered.length - visible.length);
     This is correct.

   - In the rendering, we map over `visible` and for each dsp we render an avatar. We set the zIndex to `visible.length - index` so that the first avatar (index 0) is on top (highest zIndex) and the last in the visible set is at the bottom. This creates the overlapping effect.

   - We are using the same status dots and glyphs for each avatar in the stack.

   - The overflow chip shows the count of `others` (the number of avatars not visible).

   - The popover (not shown in the diff, but implied by the existing code) still lists every provider. We don't see the popover code in the diff, but we assume it is unchanged and still works.

   - One issue: the original code had a primary avatar style that set a CSS variable. Now we are setting the same variable for each avatar in the stack via the style prop. This is acceptable.

   - However, note that the original code had:
        const primaryAvatarStyle = {
          '--system-b-dsp-avatar-color': primary.color,
        } as CSSProperties;
     and then applied it to the primary avatar span.

     Now we are doing:
        style={{
          '--system-b-dsp-avatar-color': dsp.color,
          zIndex: visible.length - index,
        } as CSSProperties}

     This is correct.

   - But note: the original code also had a ring and other classes. We are now applying the same classes to each avatar in the stack, plus the conditional classes for status and the negative margin for overlapping (index>0 && '-ml-1.5').

   - The negative margin is applied to all but the first avatar (index>0) to make them overlap.

   - This seems correct.

6. apps/web/tests/components/features/dashboard/organisms/release-provider-matrix/shell-releases/release-adapters.test.ts:
   - New test file. We are testing the releaseToDspItems function.

   - The tests look correct and cover the scenarios.

Logic errors:
   - In release-adapters.ts: 
        const majorKeys = new Set<string>(SHELL_DSP_META.map(m => m.providerKey));
     Then we filter release.providers by !majorKeys.has(p.key). This is correct.

   - However, note that SHELL_DSP_META is an array of objects with a property `providerKey`. We are mapping to get the keys. This is safe.

   - In the same file, when building the rest array, we are using:
        const config = DSP_CONFIGS[p.key];
        const label = config?.name ?? p.label;
        color: config?.color ?? PROVIDER_CONFIG[p.key]?.accent ?? FALLBACK_DSP_COLOR

     What if p.key is not in PROVIDER_CONFIG? Then we use FALLBACK_DSP_COLOR. This is by design.

   - In LibrarySurface.tsx, the function libraryProvidersToDspItems uses:
        const config = PROVIDER_CONFIG[provider.key];
        const label = config?.label ?? provider.label;
        color: config?.accent ?? LIBRARY_DSP_FALLBACK_COLOR

     Similarly, if the provider.key is not in PROVIDER_CONFIG, we use the fallback color and the provider.label.

   - However, note that in the LibrarySurface, we are only using PROVIDER_CONFIG. We are not using DSP_CONFIGS. This is because the Library table is showing the provider links from the asset, and we want to use the brand color and label from the canonical PROVIDER_CONFIG. This is acceptable.

   - But what if a provider key is in the asset's providers but not in PROVIDER_CONFIG? Then we use the fallback color and the provider.label (which comes from the asset). This is acceptable.

   - One potential issue: in the LibrarySurface, we are not handling the case where the provider.key is missing from PROVIDER_CONFIG and the provider.label is also missing? But the asset's provider should have a label. We assume the data is correct.

   - In the ShellReleaseRow, we are not changing the data, only the presentation.

   - In DspAvatarStack, we are now showing multiple avatars. We must ensure that the maxVisible prop is passed correctly. In LibrarySurface, we pass maxVisible={3}. In ShellReleaseRow, we pass maxVisible={3}. The default is 1, so existing uses (if any) are unchanged.

   - We must check if there are other uses of DspAvatarStack. The diff does not show any, but we are only changing two call sites. However, note that the original DspAvatarStack was used in ShellReleaseRow without the maxVisible prop (so it defaulted to 1). Now we are adding maxVisible={3} to ShellReleaseRow and also using it in LibrarySurface with maxVisible={3}. We are not changing any other call sites.

   - We must ensure that the prop is optional and defaults to 1. We did that.

   - One more thing: in DspAvatarStack, we are now rendering multiple avatars. We must ensure that the accessible label or aria-label is still correct. The component still has a button that triggers the popover. The popover (which we don't see in the diff) is expected to list all providers. We are not changing the popover trigger, so it should still work.

   - However, note that the button now contains multiple avatars. The screen reader might announce each avatar? We are not changing the aria-label or role. The button is still a button. The inner content is now multiple spans. This might be confusing for screen readers because they might read each avatar's glyph and status. But note that the avatars are purely decorative and the status is already conveyed by the popover. We should consider hiding the avatars from screen readers and relying on the button's label or the popover.

   - Looking at the original code: the primary avatar had no aria-label and the status dots were aria-hidden. The button itself had no aria-label. We are now adding more avatars, each with a glyph and status dots (also aria-hidden). The button still has no explicit label.

   - This might be an accessibility issue because the button does not have an accessible name. However, the original component also did not have one. We are not making it worse by adding more avatars? Actually, we are adding more content that might be announced. But note that the glyphs are just letters and the status dots are aria-hidden. So the screen reader would read the glyph (e.g., "S", "A", "Y", "T") for each avatar? That would be confusing.

   - We should consider adding an aria-label to the button that describes the state (e.g., "Provider status: 2 live, 2 missing") or at least hide the inner content from screen readers and rely on the popover for the information.

   - However, the original component had the same issue: it only showed one avatar (the primary) and its glyph. So if we are not changing the accessibility, then we are not introducing a new issue. But note that the original component was also not accessible. Since the task is about the display change and we are not asked to fix accessibility, we might leave it. However, as a reviewer, we should point out that the component lacks an accessible name.

   - But note: the popover is triggered by the button and lists all providers with their status. The button might be considered to have an accessible name via the popover? Not exactly. The button should have a label that describes its function.

   - However, the issue is pre-existing and not introduced by this PR. We are only asked to review the changes in this PR. Therefore, we should not flag it as a new issue unless we are making it worse.

   - We are adding more avatars, which means more glyphs that might be read by a screen reader. This could be more confusing. Therefore, we should consider hiding the avatars from screen readers.

   - Let's look at the code for the avatars in DspAvatarStack:

        {visible.map((dsp, index) => (
          <span
            key={dsp.id}
            className={...}
            style={...}
          >
            {dsp.glyph}
            {dsp.status === 'pending' && ( ... )}
            {dsp.status === 'error' && ( ... )}
          </span>
        ))}

   - We are not setting aria-hidden on these spans. The glyph is a letter and the status dots are aria-hidden. The glyph will be read by screen readers.

   - We should add `aria-hidden='true'` to the span that contains the glyph and the status dots? But note that the status dots are already aria-hidden. The glyph is not.

   - Alternatively, we can set `aria-hidden='true'` on the entire span for each avatar.

   - However, note that the original code did not have this issue because it only showed one avatar. We are now showing multiple. This is a change that might make the accessibility worse.

   - Since we are not required to fix accessibility in this PR, but we are introducing a potential regression in accessibility (by making the button's inner content more complex and potentially confusing), we should at least consider it.

   - But note: the button is still focusable and the popover is triggered on hover/focus. The popover content is accessible. The button itself might be announced as "button" without a label, which is not ideal. However, the original had the same problem.

   - Given that we are not changing the button's accessibility properties and the original component had the same issue, we might not flag it. However, if we want to be strict, we note that the accessibility of the button is not improved and we are adding more content that is not accessible.

   - But the instructions say: "Correctness/security/requirements only". Accessibility is a correctness issue? It can be, but the project conventions do not mention accessibility. We see in the conventions: "No native browser dialogs (alert/confirm/prompt)" and "No emoji in UI — use icons", but nothing about accessibility.

   - Therefore, we might not flag it. However, let's see if there is a specific requirement in the project for accessibility. The GBrain context does not show any.

   - We'll leave it as a note but not flag it as a blocker.

   - Another issue in DspAvatarStack: the variable `visible` is defined as:
        const visible = ordered.slice(0, Math.max(1, maxVisible));
     What if maxVisible is 0? Then we show 1 avatar. What if maxVisible is negative? Same. We are using Math.max(1, maxVisible) so we always show at least one. This is acceptable.

   - What if maxVisible is greater than the number of dsps? Then we show all and others=0. This is correct.

   - The zIndex: we are setting zIndex: visible.length - index. For example, if visible.length=3, then:
        index=0 -> zIndex=3
        index=1 -> zIndex=2
        index=2 -> zIndex=1
     This means the first avatar (index0) is on top, then the next, then the last at the bottom. This creates a stack where the first is the topmost. This is acceptable.

   - However, note that the original design was to have the primary avatar (the most actionable) on top. We are sorting by status (live first, then pending/error, then missing). So the first in the ordered array is the most actionable. We are then showing the first `maxVisible` in the ordered array. And we are setting the zIndex so that the first in the visible array (which is the most actionable) has the highest zIndex and is on top. This is correct.

   - We are using `zIndex: visible.length - index` which is an integer. This is acceptable.

   - We are not using any negative zIndex or extreme values.

   - No logic error found.

Security:
   - We are not introducing any new external API calls or changing data flow in a way that introduces security issues.
   - We are not using any dangerous innerHTML or similar.
   - We are not introducing any new dependencies.
   - No security issues found.

Data integrity:
   - We are not changing how data is stored or fetched. We are only changing the display.
   - We are not introducing any new database queries or mutations.
   - No data integrity issues found.

Error handling:
   - We are not introducing any new asynchronous operations or changing error handling in existing ones.
   - The functions we are changing are synchronous and do not throw exceptions under normal circumstances (assuming the input data is as expected).
   - We are using optional chaining and fallbacks to avoid null dereferences.
   - No error handling issues found.

Cross-file impact:
   - We are changing the shape of the DspAvatarStack component by adding a maxVisible prop. We are updating two call sites: LibrarySurface and ShellReleaseRow. We are not changing any other call sites (if they exist). We must ensure that if there are other call sites, they still work because the prop is optional and defaults to 1.
   - We are not changing the type of the dsps prop.
   - We are not changing the return type of the component.
   - Therefore, other call sites should continue to work as before (showing one avatar and the overflow chip).

   - However, we must check if there are other call sites of DspAvatarStack in the codebase. The diff does not show them, but we are only allowed to judge by the diff. Since we are not given the entire codebase, we assume that the only call sites are the ones we are changing or that other call sites are unaffected because

itstimwhite (commented):

PR #13157, opened 2026-07-04
fix(ios): replace python3 with plutil in write-configuration.sh

### LINKED ISSUE #11003: Candidate follow-up: make iOS write-configuration.sh resilient to a broken python3 on PATH

## Source

* Current issue: ad-hoc (2026-06-16 session, while shipping JovieInc/Jovie#10988/JovieInc/Jovie#10995)
* Source: `apps/ios/scripts/write-configuration.sh`

## Follow-up

`write-configuration.sh` generates `Configuration.local.plist` via `python3 - <<PY ... plistlib ...`. On a machine where `python3` resolves to a broken Homebrew interpreter (observed: `python@3.14` with a `pyexpat`/`libexpat` symbol mismatch — `_XML_SetAllocTrackerActivationThreshold`), the whole `pnpm run ios:setup-env` / `ensure-configuration.sh` path fails, which blocks local iOS builds. Worked around this session by writing the plist by hand.

Make the script robust: prefer `/usr/bin/plutil` (always present on macOS, no Python) to emit the plist, or fall back to `/usr/bin/python3` explicitly instead of the first `python3` on PATH.

## Why it matters

Removes a sharp edge that silently blocks local iOS dev/builds on otherwise-fine machines; CI is unaffected (clean python) so it won't surface there.

## Classification

Candidate

Decide: rewrite with `plutil`, or pin `/usr/bin/python3`? Either should make `apps/ios/scripts/ensure-configuration.sh` succeed regardless of a broken Homebrew `python3`.

## Dependency

None.

itstimwhite:
## Summary

- `apps/ios/scripts/write-configuration.sh` now uses `/usr/bin/plutil` (always available on macOS) instead of a `python3` heredoc to write `Configuration.local.plist`. Fixes local iOS builds that fail when a broken Homebrew interpreter (e.g. `python@3.14` with a `pyexpat`/`libexpat` `_XML_SetAllocTrackerActivationThreshold` symbol mismatch) is first on PATH.
- Adds a `TARGET_PLIST` env override so tests can redirect script output to a temp directory without needing the repo structure.
- Adds `apps/ios/scripts/write-configuration.test.mjs` — 4 regression tests using `node:test`: broken-python resilience, all 9 key correctness, empty optionals, and `plutil -lint` well-formedness.
- `ios-ci.yml` runs the new test suite between the best-practices lint and xcodebuild steps.

## Root cause

The script's `python3` heredoc was using the system PATH to find Python. On machines where `brew install python@3.14` or a Python upgrade left a broken interpreter first on PATH (manifesting as `ImportError: cannot import name 'plistlib'` or a `libexpat` symbol mismatch), the script would fail and block `git commit` hooks and Xcode builds.

## Fix

`/usr/bin/plutil` is always at that absolute path on macOS (it is part of `CoreFoundation`). `plutil -create xml1` + `plutil -insert -string` handles XML escaping, empty strings, and special characters identically to `plistlib.dump`. Key ordering changes (alphabetical vs insertion order) but `NSDictionary` lookups in Swift are key-based so this is functionally equivalent.

## Test plan

- [x] `node --test apps/ios/scripts/write-configuration.test.mjs` — 4/4 pass locally
- [x] Pre-push hooks (Biome, typecheck, proxy guard, Tailwind) — all green
- [ ] `ios-ci.yml` "Test write-configuration script" step — will verify on macOS runner

Fixes #11003

### DIFF
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -38,6 +38,9 @@ jobs:
       - name: iOS best-practices lint
         run: bash scripts/ios-best-practices-lint.sh apps/ios/Jovie
 
+      - name: Test write-configuration script
+        run: node --test apps/ios/scripts/write-configuration.test.mjs
+
       - name: Ensure iOS local configuration exists
         run: bash apps/ios/scripts/ensure-configuration.sh
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project uses [Calendar Versioning](https://calver.org/) (`YY.M.PATCH`).
 
 ## [Unreleased]
 
+- [internal] **iOS write-configuration resilience (GH-11003)**: Rewrote `apps/ios/scripts/write-configuration.sh` to use `/usr/bin/plutil` (always present on macOS) instead of `python3`, preventing failures when a broken Homebrew interpreter (e.g. pyexpat/libexpat symbol mismatch with python@3.14) is first on PATH. Adds a `TARGET_PLIST` env override for testability and four regression tests (`write-configuration.test.mjs`) run in `ios-ci.yml`.
 - **Smoother icon transitions (GH-11472)**: Copy buttons and the mobile menu now morph between states with a soft, interruptible animation instead of snapping. Adds shared animation primitives (`lib/animation/motion-primitives.ts` + `AnimatedIconSwap`) for icon swaps, staggered entrances, and layered depth shadows.
 - [internal] **Tool discovery skill (GH-13124)**: New `.claude/skills/tool-discovery/SKILL.md` — when a link/tool is shared without context, the agent searches GitHub (`gh search repos`) and the web for docs/pricing/reviews and returns a structured evaluation instead of asking for manual research or a gated unlock action.
 - [internal] **Ovie HUD auto-reload (GH-12988)**: Electron now polls `/api/health/build-info` every 60 seconds and reloads `/hud` windows when the deployed commit/build fingerprint changes, with no-store build-info responses so running operator shells pick up new HUD deploys without a manual restart.
--- a/apps/ios/scripts/write-configuration.sh
+++ b/apps/ios/scripts/write-configuration.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 IOS_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
-TARGET_PLIST="$IOS_DIR/Jovie/Configuration.local.plist"
+TARGET_PLIST="${TARGET_PLIST:-$IOS_DIR/Jovie/Configuration.local.plist}"
 
 CLERK_PUBLISHABLE_KEY="${CLERK_PUBLISHABLE_KEY:-${NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY:-pk_test_ci_placeholder}}"
 API_BASE_URL="${API_BASE_URL:-http://localhost:3100}"
@@ -18,27 +18,20 @@ OBSERVABILITY_INGEST_SECRET="${JOVIE_IOS_OBSERVABILITY_INGEST_SECRET:-${OBSERVAB
 CLERK_REDIRECT_URL="${CLERK_REDIRECT_URL:-${JOVIE_IOS_CLERK_REDIRECT_URL:-ie.jov.jovie://callback}}"
 CLERK_CALLBACK_URL_SCHEME="${CLERK_CALLBACK_URL_SCHEME:-${JOVIE_IOS_CLERK_CALLBACK_URL_SCHEME:-ie.jov.jovie}}"
 
-python3 - <<PY
-import plistlib
-from pathlib import Path
+mkdir -p "$(dirname "$TARGET_PLIST")"
 
-target = Path(r"$TARGET_PLIST")
-target.parent.mkdir(parents=True, exist_ok=True)
-
-payload = {
-    "ClerkPublishableKey": r"$CLERK_PUBLISHABLE_KEY",
-    "ApiBaseUrl": r"$API_BASE_URL",
-    "WebBaseUrl": r"$WEB_BASE_URL",
-    "SentryDsn": r"$SENTRY_DSN",
-    "ObservabilityEnvironment": r"$OBSERVABILITY_ENVIRONMENT",
-    "ObservabilityIngestUrl": r"$OBSERVABILITY_INGEST_URL",
-    "ObservabilityIngestSecret": r"$OBSERVABILITY_INGEST_SECRET",
-    "ClerkRedirectUrl": r"$CLERK_REDIRECT_URL",
-    "ClerkCallbackUrlScheme": r"$CLERK_CALLBACK_URL_SCHEME",
-}
-
-with target.open("wb") as file_handle:
-    plistlib.dump(payload, file_handle, sort_keys=False)
-PY
+# ponytail: use /usr/bin/plutil (always present on macOS) instead of python3 to
+# avoid breakage from a broken Homebrew python on PATH (e.g. pyexpat/libexpat
+# symbol mismatch with python@3.14).
+/usr/bin/plutil -create xml1 "$TARGET_PLIST"
+/usr/bin/plutil -insert ClerkPublishableKey    -string "$CLERK_PUBLISHABLE_KEY"    "$TARGET_PLIST"
+/usr/bin/plutil -insert ApiBaseUrl             -string "$API_BASE_URL"             "$TARGET_PLIST"
+/usr/bin/plutil -insert WebBaseUrl             -string "$WEB_BASE_URL"             "$TARGET_PLIST"
+/usr/bin/plutil -insert SentryDsn              -string "$SENTRY_DSN"               "$TARGET_PLIST"
+/usr/bin/plutil -insert ObservabilityEnvironment -string "$OBSERVABILITY_ENVIRONMENT" "$TARGET_PLIST"
+/usr/bin/plutil -insert ObservabilityIngestUrl -string "$OBSERVABILITY_INGEST_URL" "$TARGET_PLIST"
+/usr/bin/plutil -insert ObservabilityIngestSecret -string "$OBSERVABILITY_INGEST_SECRET" "$TARGET_PLIST"
+/usr/bin/plutil -insert ClerkRedirectUrl       -string "$CLERK_REDIRECT_URL"       "$TARGET_PLIST"
+/usr/bin/plutil -insert ClerkCallbackUrlScheme -string "$CLERK_CALLBACK_URL_SCHEME" "$TARGET_PLIST"
 
 echo "Wrote $TARGET_PLIST"
--- a/apps/ios/scripts/write-configuration.test.mjs
+++ b/apps/ios/scripts/write-configuration.test.mjs
@@ -1,0 +1,119 @@
+import assert from 'node:assert/strict';
+import { execFileSync } from 'node:child_process';
+import {
+  chmodSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+import test from 'node:test';
+import { fileURLToPath } from 'node:url';
+
+const scriptPath = path.resolve(
+  path.dirname(fileURLToPath(import.meta.url)),
+  'write-configuration.sh'
+);
+
+function runScript(t, extraEnv = {}) {
+  const root = mkdtempSync(path.join(tmpdir(), 'write-config-'));
+  t.after(() => rmSync(root, { recursive: true, force: true }));
+  const plistPath = path.join(root, 'Configuration.local.plist');
+  execFileSync('bash', [scriptPath], {
+    env: {
+      ...process.env,
+      CLERK_PUBLISHABLE_KEY: 'pk_test_example',
+      TARGET_PLIST: plistPath,
+      ...extraEnv,
+    },
+    encoding: 'utf8',
+  });
+  return { plistPath, root };
+}
+
+// --- resilience test ---
+
+test('succeeds even when python3 on PATH is broken', t => {
+  const root = mkdtempSync(path.join(tmpdir(), 'broken-python-'));
+  t.after(() => rmSync(root, { recursive: true, force: true }));
+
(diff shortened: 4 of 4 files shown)


itstimwhite:
### Merge activity

* **Jul 4, 8:45 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 9:39 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 4, 9:40 PM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.

itstimwhite:
We are given a PR that changes the iOS write-configuration script from using a Python heredoc to using plutil.
 We are to review for logic errors, security, data integrity, error handling, and cross-file impact.

 Let's break down the changes:

 1. In `apps/ios/scripts/write-configuration.sh`:
    - Changed the way TARGET_PLIST is set: now uses an environment variable override with a default.
    - Removed the Python block and replaced it with a series of `plutil` commands.
    - Added a comment explaining the change.

 2. Added a test file `apps/ios/scripts/write-configuration.test.mjs` with 4 tests.

 3. Updated the CI workflow (`.github/workflows/ios-ci.yml`) to run the new test.

 4. Updated the changelog.

 We are to focus on:
   - Logic errors
   - Security
   - Data integrity
   - Error handling
   - Cross-file impact

 Let's go file by file.

 1. `apps/ios/scripts/write-configuration.sh`:

    Original (before change) had:
        TARGET_PLIST="$IOS_DIR/Jovie/Configuration.local.plist"

    Now:
        TARGET_PLIST="${TARGET_PLIST:-$IOS_DIR/Jovie/Configuration.local.plist}"

    This is a standard way to allow overriding via environment variable. It's safe.

    The Python block was removed and replaced by:

        mkdir -p "$(dirname "$TARGET_PLIST")"
        /usr/bin/plutil -create xml1 "$TARGET_PLIST"
        ... (a series of plutil -insert commands)

    We must check:
      - Does the script still create the directory if it doesn't exist? Yes, the `mkdir -p` is there.
      - Are we using the correct plutil commands? The plutil commands are:
          plutil -create xml1 "$TARGET_PLIST"   -> creates an empty plist in XML1 format.
          Then a series of inserts for each key.

    However, note that the original Python code did:
        payload = { ... }   # a dictionary with 9 keys
        and then dumped it with sort_keys=False (so insertion order preserved).

    The new script inserts the keys in the order:
        ClerkPublishableKey, ApiBaseUrl, WebBaseUrl, SentryDsn, ObservabilityEnvironment,
        ObservabilityIngestUrl, ObservabilityIngestSecret, ClerkRedirectUrl, ClerkCallbackUrlScheme.

    This is the same order as in the original Python payload.

    But note: the original Python code did not sort the keys (sort_keys=False) so the order was insertion order.
    The new script inserts in the same order, so the plist will have the same key order.

    However, note that the plutil tool might not preserve the order? Actually, plutil when inserting does not change the order of existing keys, and we are inserting into an empty plist one by one, so the order will be the insertion order.

    Therefore, the plist produced should be equivalent.

    Potential issues:

      a) What if one of the environment variables contains characters that need XML escaping?
         The original Python code used `plistlib.dump` which properly escapes the string for XML.
         The plutil tool, when inserting a string, should also properly escape the string.

         According to the plutil man page: 
            -insert key value
                   Inserts the specified key and value into the property list.

         And the value is provided as a string. The tool should handle the necessary escaping.

         However, note that the original Python code used raw strings (r"...") for the values, meaning that 
         backslashes were treated literally. But in the context of a plist string, we don't want to interpret 
         backslashes as escapes in the string content? Actually, the plist string value is just text.

         Example: if the value contains a backslash, in the plist it should be represented as a single backslash.
         The plutil tool should handle that.

         But note: the original Python code did not do any extra escaping beyond what plistlib does.

         We must trust that plutil does the right thing. Since plutil is the Apple-provided tool for manipulating 
         plists, it should be correct.

      b) What if the directory creation fails? We have `set -euo pipefail` so the script will exit on any error.

      c) What if plutil fails? Similarly, the script will exit.

    However, note that the original script did not check for the success of the Python block either (it relied on 
    the Python interpreter exiting with 0 on success). The new script relies on plutil exiting with 0.

    This is acceptable.

    But note: the original script did not create the directory if it didn't exist? Actually, it did:
        target.parent.mkdir(parents=True, exist_ok=True)

    The new script does:
        mkdir -p "$(dirname "$TARGET_PLIST")"

    So same effect.

    Therefore, no logic error found in the script.

 2. The test file `apps/ios/scripts/write-configuration.test.mjs`:

    We are to check for logic errors in the tests.

    The tests are:

      a) 'succeeds even when python3 on PATH is broken'
         - Creates a fake python3 that exits with 1 and puts it first in PATH.
         - Runs the script and checks that the plist is written and contains the expected key.

         This test is valid.

      b) 'writes all nine expected plist keys'
         - Sets all 9 environment variables and runs the script (with TARGET_PLIST set to a temp file).
         - Then checks that the plist contains each key and two of the values (for ApiBaseUrl and ObservabilityEnvironment and ClerkRedirectUrl).

         Note: It only checks three of the values? Actually, it checks:
            - ApiBaseUrl: 'https://api.jov.ie'
            - ObservabilityEnvironment: 'staging'
            - ClerkRedirectUrl: 'ie.jov.jovie://callback'

         But it does check for the presence of all 9 keys.

         Why not check all values? It might be because the test is focused on the keys and a couple of values to 
         ensure the insertion works. However, note that the test for empty optionals and the well-formedness test 
         might cover other aspects.

         But we are only concerned with logic errors. The test does not have an obvious logic error.

      c) 'handles empty optional values without breaking the plist'
         - Runs the script without setting the optional ones (SentryDsn, ObservabilityIngestUrl, ObservabilityIngestSecret, etc.) 
           so they default to empty strings.
         - Then checks that the plist is valid XML (starts with '<?xml') and that the SentryDsn key is present and 
           the ClerkPublishableKey is present.

         This test is valid.

      d) 'produces a well-formed plist that plutil accepts'
         - Runs the script with default values (so all variables are set to their defaults, which are mostly empty or placeholder) 
           and then runs `plutil -lint` on the generated plist.

         This test is valid.

    However, note that in the test for empty optionals, the test does:
        assert.match(contents, /^<\?xml/, 'not a valid XML plist');

    But note: the plist generated by plutil -create xml1 will start with:
        <?xml version="1.0" encoding="UTF-8"?>
        <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
        <plist version="1.0">
        <dict>
        ... 
        </dict>
        </plist>

    So the first line is indeed the XML declaration.

    But note: the test uses `^<\?xml` which matches the start of the string and then the literal string "<?xml". 
    This is correct.

    However, note that the plutil output might have a newline at the end? The test reads the file as UTF8 and then 
    does a regex match. The `^` matches the start of the string, so if there's a leading newline it would fail.

    But the plutil tool does not output a leading newline. It starts with the XML declaration.

    Therefore, the test is safe.

    But wait: what if the plutil output has a BOM? It shouldn't because we are creating an XML plist and plutil 
    does not add a BOM by default.

    Therefore, no logic error found in the tests.

 3. The CI workflow change:

    We added a step:
        - name: Test write-configuration script
          run: node --test apps/ios/scripts/write-configuration.test.mjs

    This step runs the tests we just discussed.

    There is no logic error here.

 4. The changelog:

    We added a bullet point describing the change.

    No logic error.

 However, note: the original script had a potential issue that the new script fixes: 
    It was using `python3` from the PATH, which could be broken.

    The new script uses `/usr/bin/plutil` which is always present on macOS.

    Therefore, the change is safe.

 But wait: what about the environment variable `TARGET_PLIST`? 
    The script now allows overriding the target plist location via `TARGET_PLIST`. 
    This is used in the tests to write to a temporary directory.

    However, note that the script also uses `TARGET_PLIST` in the `mkdir -p` command and in the plutil commands.

    This is safe.

 One potential issue: 
    The script uses:
        TARGET_PLIST="${TARGET_PLIST:-$IOS_DIR/Jovie/Configuration.local.plist}"

    What if `TARGET_PLIST` is set to an empty string? Then it would become the default? 
    Actually, no: if `TARGET_PLIST` is set to an empty string, then the expansion becomes the empty string, 
    and then we would be trying to write to an empty string as a file path.

    But note: the `${VAR:-default}` syntax uses the default only if VAR is unset or set to the empty string? 
    Actually, in bash, if VAR is set to the empty string, then it is considered set and the empty string is used.

    From the bash manual: 
        ${parameter:-word}
            Use Default Values.  If parameter is unset or null, the expansion of word is substituted. 
            Otherwise, the value of parameter is substituted.

    So if TARGET_PLIST is set to the empty string, then we use the empty string.

    This is dangerous because then we would be trying to create a file at an empty string path.

    However, note that the script also does:
        mkdir -p "$(dirname "$TARGET_PLIST")"

    If TARGET_PLIST is empty, then dirname of an empty string is '.' (current directory). 
    So we would try to create the current directory (which exists) and then try to write the plist to an empty string.

    The plutil commands would then fail because they would try to write to a file named "" (empty string) in the current directory.

    But note: the script has `set -e` so it would exit on the first error (the mkdir would succeed, but the plutil -create would fail).

    However, this is a misuse of the environment variable. The tests set TARGET_PLIST to a valid path.

    But what if a user sets TARGET_PLIST to the empty string? Then the script would fail.

    Is this a regression? 
        In the original script, TARGET_PLIST was hardcoded. There was no way to set it to empty string via environment.

    So the original script did not have this problem.

    We have introduced a potential footgun: if someone sets TARGET_PLIST to the empty string, the script breaks.

    However, note that the script is intended to be used in a controlled environment (the iOS build and the tests). 
    The tests set it to a temporary directory, and the build does not set it (so it uses the default).

    We should consider whether we want to guard against empty string.

    But note: the environment variable is also used in the test to override the location. The test sets it to a non-empty string.

    We can argue that it is the responsibility of the caller to not set it to an empty string.

    However, to be safe, we might want to check that TARGET_PLIST is non-empty? 
    But the script does not do that.

    Alternatively, we could change the defaulting to only use the default if TARGET_PLIST is unset, and if set to empty string then treat it as unset? 
    But that would be a change in behavior.

    Since the original script did not have this variable at all, and we are adding it for testability, 
    and the only callers (the build and the tests) set it appropriately, we can leave it as is.

    However, note that the script is also called by `ensure-configuration.sh`? 
    Let's look at the context: the PR does not change that script, but we see in the CI workflow:

        - name: Ensure iOS local configuration exists
          run: bash apps/ios/scripts/ensure-configuration.sh

    We don't have the code for `ensure-configuration.sh` in the diff, but we can assume it does not set TARGET_PLIST to empty.

    Therefore, this is not a bug in the current context.

    But note: the PR description says: "Adds a TARGET_PLIST env override so tests can redirect script output to a temp directory without needing the repo structure."

    So the override is for tests and possibly for other scripts? We don't see any other script setting it.

    Given that, and that the build does not set it (so it uses the default), we are safe.

    However, to be thorough, let's consider if we should guard against empty string.

    We could change the line to:

        TARGET_PLIST="${TARGET_PLIST:-$IOS_DIR/Jovie/Configuration.local.plist}"
        if [ -z "$TARGET_PLIST" ]; then
            echo "Error: TARGET_PLIST is empty" >&2
            exit 1
        fi

    But the PR does not do that, and we are only to review the given diff.

    Since the diff does not include such a guard, and the problem is only theoretical (and not introduced by the change in a way that affects the intended use cases), 
    we might not consider it a bug that needs fixing in this PR.

    However, note: the change introduces a new way to fail (by setting TARGET_PLIST to empty) that did not exist before.

    But the original script did not have this environment variable at all, so the only way to get an empty string would be if the hardcoded path was empty? 
    Which it isn't.

    Therefore, we are introducing a new potential failure mode. 

    However, the PR author might argue that the environment variable is intended to be set to a valid path, and it's the caller's responsibility.

    Given the context of the PR (fixing a specific breakage due to Python) and the fact that the override is for tests, 
    and the tests set it correctly, we can accept this.

    But note: the PR description says: "Adds a TARGET_PLIST env override so tests can redirect script output to a temp directory"

    So the primary use is for tests. The build does not use it.

    Therefore, we will not flag this as a bug.

    However, let me note: if we were to be extremely cautious, we might want to add a check. 
    But since the diff does not include it and we are only to review the given changes, we will not suggest it as a required fix.

    Alternatively, we can note it as a potential improvement but not a bug.

    Since the task is to find bugs, and this is a potential issue only in a misconfiguration that is not part of the intended use, 
    we will not list it as a bug.

 5. Security:

    We are to look for hardcoded secrets, SQL injection, XSS, auth bypass, CSP violations.

    In the script:
        We are using environment variables to set the values in the plist.
        The environment variables are:
            CLERK_PUBLISHABLE_KEY, API_BASE_URL, WEB_BASE_URL, SENTRY_DSN, 
            OBSERVABILITY_ENVIRONMENT, OBSERVABILITY_INGEST_URL, OBSERVABILITY_INGEST_SECRET,
            CLERK_REDIRECT_URL, CLERK_CALLBACK_URL_SCHEME.

        None of these are hardcoded in the script. They are taken from the environment.

        The script does not print them to the console in a way that would leak (except the echo at the end: 
            echo "Wrote $TARGET_PLIST"

        This only prints the path, not the content.

        Therefore, no hardcoded secrets.

        There is no SQL, so no SQL injection.

        The plist is an XML file that is read by the iOS app. We are not generating HTML, so no XSS.

        No auth bypass or CSP issues.

    In the test:
        We are using hardcoded test values (like 'pk_test_example', 'https://api.jov.ie', etc.) 
        but these are clearly test values and not secrets.

        The test writes to a temporary directory and cleans up.

        Therefore, no security issues.

 6. Data integrity:

    We are to look for missing validation, incorrect DB queries, migration issues.

    There are no DB queries or migrations in this change.

    The script is generating a plist file. We have already discussed that the plutil commands should produce 
    a valid plist equivalent to the original Python script.

    The tests check:
        - That the plist is written when Python is broken (resilience)
        - That all nine keys are present (correctness)
        - That empty optionals are handled (so we don't missing keys when the value is empty string)
        - That the plist is well-formed (by running plutil -lint)

    Therefore, data integrity seems intact.

 7. Error handling:

    The script has `set -euo pipefail` so it will exit on any error, unset variable, or failed pipe.

    The original script did not have `set -e`? Actually, we don't see the original script in the diff, 
    but we can assume it did not have it because the Python block was not checked for errors.

    However, note: the original

itstimwhite (commented):

PR #13163, opened 2026-07-04
Homepage: System B token sweep + hero CSS + unlock all sections

itstimwhite:
### System B Token Sweep
All raw colors in `home.css` converted to System B CSS custom properties:
- `#020303` → `--color-bg-base` (section backgrounds)
- `rgba(255,255,255,0.96)` → `--color-text-primary-token`
- `rgba(255,255,255,0.94/0.88)` → `--color-text-secondary-token`
- `rgba(255,255,255,0.6)` → `--color-text-tertiary-token`
- `rgba(255,255,255,0.38)` → `--color-text-quaternary-token`
- Button background/text raw values → `--system-b-primary-bg/fg`
- Border raw values → `--color-border-default`

### Hero CSS Improvements
- Headline weight 800 → 600 (per DESIGN.md)
- Primary CTA: canonical 44px min-height
- Secondary CTA: canonical 36px min-height
- Satoshi brand display font persists at all breakpoints (was dropping to system-ui on mobile)
- Tighter hero top padding: `clamp(4.5rem, 8.5vh, 6.5rem)`

### Product Statement
- Body text upgraded from quaternary to tertiary token (actually readable)

### Flag Flips
- `SHOW_HOMEPAGE_UNLOCKED_SECTIONS: true` — 12 V2 sections now render below the fold
- `SHOW_HOMEPAGE_V2_FINAL_CTA: true` — page no longer dead-ends at FAQ

### Test Updates
- Removed hero-only collapsing assertions, updated to reflect always-on flags

## Verification
- All 6 static flags tests pass
- Braces balanced (728/728)
- Zero target raw colors remaining in `home.css`
- Token count: Primary 30x, Secondary 13x, Tertiary 18x, Quaternary 2x, Bg base 8x

PR #13207, opened 2026-07-05
Homepage: System B token sweep + hero CSS + unlock all sections

itstimwhite:
### System B Token Sweep
All raw colors in `home.css` converted to System B CSS custom properties. See individual commits.

### Hero CSS Improvements
- Headline weight 800 → 600 (per DESIGN.md)
- Primary CTA: canonical 44px min-height
- Secondary CTA: canonical 36px min-height  
- Satoshi brand display font persists at all breakpoints
- Tighter hero top padding: `clamp(4.5rem, 8.5vh, 6.5rem)`

### Product Statement
- Body text upgraded from quaternary to tertiary token

### Flag Flips
- `SHOW_HOMEPAGE_UNLOCKED_SECTIONS: true` — 12 V2 sections now render below the fold
- `SHOW_HOMEPAGE_V2_FINAL_CTA: true` — page no longer dead-ends at FAQ

Supersedes #13163.


itstimwhite:
We are given a PR diff with changes to three files. We are to focus on:
 1. Logic errors
 2. Security
 3. Data integrity
 4. Error handling
 5. Cross-file impact

 We are to ignore style, formatting, and known false positives.

 1. apps/web/app/(home)/home.css
    - This is a CSS file. We are changing raw colors to System B CSS custom properties.
    - We are also adjusting some values (like font-weight, min-height, padding, etc.)

    Since it's CSS, we are less likely to have logic errors, security issues, or data integrity issues.
    However, we must check for:
      - Any use of expressions that might be unsafe (but CSS doesn't have that in the same way as JS)
      - Any potential for breaking the layout in a way that causes security issues (unlikely)
      - Any potential for causing a runtime error (CSS errors are usually just visual)

    But note: we are using CSS custom properties (variables). We must ensure that the variables we are using are defined.

    Looking at the changes, we see:
      --linear-text-primary: var(--color-text-primary-token);
      ... and many others.

    We must check if these variables are defined elsewhere (likely in a global CSS file or via a theme provider).

    However, note that the PR description says: "All raw colors in `home.css` converted to System B CSS custom properties."

    We are not given the context of where these variables are defined, but we can assume they are defined in a base stylesheet.

    Since we are only reviewing the diff, and the diff only shows the usage, we cannot see the definition.

    But note: the PR is about converting to System B tokens, so we assume the tokens are defined.

    However, we must check for any potential typo in the variable names.

    Let's list the variable names we see in the diff:

      --color-text-primary-token
      --color-text-secondary-token
      --color-text-tertiary-token
      --color-text-quaternary-token   [Note: in the product statement section, we see a change from quaternary to tertiary? Actually, let's check:]

    In the product statement section:

      .homepage-product-statement {
        ... 
        color: var(--color-text-quaternary-token);   [changed from rgba(255,255,255,0.38)]
      }

      Then inside:
        .homepage-product-statement__lead: color: var(--color-text-primary-token);
        .homepage-product-statement__body: color: var(--color-text-tertiary-token);   [changed from rgba(255,255,255,0.38)]
        ... and so on.

    We also see:
        --color-bg-base
        --system-b-primary-bg
        --system-b-primary-fg
        --color-border-default

    We must check if these are defined. Since we don't have the base file, we cannot be 100% sure, but note:

    - The PR is part of a sweep to convert to System B, so it's likely that these tokens are defined in a base file that is included.

    However, we must note that if any of these variables are not defined, the CSS will fall back to the browser's default (which might break the design) but it's not a security or logic error.

    But wait: the PR also changes the background of several sections from `#020303` to `var(--color-bg-base)`. If `--color-bg-base` is not defined, then the background becomes the default (transparent or whatever the parent is) which might break the design but again, not a security issue.

    However, note that the PR description says: "System B Token Sweep", so we assume the tokens are defined.

    But we must check for any typo in the variable names.

    Let's check the variable names we see:

      --color-text-primary-token
      --color-text-secondary-token
      --color-text-tertiary-token
      --color-text-quaternary-token   [only in one place: the product statement container]
      --color-bg-base
      --system-b-primary-bg
      --system-b-primary-fg
      --color-border-default

    We see that in the product statement container, we are using `--color-text-quaternary-token` for the container's color, but then inside we are using primary, tertiary, etc.

    This seems consistent.

    However, note that in the product statement section, the container's color was changed from `rgba(255,255,255,0.38)` to `var(--color-text-quaternary-token)`. Then the lead uses primary, the body uses tertiary, etc.

    This is acceptable.

    But we must check: are there any variables that are misspelled?

    I don't see any obvious typos.

    However, note that in the hero section, we changed the headline weight from 800 to 600 (per DESIGN.md). That's a design change, not a bug.

    We also changed the min-height of the primary CTA to 44px and secondary to 36px. That's per the design.

    We changed the hero top padding to `clamp(4.5rem, 8.5vh, 6.5rem)`.

    We changed the font-family of the hero headline to use the Satoshi token.

    All these are design changes and not in the scope of logic/security/etc.

    However, note: we are using `var(--font-satoshi)`. We must check if that is defined.

    Again, we assume it is defined in the base.

    Since we are only reviewing the diff and we don't have the base, we cannot verify the existence of the variables. But note that the PR is part of a sweep, so it's likely that the base has been updated to define these.

    Therefore, in the CSS file, we don't see any logic, security, data integrity, or error handling issues.

    But wait: what about the use of `!important`? We see:

        top: 0 !important;
        min-height: var(--linear-header-height);
        border-bottom: 1px solid transparent !important;
        background: transparent !important;

    The use of `!important` is sometimes a code smell, but it's not a bug per se. It might be necessary to override other styles.

    However, note that the PR description does not mention any issue with `!important` and we are not to flag style issues.

    So, for the CSS file, we have no issues to report in the categories we are focusing on.

 2. apps/web/lib/flags/marketing-static.ts
    - This is a TypeScript file that exports feature flags.

    Changes:
      - SHOW_HOMEPAGE_V2_FINAL_CTA_CTA: changed from false to true
      - SHOW_HOMEPAGE_UNLOCKED_SECTIONS: changed from false to true

    We also see that the test file was updated accordingly.

    We must check for:
      - Logic errors: Are we setting the flags correctly? 
        The change is straightforward: two flags are being turned on.

      - Security: No secrets are being exposed here.

      - Data integrity: Not applicable.

      - Error handling: Not applicable.

      - Cross-file impact: 
          The flags are used in the marketing pages to conditionally render sections.
          Turning these flags on will cause the unlocked sections and the final CTA to appear.

          We must check if there are any callers that might break if these flags are true.

          However, note that the PR description says: "Supersedes #13163", meaning it's replacing a previous PR that had these flags off.

          Also, the test file was updated to reflect that these flags are now on by default.

    But wait: look at the test file change.

 3. apps/web/tests/unit/marketing/marketing-static-flags.test.ts
    - The test was updated to remove the checks for the two flags that were previously off (SHOW_HOMEPAGEPAGE_UNLOCKED_SECTIONS and SHOW_HOMEPAGE_V2_FINAL_CTA) and also removed two tests that were checking that these flags are off in production.

    The test now only checks that all flags are on except the inverted V1 design flags.

    We must check the test for logic errors.

    The test says:
        it('keeps every static marketing flag default-on except inverted V1 design flags', async () => {
          const { FEATURE_FLAGS } = await importFlags();

          // *_V1_DESIGN flags use inverted semantics: true = OLD design. Production ships V2.
          const invertedV1DesignFlags = new Set([
            'SHOW_HOME_V1_DESIGN',
            'SHOW_PUBLIC_PROFILE_V1_DESIGN',
          ]);

          const defaultOnFlags = Object.entries(FEATURE_FLAGS)
            .filter(([key]) => !invertedV1DesignFlags.has(key))
            .map(([, value]) => value);

          expect(defaultOnFlags.every(Boolean)).toBe(true);
          expect(FEATURE_FLAGS.SHOW_HOME_V1_DESIGN).toBe(false);
          expect(FEATURE_FLAGS.SHOW_PUBLIC_PROFILE_V1_DESIGN).toBe(false);
        });

    This test expects that every flag that is not in the invertedV1DesignFlags set is true.

    We just turned two flags on (SHOW_HOMEPAGE_V2_FINAL_CTA and SHOW_HOMEPAGE_UNLOCKED_SECTIONS) so they should be true and pass the test.

    However, note that the test also expects that the inverted V1 design flags are false (which they are).

    But wait: what about other flags that might be false by design? For example, are there any flags that are intentionally false?

    The test assumes that all flags except the inverted V1 ones are true.

    We must check if there are any flags that are supposed to be false.

    Looking at the flag file, we see:

      export const FEATURE_FLAGS = {
        SHOW_HOMEPAGE_V2_CAPTURE_REACTIVATE: true,
        SHOW_HOMEPAGE_V2_POWER_GRID: true,
        SHOW_HOMEPAGE_V2_PRICING: true,
        SHOW_HOMEPAGE_V2_FINAL_CTA: true,   // we changed this to true
        SHOW_HOMEPAGE_GO_LIVE_SECTION: true,
        SHOW_HOMEPAGE_FAQ: true,
        SHOW_HOMEPAGE_AI_COMPOSER_SECTION: true,
        SHOW_HOMEPAGE_CENTER_NAV: true,
        SHOW_HOMEPAGE_FRIDAY_RHYTHM: true,
        SHOW_HOMEPAGE_UNLOCKED_SECTIONS: true,   // we changed this to true
        SHOW_HOME_REFRESH_2026: true,
        SHOW_MARKETING_FULL_FOOTER: true,
        SHOW_MARKETING_CENTER_NAV: true,
        // ... and the inverted V1 design flags at the bottom
      };

    All the flags we see are set to true, except the two inverted V1 design flags (which are set to false).

    Therefore, the test is correct.

    However, note that the test file also removed two tests that were specifically checking that the two homepage flags are off in production.

    Why were they removed? Because now we are turning them on by default.

    This is acceptable.

    But we must check: are there any other flags that are set to false in the file? We don't see any.

    Therefore, the test passes.

    However, note that the test imports the flags and then checks. We must ensure that the flag file does not have any syntax errors.

    The change is simple: two booleans changed from false to true.

    No logic error in the flag file.

    But wait: what about the cross-file impact? 
        The flags are used in the marketing pages. We are turning on two flags that were previously off.
        This means that the unlocked sections and the final CTA will now render.

        We must check if there are any components that rely on these flags being off and might break when they are on.

        However, note that the PR description says: "Flag Flips" and that it supersedes #13163 (which was likely the PR that turned them off for a test).

        Also, the test was updated to reflect the new default.

        Since we are not given the code that uses these flags, we cannot verify the impact.

        But note: the PR is about the homepage, and the changes in the CSS file are also part of this PR.

        We are to flag only if we can point to specific lines.

        We don't see any obvious issue.

    However, note: the test file change removed two tests. We must check if the removal of tests is justified.

    The two removed tests were:

        it('collapses the homepage to hero + minimal footer in production by default', async () => {
          vi.stubEnv('VERCEL_ENV', 'production');
          vi.stubEnv('NEXT_PUBLIC_SHOW_HOMEPAGE_UNLOCKED_SECTIONS', '');

          const { FEATURE_FLAGS } = await importFlags();

          expect(FEATURE_FLAGS.SHOW_HOMEPAGE_UNLOCKED_SECTIONS).toBe(false);
        });

        it('keeps the homepage story stack off regardless of the env flag', async () => {
          vi.stubEnv('VERCEL_ENV', 'production');
          vi.stubEnv('NEXT_PUBLIC_SHOW_HOMEPAGE_UNLOCKED_SECTIONS', 'true');

          const { FEATURE_FLAGS } = await importFlags();

          expect(FEATURE_FLAGS.SHOW_HOMEPAGE_UNLOCKED_SECTIONS).toBe(false);
        });

    These tests were checking that in production, the flag SHOW_HOMEPAGE_UNLOCKED_SECTIONS is false (and even if you try to set it via env, it remains false).

    But now we have changed the flag to be true by default.

    Therefore, these tests are no longer valid and were removed.

    This is acceptable.

    However, note that the flag file does not show any logic that would override the flag based on VERCEL_ENV or the env variable.

    We must check the flag file to see how the flags are determined.

    Unfortunately, we are not given the full flag file, only the diff.

    But note: the test file change shows that the test was using `vi.stubEnv` to set environment variables and then importing the flags.

    This suggests that the flag file might be reading from environment variables.

    Let's look at the flag file diff: we only see the FEATURE_FLAGS object being exported.

    We don't see any logic that reads from process.env.

    Therefore, the flag file is a static object.

    Then how were the previous tests passing? They were setting environment variables and then importing, but if the flag file doesn't read from env, then the stubEnv would have no effect.

    This suggests that the flag file might have been changed in a way that we don't see in the diff? Or perhaps the flag file does read from env but we are not shown that part.

    However, note the test file change: the test was removed because the flag is now true by default.

    And the flag file diff shows that we set the two flags to true.

    Therefore, if the flag file is static, then the environment stubbing in the test would have no effect, and the flag would always be the value in the object.

    So the previous tests were flawed because they were stubbing env but the flag file didn't use it.

    But wait: the test file we are given in the diff is the updated test file. The old test file (before this PR) had those two tests.

    Now, in the updated test file, we removed those tests.

    This is consistent with the change in the flag file.

    Therefore, we don't see a logic error in the flag file or the test file.

    However, we must note: the test file now only has one test (the first one) and the test for marketing center navigation.

    We are not given the full test file, but the diff shows we removed two tests and changed one.

    We must check if the remaining test is correct.

    The remaining test:

        it('keeps every static marketing flag default-on except inverted V1 design flags', async () => {
          ... 
        });

    This test now passes because we set the two flags to true and the inverted V1 design flags are false.

    But note: what if there are other flags that are false? We don't see any in the diff.

    However, we are only given the diff of the flag file. We don't see the entire flag file.

    But note: the flag file diff shows that we changed two flags from false to true, and the rest are true (as per the lines we see).

    We see:

        SHOW_HOMEPAGE_V2_CAPTURE_REACTURE_REACTIVATE: true,
        SHOW_HOMEPAGE_V2_POWER_GRID: true,
        SHOW_HOMEPAGE_V2_PRICING: true,
        SHOW_HOMEPAGE_V2_FINAL_CTA: true,   // changed
        SHOW_HOMEPAGE_GO_LIVE_SECTION: true,
        SHOW_HOMEPAGE_FAQ: true,
        SHOW_HOMEPAGE_AI_COMPOSER_SECTION: true,
        SHOW_HOMEPAGE_CENTER_NAV: true,
        SHOW_HOMEPAGE_FRIDAY_RHYTHM: true,
        SHOW_HOMEPAGE_UNLOCKED_SECTIONS: true,   // changed
        SHOW_HOME_REFRESH_2026: true,
        SHOW_MARKETING_FULL_FOOTER: true,
        SHOW_MARKETING_CENTER_NAV: true,

    And then the inverted V1 design flags (which we don't see in the diff, but the test mentions) are set to false.

    Therefore, the test should pass.

    However, note that the test file change also removed the checks for the two flags we turned on.

    So the test is now correct.

    Therefore, we don't see any logic, security, data integrity, error handling, or cross-file impact issues in the flag file or the test file.

But wait: we must check for cross-file impact between the flag file and the CSS file? 
    The flag file is used to conditionally render sections in the marketing pages. 
    The CSS file is styling the homepage.

    There is no direct impact: the CSS file is always loaded (it's for the homepage) and the flags control whether certain sections are rendered.

    However, note that the CSS file contains styles for sections that are now being turned on by the flags.

    For example, the CSS file has styles for `.homepage-workflow-strip`, `.homepage-profile-proof-band`, etc.

    If the flags are turned on, these sections will be rendered and styled by the CSS.

    If the flags were off, these sections would not be rendered, so

itstimwhite (commented):

itstimwhite:
Zoe overnight merge sweep: closing/reopening to retrigger CI (Lint failure, PR Ready failure)

Issue #10367, opened 2026-06-09
WorkflowDefinition registry + decouple workflow dispatch from lib/connectors


itstimwhite:
Small-to-medium, well-spec'd. Introduce a first-class workflow registry so adding a workflow kind is a single typed file, not a cron branch buried in the connector layer.

**Problem** — Workflow executors live inside lib/connectors/workflows/ (tight coupling); a second runner sits at lib/workflows/memory/; cron dispatches by hardcoded branch; no shared step/contract abstraction. This does not scale to 'infinite workflows'.

**MRR opportunity** — High-leverage enabler: this is the substrate the Content/Distribution Engine wedge (highest-MRR lane, currently idle) needs to ship cheaply. Effort M.

**Scope** — Create lib/workflows/registry.ts with WorkflowDefinition { kind, inputSchema (zod), executor, requiredConnectors, retry policy }. Move executor logic out of lib/connectors/workflows/ into lib/workflows/. cron process-workflow-runs iterates the registry to dispatch on workflow_runs.kind. Make workflow_runs.kind an enum sourced from the registry.

**Acceptance criteria** — adding a new workflow kind touches one registry file + its executor; existing Gmail->Calendar flow and the memory loop run through the registry unchanged; cron has no per-kind branches; kind values validated against the registry.

**Likely files** — new lib/workflows/registry.ts, lib/connectors/workflows/execute-approved-action.ts (moved), lib/workflows/memory/studio-session-loop.ts, app/api/cron/process-workflow-runs/route.ts, db/schema/connectors.ts (workflow_runs.kind), enums.ts. Risk: in-flight workflow_runs rows during migration. Consider whether to align with the planned Trigger.dev direction.

**Shape & model** — Small/medium well-spec'd; model: Codex.

itstimwhite:
Jovie agent (codex issue shipper) claimed this issue.

Branch: `codex/gh-10367-workflowdefinition-registry-decouple-workflow-di`
Risk: `high`
Model route: `escalation` using `grok-composer-2.5-fast`
Integration branch: not used for this issue

itstimwhite:
Jovie agent (codex issue shipper) released this issue for retry.

Agent was interrupted (status=143) - releasing claim for retry.

Log: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-10367-2026-07-01T22-58-34-666z.log`

itstimwhite:
Jovie agent (codex issue shipper) claimed this issue.

[...]

Issue #10360, opened 2026-06-09
Integrate Agnost AI analytics (Vercel AI SDK via OpenTelemetry)


itstimwhite:
Ships next (P0). Wire **Agnost AI analytics** into Jovie web so every Vercel AI SDK call exports OpenTelemetry traces to Agnost, tagged with real userId/sessionId. Requested by Tim (agent owner).

**Problem** — We have no Agnost-side visibility into LLM calls (latency, cost, errors, per-user/per-session traces) for the Jovie web app's AI features.

**Severity & Priority** — Urgent / P0 per Tim's explicit "ships next" directive. Observability gap on the core AI surface; placed in Todo for current-cycle pickup.

**MRR opportunity score** — Low direct score (it's an enabler, not a revenue feature): Reach=all AI-feature users, Frequency=every AI call, Annoyance/WTP/ATP≈n/a (internal tooling), Effort=S–M. Justified as a founder-directed P0, not by the MRR model — it de-risks and speeds the AI surface that downstream MRR features depend on.

**Impact** — Internal observability for all server-side AI SDK traffic (chat, title generation, evals). No end-user-facing behavior change. Additive to existing Braintrust tracing — does not replace it.

**Evidence / current state** (JovieInc/Jovie @ apps/web):

* No `@opentelemetry/*` packages installed yet → this is the "No OTel yet" path (boot a NodeSDK), NOT the "append a span processor" path.
* `apps/web/instrumentation.ts` already exists and runs Braintrust `initLogger` (Node runtime only, fail-open on init). Agnost's NodeSDK should boot here under the same Node-runtime guard.
* All AI SDK calls funnel through one chokepoint: `apps/web/lib/ai/sdk.ts` re-exports `generateText`/`streamText`/`generateObject`/`streamObject` wrapped by `braintrust.wrapAISDK`. Callers import from `@/lib/ai/sdk`, not `'ai'` (enforced; see `apps/web/lib/chat/README.md` "Telemetry contract").
* Known caller: `apps/web/app/api/chat/conversations/[id]/messages/route.ts` (uses `getSessionContext`) — session/user context is available server-side to populate metadata.
* Agnost org ID: `384ebd06-d9a5-48dd-ba22-7a51e430c173` (a header identifier, not a rotating secret — still store in env, do not hardcode). Endpoint `https://otel.agnost.ai/v1/traces`, header `X-Agnost-Org-ID`.

**Proposed fix**

[...]

Issue #8684, opened 2026-05-14
[P1][Demo] Build release workflow system from imported catalog and playbooks


itstimwhite:
## Evidence
- Screenshot/frame: 90-second founder demo on /demovideo
- Route: /demovideo
- Observed behavior: The demo references catalog import and release workflows created from music-industry playbooks.
- Expected behavior: Imported releases should generate release workflows with tasks for metadata, smart links, pre-save/save flows, press, playlists, content, fan notifications, merch, and reporting.

Shown cinematically in PR #8671; production implementation needed.

## Impact
This turns domain expertise into reusable operating system behavior and makes Jovie useful beyond one campaign moment.

## Likely root cause
Release tasks exist in pieces, but there is not yet a canonical release workflow generator driven by imported catalog state and playbooks.

## Acceptance criteria
- Imported releases generate one canonical workflow.\n- Tasks are deduped across repeated imports.\n- Workflow templates are versioned and auditable.\n- The artist can see task state and generated artifacts from release detail.

## Test plan
- Unit tests for playbook expansion and dedupe keys.\n- Integration test for repeated import not creating duplicate workflows/tasks.\n- Manual release detail check.\n- Migration/data backfill plan before enabling on production catalog imports.

## Suggested fix
Build a release-workflow generator with versioned playbooks, import dedupe, and task/artifact linkage to release detail surfaces.

itstimwhite:
Jovie agent (codex issue shipper) claimed this issue.

Branch: `codex/gh-8684-p1-demo-build-release-workflow-system-from-impor`
Risk: `high`
Model route: `escalation` using `grok-composer-2.5-fast`
Integration branch: not used for this issue

itstimwhite:
Jovie agent (codex issue shipper) stopped on a real blocker.

Agent exited without successfully shipping.

Status: `1`
Signal: `null`
Log: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-8684-2026-07-02T03-08-15-714z.log`
Prompt: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-8684-2026-07-02T03-08-15-714z.prompt.md`

PR #13235, opened 2026-07-05
analytics: agent task result log + release-step classification + opportunity cycle time

### LINKED ISSUE #12140: Opportunity lifecycle cycle-time timer (detected_at + shipped_at)
Child of JovieInc/Jovie#12139.

## Problem

A core VC-traction metric is the 'opportunity detected → opportunity shipped' cycle time — how fast Jovie turns a detected signal into a shipped automation. Today `suggested_actions` has `status` but no `detected_at`; `workflow_runs` has no `shipped_at`. Cycle time cannot be measured.

## What to instrument

* Add `detected_at timestamptz` to `suggested_actions` (set when a detector emits the row).
* Add `shipped_at timestamptz` to `workflow_runs` (set when status→done and the automation's output was published/sent).
* `cycle_time = shipped_at(linked run) − detected_at(suggested_action)`.
* Aggregate: median + p90 cycle time per opportunity type, per artist, rolling 30-day.

## Acceptance

* A query returns median + p90 opportunity→ship cycle time per artist per window.
* Backfill `detected_at` for existing open rows as `created_at` (labeled assumption).

## Dependencies

* Reuses existing `suggested_actions` + `workflow_runs` tables; links via `step_outputs->>'approvalId'`.

### LINKED ISSUE #12144: Manual-vs-automated task classification (manual tasks eliminated per release cycle)
Child of JovieInc/Jovie#12139.

## Problem

A VC-traction metric is 'N manual tasks eliminated per release cycle' — proof that Jovie automation replaces human labor. Today there is no classification of which release-cycle steps were done by automation vs would-have-been-manual, so this count cannot be produced.

## What to instrument

* Define the canonical release-cycle step list (metadata, smart links, pre-save, press, playlists, content posts, fan notifications, merch, reporting — per JovieInc/Jovie#8684 demo workflow).
* For each step, tag the execution source: `automation` (workflow_run done) vs `manual` (artist action with no linked run) vs `skipped`.
* `manual_tasks_eliminated per release = (steps that would be manual) − (steps actually manual)`; proxy baseline = steps a non-Jovie artist does manually (from Tim's 15-yr workflow canon, labeled assumption).

## Acceptance

* A query returns, per release: count of automated vs manual vs skipped steps, and manual-tasks-eliminated vs the labeled baseline.
* Time-saved proxy = (manual steps eliminated × per-step minutes from the workflow canon).

## Dependencies

* Reuses `workflow_runs` typed by JovieInc/Jovie#10367; step list anchored to the release-workflow demo (JovieInc/Jovie#8684).

### LINKED ISSUE #12145: Agent task result log (workflow-level success / override / cost-per-opportunity)
Child of JovieInc/Jovie#12139.

## Problem

Jovie's multi-agent contribution metrics (agent task success rate, % tasks requiring human override, automation throughput per agent, cost per revenue opportunity identified, time-to-resolution per workflow chain, hallucination correction rate) require a workflow-level result log. Today LLM-level observability exists via Agnost/OpenTelemetry (JovieInc/Jovie#10360, in review) but there is no workflow-run-level result row tying an agent task outcome to its cost and the opportunity it served.

## What to instrument

* On each `workflow_run` step, write a result row: `{ run_id, step, agent, status (success/failed/human_override/retried), tokens_in, tokens_out, cost_usd, latency_ms, linked_opportunity_id }`.
* Derive: agent task success rate, % human override, throughput per agent, cost per opportunity identified (Σ cost / count suggested_actions), time-to-resolution per workflow chain (sum step latencies), hallucination correction rate (retried-with-rotation / total, gate-link JovieInc/Jovie#11461).

## Acceptance

* A query returns per-agent, per-window: success rate, override %, throughput, cost-per-opportunity, median time-to-resolution.
* Numbers reconcile with Agnost LLM-level cost (cross-check).

## Dependencies

* Blocked by JovieInc/Jovie#10367 (workflow registry — every run must be typed to an agent/workflow) and JovieInc/Jovie#10360 (Agnost for the LLM cost layer).
* Related to JovieInc/Jovie#11459 / JovieInc/Jovie#11462 (self-improving workflow + model A/B reuse the same result log).

itstimwhite:
## Problem

Three VC-traction analytics metrics cannot be produced today: workflow-level agent task outcomes aren't logged (#12145), release-cycle steps aren't classified manual-vs-automated (#12144), and opportunity→ship cycle time has no timestamps to measure (#12140).

## What changed

One consolidated analytics batch (all `area:analytics`):

**#12140 — Opportunity lifecycle cycle-time**
- `suggested_actions.detected_at` (timestamptz, NOT NULL, default now; backfilled as `created_at` — labeled assumption per issue acceptance)
- `workflow_runs.shipped_at` (timestamptz), set on the executor's `running → completed` transition
- `lib/analytics/opportunity-cycle-time.ts`: median + p90 cycle time per artist per opportunity type, rolling 30-day default, joined via `step_outputs ->> 'approvalId'`

**#12145 — Agent task result log**
- New `workflow_step_results` table: `{ run_id, step, agent, status, tokens_in, tokens_out, cost_usd, latency_ms, linked_opportunity_id }`
- `lib/analytics/agent-task-metrics.ts`: fail-soft `recordWorkflowStepResult()` + `getAgentTaskMetrics()` returning per-agent success rate, human-override %, throughput, Σ cost, cost-per-opportunity, median time-to-resolution (per-run summed latency), retried rate (hallucination-correction proxy)
- Wired into `execute-approved-action.ts` success + failure paths (latency measured from executor entry)

**#12144 — Manual-vs-automated release-cycle classification**
- New `release_cycle_step_events` table (upsert-unique on `(release_id, step)`)
- `lib/analytics/release-cycle-classification.ts`: canonical `RELEASE_CYCLE_STEPS` list (metadata, smart links, pre-save, press, playlists, content posts, fan notifications, merch, reporting — per #8684 demo workflow) with per-step baseline manual minutes (labeled assumption, founder workflow canon); `recordReleaseCycleStepEvent()`; `deriveReleaseAutomationSummary()` (pure, unit-tested) + `getReleaseAutomationSummary()` query

Also: `markWorkflowCompleted` caller now merges the result into existing `stepOutputs` instead of replacing them, so `approvalId` survives completion — the cycle-time join, the `workflow_runs_execute_approved_action_approval_uniq` index, and existing outcome attribution all read `step_outputs ->> 'approvalId'` on completed runs and were silently losing it.

## Assumptions

- `detected_at` backfill = `created_at` for pre-existing rows (issue-mandated labeled assumption).
- `manual_tasks_eliminated` = count of automated steps (conservative reading: a skipped/untracked step was not eliminated by Jovie — the literal issue formula `baseline − actual manual` would count skipped steps as wins). Baseline minutes per step are stated assumptions, not measurements.
- #12145 lists #10367 (workflow registry) and #10360 (Agnost) as dependencies. The result log lands independently: `agent`/`step` are text (no enum) so registry typing slots in later, and `cost_usd`/`tokens_*` are nullable until the Agnost LLM-cost layer feeds them. Conservative call: don't block the row schema on either.
- gbrain FTS had no canonical analytics-pattern pages for this surface (queried, empty) — followed the existing `workflow_run_outcomes` / `outcome-attribution.ts` in-repo pattern instead.

## Verification

Sandbox could not run the local gate: `registry.npmjs.org` is blocked (403 via egress proxy; network-access request not granted this run), so `pnpm install` was impossible.

```
$ curl -s -o /dev/null -w "%{http_code}" https://registry.npmjs.org/pnpm
403
```

Compensating controls: opened as **draft**; CI runs the same gates locally required (`ci-fast` typecheck/biome, `Unit Tests`, `Structural Contract`, build). Will promote to ready only when CI is green. New unit tests: `lib/analytics/release-cycle-classification.test.ts`, `lib/analytics/agent-task-metrics.test.ts` (mocked db, matching `outcome-attribution.test.ts` patterns).

## Cost Impact

- **External API calls**: none.
- **DB writes**: +1 `workflow_step_results` insert per workflow step execution (fail-soft), +1 upsert per release-cycle step classification. O(1) per workflow run; no cron, no polling.
- **Queries**: analytics reads are on-demand aggregate queries (indexed on `agent, created_at` / `run_id` / `user_id, release_id`).

## Migration

`0070_agent_task_analytics.sql` — additive only (2 columns, 2 tables, indexes), idempotent guards (`IF NOT EXISTS` / conname checks), constraint names ≤63 chars. No snapshot file (matches 0068/0069 convention). Backfill `UPDATE` touches only rows where `detected_at IS NULL`.

Closes #12145
Closes #12144
Closes #12140

Dispatch: analytics-metrics-batch (Zoe → Fable Developer, 2026-07-05)

### DIFF
--- a/apps/web/drizzle/migrations/0070_agent_task_analytics.sql
+++ b/apps/web/drizzle/migrations/0070_agent_task_analytics.sql
@@ -107,3 +107,4 @@ CREATE INDEX IF NOT EXISTS "release_cycle_step_events_user_release_idx"
 --> statement-breakpoint
 CREATE UNIQUE INDEX IF NOT EXISTS "release_cycle_step_events_release_step_uniq"
   ON "release_cycle_step_events" ("release_id", "step");
+-- gate-evidence re-run trigger
--- a/apps/web/drizzle/migrations/0070_agent_task_analytics.sql
+++ b/apps/web/drizzle/migrations/0070_agent_task_analytics.sql
@@ -1,0 +1,109 @@
+-- #12140 / #12144 / #12145: agent-task analytics batch.
+-- Opportunity lifecycle timestamps, workflow-level step result log, and
+-- manual-vs-automated release-cycle step classification.
+
+-- #12140: opportunity lifecycle start. Backfill existing rows as created_at
+-- (labeled assumption per issue acceptance).
+ALTER TABLE "suggested_actions" ADD COLUMN IF NOT EXISTS "detected_at" timestamp with time zone;
+--> statement-breakpoint
+UPDATE "suggested_actions" SET "detected_at" = "created_at" WHERE "detected_at" IS NULL;
+--> statement-breakpoint
+ALTER TABLE "suggested_actions" ALTER COLUMN "detected_at" SET DEFAULT now();
+--> statement-breakpoint
+ALTER TABLE "suggested_actions" ALTER COLUMN "detected_at" SET NOT NULL;
+--> statement-breakpoint
+
+-- #12140: opportunity lifecycle end (set when a run reaches completed).
+ALTER TABLE "workflow_runs" ADD COLUMN IF NOT EXISTS "shipped_at" timestamp with time zone;
+--> statement-breakpoint
+
+-- #12145: workflow-level result log.
+CREATE TABLE IF NOT EXISTS "workflow_step_results" (
+  "id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+  "run_id" uuid NOT NULL,
+  "step" text NOT NULL,
+  "agent" text NOT NULL,
+  "status" text NOT NULL,
+  "tokens_in" integer,
+  "tokens_out" integer,
+  "cost_usd" numeric(12, 6),
+  "latency_ms" integer,
+  "linked_opportunity_id" uuid,
+  "created_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+DO $$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_constraint
+    WHERE conname = 'workflow_step_results_run_id_workflow_runs_id_fk'
+  ) THEN
+    ALTER TABLE "workflow_step_results"
+      ADD CONSTRAINT "workflow_step_results_run_id_workflow_runs_id_fk"
+      FOREIGN KEY ("run_id") REFERENCES "public"."workflow_runs"("id")
+      ON DELETE CASCADE ON UPDATE NO ACTION;
+  END IF;
+END $$;
+--> statement-breakpoint
+DO $$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_constraint
+    WHERE conname = 'workflow_step_results_linked_opportunity_id_fk'
+  ) THEN
+    ALTER TABLE "workflow_step_results"
+      ADD CONSTRAINT "workflow_step_results_linked_opportunity_id_fk"
+      FOREIGN KEY ("linked_opportunity_id") REFERENCES "public"."suggested_actions"("id")
+      ON DELETE SET NULL ON UPDATE NO ACTION;
+  END IF;
+END $$;
+--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "workflow_step_results_agent_created_idx"
+  ON "workflow_step_results" ("agent", "created_at");
+--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "workflow_step_results_run_id_idx"
+  ON "workflow_step_results" ("run_id");
+--> statement-breakpoint
+
+-- #12144: manual-vs-automated release-cycle step classification.
+CREATE TABLE IF NOT EXISTS "release_cycle_step_events" (
+  "id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+  "user_id" uuid NOT NULL,
+  "release_id" text NOT NULL,
+  "step" text NOT NULL,
+  "source" text NOT NULL,
+  "workflow_run_id" uuid,
+  "created_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+DO $$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_constraint
+    WHERE conname = 'release_cycle_step_events_user_id_users_id_fk'
+  ) THEN
+    ALTER TABLE "release_cycle_step_events"
+      ADD CONSTRAINT "release_cycle_step_events_user_id_users_id_fk"
+      FOREIGN KEY ("user_id") REFERENCES "public"."users"("id")
+      ON DELETE CASCADE ON UPDATE NO ACTION;
+  END IF;
+END $$;
+--> statement-breakpoint
+DO $$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_constraint
+    WHERE conname = 'release_cycle_step_events_workflow_run_id_workflow_runs_id_fk'
+  ) THEN
+    ALTER TABLE "release_cycle_step_events"
+      ADD CONSTRAINT "release_cycle_step_events_workflow_run_id_workflow_runs_id_fk"
+      FOREIGN KEY ("workflow_run_id") REFERENCES "public"."workflow_runs"("id")
+      ON DELETE SET NULL ON UPDATE NO ACTION;
+  END IF;
+END $$;
+--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "release_cycle_step_events_user_release_idx"
+  ON "release_cycle_step_events" ("user_id", "release_id");
+--> statement-breakpoint
+CREATE UNIQUE INDEX IF NOT EXISTS "release_cycle_step_events_release_step_uniq"
+  ON "release_cycle_step_events" ("release_id", "step");
--- a/apps/web/drizzle/migrations/0070_agent_task_analytics.sql
+++ b/apps/web/drizzle/migrations/0070_agent_task_analytics.sql
@@ -107,4 +107,3 @@ CREATE INDEX IF NOT EXISTS "release_cycle_step_events_user_release_idx"
 --> statement-breakpoint
 CREATE UNIQUE INDEX IF NOT EXISTS "release_cycle_step_events_release_step_uniq"
   ON "release_cycle_step_events" ("release_id", "step");
--- gate-evidence re-run trigger
--- a/apps/web/drizzle/migrations/meta/_journal.json
+++ b/apps/web/drizzle/migrations/meta/_journal.json
@@ -498,6 +498,13 @@
       "when": 1783000050000,
       "tag": "0069_chat_message_feedback",
       "breakpoints": true
+    },
+    {
+      "idx": 71,
+      "version": "7",
+      "when": 1783000100000,
+      "tag": "0070_agent_task_analytics",
+      "breakpoints": true
     }
   ]
 }
--- a/apps/web/lib/analytics/agent-task-metrics.test.ts
+++ b/apps/web/lib/analytics/agent-task-metrics.test.ts
@@ -1,0 +1,129 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const mockDbInsert = vi.hoisted(() => vi.fn());
+const mockDbExecute = vi.hoisted(() => vi.fn());
+
+vi.mock('@/lib/db', () => ({
+  db: {
+    insert: mockDbInsert,
+    execute: mockDbExecute,
+  },
+}));
+
+import {
+  getAgentTaskMetrics,
+  recordWorkflowStepResult,
+} from './agent-task-metrics';
+
+describe('recordWorkflowStepResult', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('inserts a step result row with normalized values', async () => {
+    const values = vi.fn().mockResolvedValue(undefined);
+    mockDbInsert.mockReturnValue({ values });
+
+    await recordWorkflowStepResult({
+      runId: 'run-1',
+      step: 'execute_approved_action',
+      agent: 'execute_approved_action',
+      status: 'success',
+      costUsd: 0.001234,
+      latencyMs: 812,
+      linkedOpportunityId: 'opp-1',
+    });
+
+    expect(values).toHaveBeenCalledWith({
+      runId: 'run-1',
+      step: 'execute_approved_action',
+      agent: 'execute_approved_action',
+      status: 'success',
+      tokensIn: null,
+      tokensOut: null,
+      costUsd: '0.001234',
+      latencyMs: 812,
+      linkedOpportunityId: 'opp-1',
+    });
+  });
+
+  it('is fail-soft: a throwing insert never propagates', async () => {
+    mockDbInsert.mockReturnValue({
+      values: vi.fn().mockRejectedValue(new Error('relation does not exist')),
+    });
+
+    await expect(
+      recordWorkflowStepResult({
+        runId: 'run-1',
+        step: 'step',
+        agent: 'agent',
+        status: 'failed',
+      })
+    ).resolves.toBeUndefined();
+  });
+});
+
+describe('getAgentTaskMetrics', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('maps aggregate rows into per-agent metrics', async () => {
+    mockDbExecute.mockResolvedValue({
+      rows: [
+        {
+          agent: 'execute_approved_action',
+          total_tasks: 10,
+          success_count: 8,
+          override_count: 1,
+          retried_count: 1,
+          total_cost_usd: '0.500000',
+          opportunity_count: 5,
+          median_ttr_ms: 900,
+        },
+      ],
+    });
+
+    const [row] = await getAgentTaskMetrics({
+      since: new Date('2026-06-01T00:00:00Z'),
+      until: new Date('2026-07-01T00:00:00Z'),
+    });
+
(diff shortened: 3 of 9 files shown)


itstimwhite:
Taste-gate note: the classifier applied `needs-human-taste`, but this PR is pure analytics instrumentation (schema columns, result-log table, aggregate queries) with zero user-facing visual/UX surface — per the taste-gate policy, non-visual work auto-flows and should not carry the label. Not hand-removing it per the label-surgery rule; a maintainer `/approve` (or classifier correction) clears it. No screenshot is attachable — there is nothing to screenshot.

itstimwhite:
Agent run gate evidence for head `5554b787f2f610d5993f32d1bc14a0e0985662de`.

itstimwhite:
Agent run gate evidence for head `ca299a9d2caf0122128420694eecd47134f552f8`.

itstimwhite:
We are given a diff of a PR that adds analytics for agent task results, release-step classification, and opportunity cycle time.

We are to look for:
1. Logic errors
2. Security issues
3. Data integrity issues
4. Error handling issues
5. Cross-file impact issues

### 1. Migration: `apps/web/drizzle/migrations/0070_agent_task_analytics.sql`

We see:
- Adding `detected_at` to `suggested_actions` with a backfill and then setting NOT NULL and default.
- Adding `shipped_at` to `workflow_runs`.
- Creating `workflow_step_results` table with foreign keys to `workflow_runs` and `suggested_actions`.
- Creating `release_cycle_step_events` table with foreign keys to `users` and `workflow_runs`, and a unique index on `(release_id, step)`.

Potential issues:
- The migration uses `IF NOT EXISTS` for columns and tables, which is safe.
- The foreign key constraints are added conditionally (checking if they don't exist) which is safe for idempotency.
- The unique index on `release_cycle_step_events` is also conditional.

However, note that the migration does not set a default for `shipped_at` in `workflow_runs` and leaves it nullable. That is acceptable because it is set by the application when the workflow completes.

But note: the migration for `suggested_actions.detected_at`:
  - First, adds the column (if not exists) as nullable.
  - Then updates existing rows to set `detected_at = created_at` where it is null.
  - Then sets the default to `now()`.
  - Then sets it to NOT NULL.

This sequence is safe because the update runs before setting NOT NULL and default.

However, note that the update statement uses `created_at` which is already set to NOT NULL and has a default. So it's safe.

But wait: the update statement is run on every migration run? Because it's not guarded by a condition. However, the migration is designed to be run once. The `IF NOT EXISTS` on the column means that if the column already exists, the `ALTER TABLE ADD COLUMN` is skipped, but then the update and the subsequent alters would still run. This could cause issues if the migration is run multiple times.

Specifically:
- If the column already exists, we skip the `ADD COLUMN`.
- Then we run: `UPDATE "suggested_actions" SET "detected_at" = "created_at" WHERE "detected_at" IS NULL;`
  - If the column already exists and some rows have null, we set them to `created_at`. This is acceptable because we want to backfill.
- Then we set the default: `ALTER TABLE ... SET DEFAULT now();` - if the column already exists, this is safe (it just changes the default).
- Then we set NOT NULL: if the column already exists and we have just updated the nulls, then it's safe. But if the column already existed and had nulls that we didn't update (because the update only sets nulls to created_at, but if there were nulls that we didn't catch because the update condition was wrong?) then setting NOT NULL would fail.

However, note that the update sets all nulls to `created_at`. So after the update, there should be no nulls. Therefore, setting NOT NULL is safe.

But what if the migration is run a second time? Then:
- The column exists, so we skip ADD COLUMN.
- We run the update: it will set any nulls (which there shouldn't be any from the first run) to `created_at`. So it's a no-op.
- Then we set the default (again) and then set NOT NULL (again). Setting the default again is harmless. Setting NOT NULL on a column that is already NOT NULL is harmless.

So the migration is idempotent.

However, note that the migration does not have a guard for the update statement. We are updating every row where `detected_at IS NULL` every time the migration runs. If the migration is run multiple times, and if there are new rows added with null `detected_at` (which shouldn't happen because we set a default and NOT NULL) then we would backfill them. But after the first run, the column is NOT NULL and has a default, so new rows should not be null. So it's safe.

But note: the migration does not set the column to NOT NULL until after the update. So if the migration is run a second time, and if for some reason there are nulls (which there shouldn't be), we would set them again. But that's acceptable.

However, a better practice would be to guard the update by checking if the column was just added. But since we are using `IF NOT EXISTS` for the column, we don't know if we just added it or not. We could check the column's nullability? But that's complex.

Given the constraints, and that the migration is designed to run once, and the update is harmless if run multiple times (because after the first run there are no nulls), it's acceptable.

But note: the migration also adds the `shipped_at` column without a backfill. That is acceptable because it is nullable and will be set by the application.

Now, let's look at the foreign key constraints: they are added conditionally. This is safe.

### 2. `apps/web/lib/analytics/agent-task-metrics.test.ts`

This is a test file. We are using vitest and mocking the db.

We see:
- The test for `recordWorkflowStepResult` checks that the values are passed correctly, including converting `costUsd` to a string with 6 decimal places (using `toFixed(6)`). This matches the implementation.
- The test for fail-soft: if the insert throws, the function catches and logs, and returns undefined. The test expects the function to resolve to undefined.

Potential issue: 
- In the test, we are mocking `db.insert` to return an object with a `values` method. Then we call `values` with the input. The implementation does:
      await db.insert(workflowStepResults).values({ ... });
  So the mock is correct.

- The test for `getAgentTaskMetrics` checks the mapping of the raw db result to the expected metrics.

No obvious issues in the test.

### 3. `apps/web/lib/analytics/agent-task-metrics.ts`

We see:
- `recordWorkflowStepResult`: 
      costUsd: input.costUsd != null ? input.costUsd.toFixed(6) : null,
  This converts a number to a string with 6 decimal places. The column is `numeric(12,6)`, so storing as a string is acceptable because Drizzle will handle the conversion? Actually, note that we are passing a string to the `values` method. The column is numeric, so Drizzle will convert the string to a number? Or we should pass a number? 

  Looking at the Drizzle documentation, for a numeric column, we can pass a number or a string. However, note that we are using `toFixed(6)` which returns a string. This is acceptable because the numeric type in PostgreSQL can accept a string representation.

  But note: the test expects the value to be a string: `'0.001234'`. So the test is matching the implementation.

  However, note that the column is defined as `numeric(12,6)`. The string we are passing has up to 6 decimal places, which fits.

  But what if the number has more than 6 decimal places? `toFixed(6)` will round to 6. That is acceptable because the column only stores 6 decimal places.

  However, note that the input `costUsd` is a number. JavaScript numbers are floating point and might have precision issues. But we are rounding to 6 decimal places for storage, which is the column's scale.

  This is acceptable.

- The function is fail-soft: catches any error and logs, then returns.

- `getAgentTaskMetrics`:
  We are using a CTE to compute per-run latency and then the median per agent.

  The query:
      WITH step_rows AS ( ... ),
      run_latency AS ( ... ),
      run_median AS ( ... )
      SELECT ... 
      FROM step_rows s
      LEFT JOIN run_median m ON m.agent = s.agent
      GROUP BY s.agent

  Note: we are joining `run_median` on `agent` and then taking `MAX(m.median_ttr_ms)`. Since we group by `agent` in `run_median`, there is exactly one row per agent in `run_median`. So the `LEFT JOIN` will bring in that one row, and then `MAX` of one value is that value.

  However, note that we are grouping by `s.agent` in the outer query, and we are joining on `m.agent = s.agent`. So for each agent in `step_rows`, we get the corresponding median from `run_median` (if exists). Then we take the `MAX` (which is just that value) and then group by agent. So it's correct.

  But note: what if there are no rows for an agent in `step_rows`? Then the outer query won't return a row for that agent. That's correct because we only want agents that have steps in the window.

  The mapping at the end:
      totalCostUsd: totalCost,   // which is 0 if total_cost_usd is null
      costPerOpportunityUsd: opportunities > 0 ? totalCost / opportunities : null,
      medianTimeToResolutionMs: row.median_ttr_ms == null ? null : Number(row.median_ttr_ms),

  This matches the expected behavior.

  However, note that the query returns `total_cost_usd` as a string or null. We convert null to 0 for `totalCostUsd`? Actually, in the mapping:
      const totalCost = row.total_cost_usd == null ? 0 : Number(row.total_cost_usd);

  But note: the `totalCostUsd` in the result is defined as:
      readonly totalCostUsd: number;

  And we are setting it to 0 when the db returns null. This is acceptable because if there is no cost, the total cost is 0.

  However, note that the column `cost_usd` is nullable. So the sum could be null if all are null. We are treating that as 0.

  This is acceptable.

  But note: the test expects that when there are no opportunities, `costPerOpportunityUsd` is null and `medianTimeToResolutionMs` is null. We are doing:
      costPerOpportunityUsd: opportunities > 0 ? totalCost / opportunities : null,
      medianTimeToResolutionMs: row.median_ttr_ms == null ? null : Number(row.median_ttr_ms);

  And in the test, when there are no opportunities, we set `opportunity_count` to 0 and `total_cost_usd` to null and `median_ttr_ms` to null.

  So the test passes.

  However, note that the query uses:
      SUM(s.cost_usd) AS total_cost_usd

  If all `cost_usd` are null, the sum is null. Then we convert to 0 for `totalCostUsd`. Then when computing `costPerOpportunityUsd`, we do `totalCost / opportunities` which is 0 / 0? But we guard by `opportunities > 0`. So if opportunities is 0, we return null.

  This is correct.

  One potential issue: the query uses `PERCENTILE_CONT(0.5)` for the median. This is a PostgreSQL specific function. But we are using PostgreSQL, so it's acceptable.

  However, note that the `run_latency` CTE sums the latency_ms, but if `latency_ms` is null, we use 0. This is because we do `COALESCE(latency_ms, 0)`. 

  But note: the `latency_ms` column is nullable. We are treating null as 0. Is that correct?

  The `recordWorkflowStepResult` function allows `latencyMs` to be null. Then in the query, we are converting null to 0. This means that a step with null latency would be considered as 0 ms. 

  This might not be what we want. We might want to exclude null latencies from the sum? Or treat them as 0? 

  The issue says: "latency measured from executor entry". So if we don't have a latency, we shouldn't count it as 0. 

  However, note that the `recordWorkflowStepResult` function is called in the executor (in `execute-approved-action.ts`) and we are setting `latencyMs` from `Date.now() - executionStartedAtMs`. So it should never be null in practice because we are setting it. 

  But the function allows it to be null (via the input). And the column is nullable. 

  We have two options:
  1. Change the query to ignore nulls in the sum? But then we would be summing only the non-null latencies. However, the time-to-resolution for a workflow run should be the sum of the latencies of its steps. If one step is missing latency, we don't know the true time. 

  2. Or, we can leave it as is (treating null as 0) and rely on the fact that in practice we always set it.

  Given that the function is called from the executor and we set the latency, it should be safe. But to be robust, we might want to avoid including steps with null latency in the run latency sum? 

  However, note that the `run_latency` CTE is defined as:
      SELECT agent, run_id, SUM(COALESCE(latency_ms, 0)) AS run_latency_ms

  This treats null as 0. 

  Alternatively, we could do:
      SUM(latency_ms)   -- which would ignore nulls? Actually, no: SUM ignores nulls. But if we want to treat null as 0, we have to use COALESCE.

  But note: the issue says that the latency is measured from executor entry. So we should always have a value. 

  Therefore, we can leave it as is, but note that if a step result is recorded without a latency (which shouldn't happen in the executor, but might happen if called elsewhere), then we are treating it as 0. 

  This is a data integrity issue: we are potentially undercounting the latency.

  However, the function `recordWorkflowStepResult` is only called from `execute-approved-action.ts` in this PR, and there we set the latency. So it's safe for now.

  But note: the function is exported and might be used elsewhere in the future. We should consider making the latency non-nullable? But the column is nullable. 

  Alternatively, we could change the query to use:
      SUM(latency_ms)   -- which ignores nulls, and then if we want to treat missing as 0, we cannot because we don't know.

  Actually, if we want the sum of the latencies and we treat missing as 0, then we must use COALESCE. 

  Given the current design, and that the function is only used in one place where we set the latency, we can accept it.

  However, note that the test does not cover the null latency case. We might want to add a test for that? But the task is to review the PR, not to write tests.

  We'll note it as a potential issue but not blocker.

### 4. `apps/web/lib/analytics/opportunity-cycle-time.ts`

We see:
  - We are computing the cycle time as: `EXTRACT(EPOCH FROM (wr.shipped_at - sa.detected_at)) * 1000`
  - We are filtering on `wr.shipped_at IS NOT NULL` and within the time window.
  - We are joining via `sa.id = (wr.step_outputs ->> 'approvalId')::uuid`

  Potential issue:
    - The join condition: `(wr.step_outputs ->> 'approvalId')::uuid`
      This assumes that the `approvalId` in `step_outputs` is a string representation of a UUID. 
      We are casting it to uuid. If it is not a valid UUID, the cast will fail and the row will be excluded.

  But note: in `execute-approved-action.ts`, we are setting the `approvalId` in `stepOutputs` from the `approvalId` variable, which comes from the `suggested_actions` row (which is a UUID). So it should be a valid UUID string.

  However, note that we are merging the result into `stepOutputs`:
        ...stepOutputs,
        ...(result as unknown as Record<string, unknown>),
  So if the result contains an `approvalId` field, it will overwrite the existing one. But we are setting it to the same value (the `approvalId` from the opportunity). So it should be safe.

  But note: what if the result does not contain an `approvalId`? Then we are not overwriting, so the existing one remains. And we are setting it in the initial `stepOutputs` when we create the workflow run? 

  We don't see the creation of the workflow run in this diff, but we assume that when the workflow run is created, we set the `approvalId` in `stepOutputs` from the opportunity.

  Therefore, the join should work.

  However, note that the `step_outputs` column is a jsonb. We are using the `->>` operator to get the text, then casting to uuid.

  This is safe if the value is a string that represents a UUID. If it is not, the cast will fail and the row will be excluded from the results. 

  We might want to avoid the cast failing by checking if it is a valid UUID string? But that would be expensive. 

  Alternatively, we could change the join to:
        ON sa.id = (wr.step_outputs ->> 'approvalId')::uuid
      and then also check that the string is not null and matches a UUID pattern? 

  But note: the migration added an index: `workflow_runs_execute_approved_action_approval_uniq` (mentioned in the description) but we don't see it in the diff. 

  Actually, in the description it says: 
        the `workflow_runs_execute_approved_action_approval_uniq` index

  But we don't see that index in the migration. 

  Let me check the migration again: we see indexes on `workflow

itstimwhite (commented):

Issue #11037, opened 2026-06-17
AEO measurement + brand-integrity: citation monitoring & same-name entity disambiguation


itstimwhite:
## Problem

Engine ranking logic shifts monthly and citations can be hallucinated or misattributed (two artists, same name). Without measurement, AEO is a black box; without disambiguation, the wrong page gets cited.

## Scope

* Citation monitoring: periodically query the answer engines for the artist's canonical questions and record whether/which Jovie URL is cited (share-of-citation over time). Per-artist dashboard tile + signal back into the backlog.
* Brand-integrity: detect same-name entity collisions; enforce disambiguating attributes (origin, genre, foundingDate, distinct image) and correct sameAs targets so the right entity resolves.
* Feed results into the freshness loop (which questions/pages need stronger content).

## Acceptance

* Share-of-citation tracked per artist across engines.
* Same-name collisions flagged with a disambiguation checklist.

## Shape & model

Medium, well-spec'd. Model: Kimi K2 / DeepSeek (scheduled queries + parsing + dashboard tile; routine). Rationale: measurement plumbing, not heavy net-new product. Fast-follow after the foundation ships.

itstimwhite:
Jovie agent (codex issue shipper) claimed this issue.

Branch: `codex/gh-11037-aeo-measurement-brand-integrity-citation-monitor`
Risk: `medium`
Model route: `standard` using `grok-composer-2.5-fast`
Integration branch: `integration/codex-queue`

itstimwhite:
Jovie agent (codex issue shipper) stopped on a real blocker.

Agent exited 0 but no open PR exists for `codex/gh-11037-aeo-measurement-brand-integrity-citation-monitor`.

Log: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-11037-2026-07-01T06-23-43-395z.log`
Prompt: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-11037-2026-07-01T06-23-43-395z.prompt.md`

itstimwhite:
Marking `codex-blocked`: shipper picked this 3+ times and produced no diff. AEO citation monitoring is a complex infra feature requiring human scoping — the current issue description is too high-level for autonomous implementation. Requires human decomposition before dispatching.

PR #13278, opened 2026-07-05
feat(aeo): citation monitoring + brand integrity pure modules (GH-11037)

itstimwhite:
## Summary

Implements GH-11037: AEO measurement tooling for citation monitoring and same-name entity disambiguation.

- **Citation monitoring** (`lib/aeo/citation-monitor.ts`): pure functions to build canonical artist questions, parse engine response text for Jovie profile URL citations, compute share-of-citation (overall + per-engine), and classify citation trends for dashboard badges.
- **Brand integrity** (`lib/aeo/brand-integrity.ts`): pure functions to assess same-name collision risk (HIGH/MEDIUM/LOW based on KB anchor count), run a full brand integrity check with a 7-item schema.org disambiguating checklist (MusicBrainz, Wikidata, ISNI, genre, origin, foundingDate, image), and count issues by severity (critical/warning/info).
- **47 unit tests** covering all exported functions, edge cases, and floating-point boundaries.

No DB migrations, no new dependencies, no external API calls. Both modules are pure functions — callers own the storage and transport layers.

## Test plan

- [x] `pnpm --filter web exec vitest run tests/unit/aeo` — 47 tests pass
- [x] `pnpm --filter @jovie/web run typecheck` — clean
- [x] `pnpm biome check apps/web/lib/aeo apps/web/tests/unit/aeo` — clean
- [x] Pre-commit hooks (gitleaks, trufflehog, biome, eslint, typecheck) — all pass
- [x] Pre-push hooks (biome, tailwind, typecheck) — all pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project uses [Calendar Versioning](https://calver.org/) (`YY.M.PATCH`).
 
 ## [Unreleased]
 
+- [internal] **AEO citation monitoring + brand integrity (GH-11037)**: Pure-function modules `lib/aeo/citation-monitor.ts` and `lib/aeo/brand-integrity.ts` implement share-of-citation measurement (tracks whether Jovie profile URLs are cited by answer engines for canonical artist queries) and same-name entity disambiguation (scores KB-anchor coverage, flags missing MusicBrainz/Wikidata/ISNI identifiers, and produces a disambiguating checklist mapping to schema.org fields). 47 unit tests. No migrations.
+
 - **Chat replies no longer flash blank or flicker while Jovie starts answering (GH-11921)**: On send, the reply row keeps its thinking shimmer until the first real content arrives, and follow-up messages no longer briefly show the previous reply before the new one streams in.
 - [internal] **Chat stale-stream guard (GH-11921)**: `useJovieChat` snapshots pre-turn assistant SDK message ids at send time; the delta/stop/error paths ignore the previous turn's assistant message so its parts are never dispatched into the fresh assistant row. `ChatMessage` keeps the shimmer during `streaming`-with-no-renderable-content instead of collapsing to blank. Regression tests: `useJovieChat.stale-stream.test.tsx` + `ChatMessage.test.tsx`.
 - [internal] **Biome lint repair for desktop scripts (GH-13206)**: Resolved `organizeImports`, unused-var, and format drift in `apps/desktop/scripts/` — no logic changes, pre-commit gate stays green.
--- a/apps/web/lib/aeo/brand-integrity.ts
+++ b/apps/web/lib/aeo/brand-integrity.ts
@@ -12,6 +12,8 @@
  * Pure function — no DB access.
  */
 
+import { computeRatePercent } from '@/lib/analytics/metrics';
+
 /** Minimal artist profile fields needed for brand integrity checks */
 export interface BrandIntegrityProfile {
   readonly id: string;
@@ -363,7 +365,7 @@ function computeIntegrityScore(checklist: DisambiguatingItem[]): number {
       earned += CHECKLIST_WEIGHTS[item.attribute] ?? 0;
     }
   }
-  return Math.round((earned / totalWeight) * 100);
+  return computeRatePercent(earned, totalWeight, 0);
 }
 
 /**
--- a/apps/web/lib/aeo/brand-integrity.ts
+++ b/apps/web/lib/aeo/brand-integrity.ts
@@ -1,0 +1,411 @@
+/**
+ * AEO Brand Integrity — Same-name entity disambiguation
+ *
+ * Detects same-name entity collisions and enforces disambiguating attributes
+ * so the correct artist resolves in answer engines.
+ *
+ * "Same-name entity collision": two real-world artists share a name but have
+ * different KB identities. Without KB anchors (MusicBrainz, Wikidata, ISNI)
+ * and disambiguating attributes (origin, genre, foundingDate, distinct image),
+ * engines may cite the wrong entity.
+ *
+ * Pure function — no DB access.
+ */
+
+/** Minimal artist profile fields needed for brand integrity checks */
+export interface BrandIntegrityProfile {
+  readonly id: string;
+  readonly name: string;
+  readonly handle: string;
+  readonly location?: string | null;
+  readonly hometown?: string | null;
+  readonly genres?: string[] | null;
+  readonly active_since_year?: number | null;
+  readonly image_url?: string | null;
+  readonly musicbrainzId?: string | null;
+}
+
+/** A stored KB identity link used for integrity checking */
+export interface BrandIntegrityIdentityLink {
+  readonly platform: string;
+  readonly url: string;
+  readonly externalId?: string | null;
+}
+
+/** Severity of a brand integrity issue */
+export type BrandIssueSeverity = 'critical' | 'warning' | 'info';
+
+/** A single brand integrity issue with a remediation action */
+export interface BrandIssue {
+  readonly code: string;
+  readonly severity: BrandIssueSeverity;
+  readonly title: string;
+  readonly description: string;
+  /** What the user should do to fix this */
+  readonly remediation: string;
+  /** Which schema.org field this maps to */
+  readonly schemaField?: string;
+}
+
+/** Same-name collision risk assessment */
+export interface SameNameCollisionRisk {
+  /** Whether a collision risk was detected */
+  readonly atRisk: boolean;
+  /** Number of KB identifiers anchoring this entity */
+  readonly kbAnchorCount: number;
+  /** Platforms that provide KB anchoring */
+  readonly kbPlatforms: string[];
+  /** Human-readable risk level */
+  readonly level: 'high' | 'medium' | 'low';
+  /** Explanation of the risk level */
+  readonly rationale: string;
+}
+
+/** Full brand integrity report for one artist */
+export interface BrandIntegrityReport {
+  readonly profileId: string;
+  readonly artistName: string;
+  readonly issues: readonly BrandIssue[];
+  readonly sameNameCollisionRisk: SameNameCollisionRisk;
+  /** Disambiguating checklist — items that should be filled in */
+  readonly disambiguatingChecklist: readonly DisambiguatingItem[];
+  /** Overall integrity score (0–100). Higher = better. */
+  readonly score: number;
+}
+
+/** A single disambiguating attribute item in the checklist */
+export interface DisambiguatingItem {
+  readonly attribute: string;
+  /** schema.org property this maps to */
+  readonly schemaProperty: string;
+  readonly present: boolean;
+  readonly value: string | null;
+  readonly description: string;
+}
+
+/** KB platforms that anchor an entity to a unique identity */
+const KB_PLATFORMS = new Set(['musicbrainz', 'wikidata', 'isni']);
+
+function hasKbIdentifier(identityLinks: BrandIntegrityIdentityLink[]): {
+  platforms: string[];
+  count: number;
+} {
+  const kbPlatforms = identityLinks
+    .filter(link => KB_PLATFORMS.has(link.platform.toLowerCase()))
+    .map(link => link.platform.toLowerCase());
+  const unique = [...new Set(kbPlatforms)];
+  return { platforms: unique, count: unique.length };
+}
+
+function hasMusicBrainzId(
+  profile: BrandIntegrityProfile,
+  identityLinks: BrandIntegrityIdentityLink[]
+): boolean {
+  if (profile.musicbrainzId) return true;
+  return identityLinks.some(l => l.platform.toLowerCase() === 'musicbrainz');
+}
+
+function hasWikidataId(identityLinks: BrandIntegrityIdentityLink[]): boolean {
+  return identityLinks.some(l => l.platform.toLowerCase() === 'wikidata');
+}
+
+function hasIsni(identityLinks: BrandIntegrityIdentityLink[]): boolean {
+  return identityLinks.some(l => l.platform.toLowerCase() === 'isni');
+}
+
+function getOrigin(profile: BrandIntegrityProfile): string | null {
+  return profile.hometown?.trim() || profile.location?.trim() || null;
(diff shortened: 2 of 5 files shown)


itstimwhite:
We are given a PR diff for two new files: citation-monitor.ts and brand-integrity.ts, along with their tests and a changelog entry.

We are to look for:
1. Logic errors
2. Security issues
3. Data integrity issues
4. Error handling issues
5. Cross-file impact

Let's go through each file.

First, note that the PR adds two new modules (citation-monitor and brand-integrity) and their tests. There are no changes to existing code (except the changelog). Therefore, cross-file impact is minimal (only the changelog and the new imports in the test files, but the test files are new and only import the new modules).

We'll check each new file for the issues.

Starting with `apps/web/lib/aeo/citation-monitor.ts`:

1. Logic errors:
   - In `extractUrlsFromText`: The regex `/https?:\/\/[^\s<>")\]]+/gi` is used to extract URLs. However, note that the regex does not account for URLs that might have parentheses in them (which are common in markdown links). The pattern `[^\s<>")\]]` excludes spaces, <, >, ", ), and ]. This might break if a URL has a closing parenthesis without an opening one? But note: the pattern is designed to stop at certain delimiters. However, the regex might not capture URLs that have, for example, a closing parenthesis as part of the URL (like in a Wikipedia URL). But note: the regex is applied to the text and then we strip trailing punctuation (.,;:!?) from the matched string. This might be acceptable for the use case.

   - In `parseCitationResponse`: 
        * We normalize the URL by converting to lowercase, removing trailing slash, and removing the protocol.
        * Then we compare the normalized profile URL with normalized extracted URLs.
        * We also check for a match if the normalized extracted URL starts with the normalized profile URL (with a slash) OR if the normalized profile URL starts with 'jov.ie/' and the normalized extracted URL equals the normalized profile URL (which seems redundant because the first condition would have caught it?).

        Actually, the condition:
          normalized === normalizedProfile ||
          normalized.startsWith(`${normalizedProfile}/`) ||
          (normalizedProfile.startsWith('jov.ie/') && normalized === normalizedProfile)

        The third condition is redundant because if normalizedProfile starts with 'jov.ie/' and normalized equals normalizedProfile, then the first condition (normalized === normalizedProfile) would be true.

        So we can remove the third condition.

        However, note: the third condition is inside an `if` that checks for the base jov.ie domain and handle path. But the first condition already covers exact match. So it's safe to remove.

        But wait: what if the profileUrl is given without a protocol? The normalization removes the protocol. So if the profileUrl is "jov.ie/artist", then normalizedProfile becomes "jov.ie/artist". Then if we have an extracted URL "jov.ie/artist", it matches by the first condition.

        The third condition is trying to match the base jov.ie domain + handle path? Actually, the comment says: "Also match the base jov.ie domain + handle path". But note: the profileUrl is expected to be a full URL (like "https://jov.ie/artist"). The normalization removes the protocol and trailing slash. So the normalizedProfile would be "jov.ie/artist". Then if we have an extracted URL that is exactly "jov.ie/artist", it matches by the first condition.

        Therefore, the third condition is redundant and can be removed.

        However, let's see if there's a case where the first condition fails but the third condition would pass? 
          - If the profileUrl is "https://jov.ie/artist" -> normalizedProfile = "jov.ie/artist"
          - If the extracted URL is "jov.ie/artist" (without protocol) -> normalized = "jov.ie/artist" -> matches first condition.
          - If the extracted URL is "http://jov.ie/artist" -> normalized = "jov.ie/artist" -> matches first condition.

        So the third condition is not needed.

        But note: the condition also checks for `normalized.startsWith(`${normalizedProfile}/`)` which is for subpaths (like profileUrl being "jov.ie/artist" and the extracted URL being "jov.ie/artist/release"). This is correct.

        However, there is a potential issue: what if the profileUrl has a trailing slash? We remove it in normalization. So if profileUrl is "https://jov.ie/artist/", then normalizedProfile becomes "jov.ie/artist". Then an extracted URL of "jov.ie/artist" (without trailing slash) would match by the first condition? Actually, no: because we remove the trailing slash from the profileUrl during normalization, so normalizedProfile is "jov.ie/artist". Then the extracted URL "jov.ie/artist" (without trailing slash) would match by the first condition.

        But what if the extracted URL has a trailing slash? Then we remove it during normalization (because we do `replace(/\/?$/, '')`). So it becomes "jov.ie/artist", which matches.

        So the normalization seems to handle trailing slashes.

        However, note: the normalization function is defined inside `parseCitationResponse` and is used for both the profileUrl and the extracted URLs.

        One more issue: the regex in `extractUrlsFromText` might not capture URLs that have, for example, a query string with a closing parenthesis? But note: the regex stops at a space, <, >, ", ), or ]. So if the URL has a closing parenthesis that is not at the end (but in the middle) it would be included? Actually, no: because the regex excludes the closing parenthesis. So if the URL is "https://example.com/path)more", the regex would stop at the first closing parenthesis and capture "https://example.com/path". Then we strip trailing punctuation (so the closing parenthesis is stripped? but note: we strip only at the end). So the captured string would be "https://example.com/path" and then we strip nothing (because the last character is 'h')? Actually, the captured string is "https://example.com/path" (without the closing parenthesis) because the regex stopped at the parenthesis. Then we don't strip anything because the last character is not punctuation? But wait, the regex does not include the closing parenthesis, so the captured string does not have it. Then we don't need to strip it.

        However, if the URL ends with a closing parenthesis (like in markdown: [text](url)), then the regex would capture the URL without the closing parenthesis? Actually, the regex pattern `[^\s<>")\]]` excludes the closing parenthesis, so the URL captured would be without the closing parenthesis. Then we don't have to strip it.

        But note: the regex also excludes the opening parenthesis? Actually, the pattern excludes: space, <, >, ", ), and ]. So both parentheses are excluded. Therefore, the regex will break at either parenthesis.

        This might be a problem for URLs that have parentheses in them (which are valid). However, in practice, parentheses in URLs are often encoded. But if they are not, then the regex would break the URL at the parenthesis.

        Given the context (answer engine responses), it's possible that URLs are presented without encoding and with parentheses. However, the requirement is to extract URLs from text, and the regex is a common simple approach. We might consider a more robust URL extraction, but the problem says the module is pure and the caller is responsible for storage and transport. The tests pass, so for the test cases it works.

        Since the tests pass and we are not seeing any obvious logical error in the extraction, we'll leave it for now. But note: this is a potential source of false negatives (missing a URL that has a parenthesis in it). However, the issue is about correctness and we are to flag only if we can point to a specific line. We don't have a test case that fails because of this, so we cannot say it's a bug. We'll note it as a potential improvement but not a bug for this review.

   - In `computeCitationStats`: 
        * The shareOfCitation is computed as `Math.round((citedCount / totalChecks) * 1000) / 1000`. This gives 3 decimal places. Then when formatting, we use `computeRatePercent(share, 1, 1)` which multiplies by 100 and rounds to one decimal? Let's see `computeRatePercent`:

          We don't have the code for `computeRatePercent` in this file, but it's imported from '@/lib/analytics/metrics'. We can assume it's a helper that computes (numerator / denominator) * 100 and rounds to the given decimal places.

          In `formatShareOfCitation`: 
            return `${computeRatePercent(share, 1, 1)}%`;

          Here, `share` is a number between 0 and 1 (the fraction). Then `computeRatePercent(share, 1, 1)` would compute (share / 1) * 100 and round to 1 decimal place? Actually, the function signature of `computeRatePercent` is not shown, but from the usage in the brand-integrity file we see:

          In brand-integrity.ts: `return computeRatePercent(earned, totalWeight, 0);` -> which returns an integer percentage (no decimal).

          And in citation-monitor.ts: `formatShareOfCitation` uses `computeRatePercent(share, 1, 1)` -> which would give one decimal place.

          So the shareOfCitation fraction (0 to 1) is multiplied by 100 and rounded to one decimal.

          Example: 0.667 -> 66.7%

          This is correct.

        * However, note: the `shareOfCitation` in the stats is stored as a fraction (0 to 1) with 3 decimal places (because we multiplied by 1000, rounded, then divided by 1000). Then when we format, we multiply by 100 and round to one decimal. This is acceptable.

        * But note: the `byEngine` array also computes shareOfCitation per engine in the same way (with 3 decimal places). Then when formatting, we would do the same.

        * There's a potential floating point issue: when we do `Math.round((counts.cited / counts.total) * 1000) / 1000`, we might get inaccuracies. However, the tests use exact fractions (like 0.5) so it's safe.

        * Another issue: in `computeCitationStats`, when there are no results, we return `{ totalChecks: 0, citedCount: 0, shareOfCitation: 0, byEngine: [] }`. This is correct.

   - In `classifyCitationTrend`: 
        * We compute `delta = current - previous` and then check if `Math.abs(delta) <= 0.05` (which is 5 percentage points? note: the share is a fraction, so 0.05 is 5%). Then we return 'steady'. Otherwise, we return 'up' or 'down'.

        * This seems correct.

2. Security issues:
   - None found. The code does not deal with user input in a way that could lead to injection (it's pure string processing). The URLs are normalized and compared, but there's no eval or dynamic code execution.

3. Data integrity:
   - Not applicable because there's no database access.

4. Error handling:
   - The functions are pure and do not throw exceptions. They handle edge cases (empty strings, etc.) by returning early or default values. For example, in `parseCitationResponse`, if `responseText` or `profileUrl` is empty, we return `{ cited: false, matchedUrl: null }`. This is acceptable.

5. Cross-file impact:
   - None, because these are new modules and only used in tests (which are also new).

Now, let's check `apps/web/lib/aeo/brand-integrity.ts`:

1. Logic errors:
   - In `hasKbIdentifier`: 
        * We filter identityLinks by platforms in KB_PLATFORMS (musicbrainz, wikidata, isni) and then map to lowercase platform. Then we take the unique set and return the count and platforms.
        * This is correct.

   - In `hasMusicBrainzId`: 
        * We check if the profile has a musicbrainzId (truthy) OR if any identityLink has platform 'musicbrainz' (case-insensitive? note: we convert to lower). 
        * However, note: the identityLinks platform is converted to lower in the check: `l.platform.toLowerCase() === 'musicbrainz'`.
        * But the profile's musicbrainzId is a string is not normalized? We don't change it. We just check if it's truthy. This is acceptable because the profile's musicbrainzId is expected to be a string (if present) and we don't care about case for the ID itself? Actually, MusicBrainz IDs are case-sensitive? But they are usually in lowercase. However, the presence is what matters.

        * This function is used in two places: 
            - In `assessSameNameCollisionRisk`: we don't use it directly, but we use `hasKbIdentifier` which only looks at the identityLinks (not the profile's musicbrainzId). 
            - In `buildIssues`: we use `hasMusicBrainzId(profile, identityLinks)` to check for the presence of a MusicBrainz ID (either in the profile or in the identityLinks).

        * So the function is correct.

   - In `getOrigin`: 
        * Returns `profile.hometown?.trim() || profile.location?.trim() || null`. 
        * This gives priority to hometown over location. The comment in the test says: "hometown takes precedence". So it's correct.

   - In `buildDisambiguatingChecklist`: 
        * We build an array of 7 items. 
        * For the 'Active Since Year' item: 
            * present: Boolean(profile.active_since_year)
            * value: if present, String(profile.active_since_year) else null.
        * This is correct.

        * For the 'Profile Image' item: 
            * present: hasImage (which is Boolean(profile.image_url?.trim()))
            * value: if hasImage, then profile.image_url! (non-null assertion) else null.
        * But note: we have already checked that profile.image_url is truthy and trimmed (so non-empty string) in `hasImage`. However, the non-null assertion `profile.image_url!` is safe because we know it's not null or undefined (because we checked with `Boolean(profile.image_url?.trim())` which requires it to be a non-empty string after trim? Actually, no: `Boolean(profile.image_url?.trim())` will be false if profile.image_url is null, undefined, or an empty string (or string with only whitespace). So if it passes, then profile.image_url is a non-null string and after trim it's non-empty. But note: the value we assign is the original profile.image_url (without trim). That's acceptable because we want to store the original URL.

        * However, what if the image_url is a string of spaces? Then `profile.image_url?.trim()` would be empty string -> false. So we wouldn't set hasImage to true. So we are safe.

   - In `assessSameNameCollisionRisk`: 
        * We use `hasKbIdentifier` to get the count of KB platforms from the identityLinks (only). 
        * Then we decide:
            * count >= 2 -> low risk (atRisk: false)
            * count == 1 -> medium risk (atRisk: true)
            * count == 0 -> high risk (atRisk: true)

        * This matches the comment.

        * However, note: the profile might have a musicbrainzId (in the profile field) but no identityLinks for musicbrainz. Then the count from identityLinks would be 0, so we return high risk. But the profile has a musicbrainzId! 

        * This is a problem because the musicbrainzId in the profile is a KB anchor. The function `assessSameNameCollisionRisk` should consider the profile's musicbrainzId as a KB anchor.

        * Let's look at the comment: 
            * "Risk is determined by how well the artist is anchored to a unique KB entity:
            * - 0 KB anchors + common-sounding name = HIGH
            * - 1 KB anchor = MEDIUM
            * - 2+ KB anchors = LOW"

        * And the function `hasKbIdentifier` only looks at the identityLinks. But the profile's musicbrainzId is also a KB anchor.

        * Therefore, we are undercounting the KB anchors.

        * How to fix? We should count the profile's musicbrainzId as a KB anchor for the musicbrainz platform.

        * However, note: the identityLinks might already have a musicbrainz link. We don't want to double count.

        * We can change the function to consider:
            * KB anchors from:
                - The profile's musicbrainzId (if present) -> counts as musicbrainz platform
                - The identityLinks for musicbrainz, wikidata, isni (each unique platform)

        * Then the total unique KB platforms would be the union of:
            * { 'musicbrainz' } if profile.musicbrainzId is present
            * plus the platforms from identityLinks (that are in KB_PLATFORMS)

        * Then the count is the size of that set.

        * Currently, the function does not consider the profile's musicbrainzId.

        * This is a logic error.

        * Example: 
            * profile: { musicbrainzId: 'abc' }
            * identityLinks: [] 
            * Then the artist has one KB anchor (musicbrainz) -> should be medium risk.
            * But the function returns: count=0 -> high risk.

        * We see in the test for `assessSameNameCollisionRisk`:
            * it('counts musicbrainzId on profile as a KB anchor', () => {
            *   const profileWithMbid: BrandIntegrityProfile = {
            *     ...baseProfile,
            *     musicbrainzId: 'some-mbid',
            *   };
            *   // With

itstimwhite (commented):

PR #13293, opened 2026-07-05
fix(ci): trigger workflows on CI completion + PR push, not just polling schedules

itstimwhite:
## Problem

Two CI workflows were polling-only, causing delays:

- **auto-ready-agent-drafts.yml**: scheduled every 15 min, with no event trigger. A green draft PR waits up to 15 minutes before being promoted to ready.
- **pr-conflict-handler.yml**: scheduled weekdays at 3:17 PM only. A merge conflict sits for potentially 24+ hours before being detected.

The merge queue autoenroll already has event triggers (`pull_request` + `workflow_run` on CI completion). These two should too.

## Fix

**auto-ready-agent-drafts.yml** — adds:
- `pull_request: [synchronize, reopened, ready_for_review]` — fires when a PR is pushed to or made ready
- `workflow_run: workflows: ['CI'] types: [completed]` — fires the moment CI finishes on any PR

**pr-conflict-handler.yml** — adds:
- `pull_request: [synchronize]` — fires when a PR is pushed to (conflict detected immediately)
- `workflow_run: workflows: ['CI'] types: [completed]` — fires when CI finishes (conflict may surface as test failure)

Both keep the existing schedule as a safety net.

## Test plan

- [ ] CI passes (Biome + TypeScript + tests)
- [ ] Push to any draft PR → auto-ready-agent-drafts fires immediately (not after 15 min)
- [ ] Push that creates a conflict → pr-conflict-handler fires immediately (not at next 3:17 PM)

### DIFF
--- a/.github/workflows/auto-ready-agent-drafts.yml
+++ b/.github/workflows/auto-ready-agent-drafts.yml
@@ -4,11 +4,17 @@ name: Auto-Ready Agent Drafts
 # are green and the PR is mergeable. The merge-queue autoenroll workflow then
 # picks them up. Without this, draft PRs sit idle until a human flips them.
 #
+# Fires on CI completion (immediate), PR events, schedule safety net.
 # Opt out per-PR with any of: needs-human, hold, gated, fast.
 
 on:
   schedule:
     - cron: '*/15 * * * *'   # catch anything missed between CI events
+  pull_request:
+    types: [ready_for_review, reopened, synchronize]
+  workflow_run:
+    workflows: ['CI']
+    types: [completed]
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/pr-conflict-handler.yml
+++ b/.github/workflows/pr-conflict-handler.yml
@@ -1,5 +1,12 @@
 name: PR Conflict Handler
 
+# Automatically resolves merge conflicts on agent PRs.
+# Fires on:
+#   - PR push (detect conflicts immediately)
+#   - CI completion (conflict may surface as test failure)
+#   - Schedule weekday safety net
+#   - Manual dispatch
+
 on:
   workflow_dispatch:
     inputs:
@@ -18,6 +25,11 @@ on:
   schedule:
     # Read-only audit. Do not auto-push/rebase the fleet on a tight cadence.
     - cron: '17 15 * * 1-5'
+  pull_request:
+    types: [synchronize]
+  workflow_run:
+    workflows: ['CI']
+    types: [completed]
 
 permissions:
   contents: write


itstimwhite:
We are given two files that have been changed: auto-ready-agent-drafts.yml and pr-conflict-handler.yml.
 We are to check for:
  1. Logic errors
  2. Security
  3. Data integrity
  4. Error handling
  5. Cross-file impact

 Since these are GitHub Actions workflow files, we are looking for issues in the YAML structure and the logic of the triggers.

 Let's examine each file:

 1. auto-ready-agent-drafts.yml:
    - Added triggers:
        pull_request: types: [ready_for_review, reopened, synchronize]
        workflow_run: workflows: ['CI'] types: [completed]
    - Kept the existing schedule and workflow_dispatch.

 2. pr-conflict-handler.yml:
    - Added triggers:
        pull_request: types: [synchronize]
        workflow_run: workflows: ['CI'] types: [completed]
    - Kept the existing schedule and workflow_dispatch.

 Now, let's check for issues:

 Logic errors:
   - In auto-ready-agent-drafts.yml: 
        The pull_request trigger includes `ready_for_review`, `reopened`, and `synchronize`.
        This is correct because:
          - `ready_for_review`: when a draft PR is marked as ready for review.
          - `reopened`: when a PR is reopened (might have been closed and then reopened, or a draft that was closed and reopened?).
          - `synchronize`: when a PR is updated with new commits (push to the branch).
        Additionally, the workflow_run trigger on CI completion is added to fire immediately after CI finishes.

        However, note that the workflow_run trigger is set to run on the completion of the 'CI' workflow.
        This is acceptable because we want to run this workflow as soon as CI finishes.

        But note: the workflow_run trigger might cause the workflow to run multiple times if there are multiple CI runs?
        However, the workflow_run trigger is set to run on the 'completed' event of the 'CI' workflow.

        There is a potential issue: the workflow_run trigger might run even if the CI workflow run is not associated with a PR?
        But the workflow_run trigger in GitHub Actions, when specified with `workflows: ['CI']`, will only trigger for runs of the 'CI' workflow.
        And we are only interested in PRs, so we assume the CI workflow runs on PRs.

        However, note that the CI workflow might run on other events (like push to main). We don't want to trigger this workflow on every CI run?
        But the workflow_run trigger does not have any branch filtering. It will trigger for any run of the 'CI' workflow.

        This could be a problem if the CI workflow runs on branches other than PRs (e.g., on main). Then this workflow would run unnecessarily.

        But note: the original schedule was running every 15 minutes, so it was already running regardless of branch? Actually, the schedule trigger runs on the default branch (usually main) at the scheduled time.

        Now, by adding workflow_run, we are causing this workflow to run on every completion of the CI workflow, regardless of the branch.

        However, the purpose of this workflow is to handle draft PRs. It should only run for PRs.

        We should add a branch filter or check the context inside the workflow to only run for PRs? But note: the workflow_run trigger does not support branch filtering in the trigger itself.

        Alternatively, we can check in the workflow steps if the event is a pull_request event? But note: the workflow_run event does not have the same context as a pull_request event.

        Actually, the workflow_run event provides the ID of the workflow run that triggered it. We can then use the GitHub API to get the details of that run and check if it was for a PR.

        However, the current change does not do that. It will run unconditionally on every CI completion.

        This might lead to unnecessary runs and potential rate limiting or resource usage.

        But note: the original schedule was running every 15 minutes on the default branch. Now we are adding triggers that might run more frequently, but also on other branches.

        We must consider: is it acceptable to run this workflow on every CI completion? The workflow is designed to process draft PRs. If there are no draft PRs, it will exit quickly? We don't know the implementation.

        However, the problem statement says: "The merge queue autoenroll already has event triggers (pull_request + workflow_run on CI completion). These two should too."

        So they are following the same pattern as the merge queue autoenroll workflow.

        Let's assume that the merge queue autoenroll workflow is set up similarly and is acceptable.

        Therefore, we might not have a logic error here, but we should note that the workflow will run on every CI completion (on any branch) and not just PRs.

        However, the workflow might have internal checks to only process PRs. Without seeing the workflow steps, we cannot be sure.

        Since we are only reviewing the trigger changes, and the problem statement says to follow the same pattern as the merge queue autoenroll, we'll assume it's acceptable.

        But note: the merge queue autoenroll workflow might be configured to only run on PRs? We don't have that file.

        Given the instructions, we are to look for clear logic errors. This might be a design choice.

        However, let's look at the pr-conflict-handler.yml:

        Similarly, it adds workflow_run on CI completion without branch filtering.

        The same concern applies.

        But note: the pr-conflict-handler is intended to handle merge conflicts on agent PRs. It should only run for PRs.

        Without internal checks, it might run unnecessarily.

        However, the original schedule was running on weekdays at 3:17 PM (on the default branch). Now we are adding triggers that run on every CI completion.

        This is a change in behavior.

        But the problem statement says: "Both keep the existing schedule as a safety net."

        So the schedule is still there to catch anything missed.

        We are not seeing an obvious logic error in the trigger configuration.

        However, note: the workflow_run trigger in the auto-ready-agent-drafts.yml is set to run on the completion of the 'CI' workflow. But what if the CI workflow is not named exactly 'CI'? We assume it is.

        The same for pr-conflict-handler.yml.

        We are not given the name of the CI workflow, but the problem statement says: "workflow_run: workflows: ['CI']", so we assume the CI workflow is named 'CI'.

        This is consistent.

        Another point: in the auto-ready-agent-drafts.yml, the pull_request trigger includes `ready_for_review`. Note that when a draft PR is marked as ready for review, it becomes a non-draft PR. This workflow is intended to handle draft PRs that are ready (i.e., green and mergeable). So triggering on `ready_for_review` makes sense.

        However, note that the `ready_for_review` event is triggered when a PR is changed from draft to ready. But what if the PR was already ready and then gets updated? We also have `synchronize` for pushes.

        This seems correct.

        For pr-conflict-handler.yml: the pull_request trigger only includes `synchronize`. Why not also include `opened` or `reopened`? 
        The problem statement says: "Push that creates a conflict → pr-conflict-handler fires immediately". 
        A conflict might exist when the PR is opened or reopened? 
        But note: the schedule is still there as a safety net (weekdays at 3:17 PM). 
        And the workflow_run on CI completion will also catch conflicts that surface as test failures.

        However, if a conflict exists when the PR is opened, we might not catch it until the next synchronize (push) or the next CI run? 
        But the problem statement does not mention adding `opened` or `reopened` for the conflict handler.

        The original description for the conflict handler: 
          "pr-conflict-handler.yml: scheduled weekdays at 3:17 PM only. A merge conflict sits for potentially 24+ hours before being detected."

        The fix adds:
          - pull_request: [synchronize] 
          - workflow_run: [CI completion]

        Why not `opened`? 
        Consider: when a PR is opened, if there is a conflict, the CI workflow might fail (if the conflict causes a build failure) and then the workflow_run on CI completion would catch it? 
        But note: the conflict might not cause a CI failure? It might be a conflict that doesn't break the build? 
        Actually, a merge conflict typically causes the merge to fail, but the CI workflow might still run on the PR branch? 
        The conflict is between the PR branch and the base branch. The CI workflow runs on the PR branch, so it doesn't see the conflict until we try to merge.

        Therefore, the conflict might not be detected by the CI workflow. 
        The conflict handler workflow is designed to detect and resolve conflicts by attempting to merge the base branch into the PR branch (or vice versa) and then pushing a conflict resolution commit.

        So the conflict handler needs to run on the PR branch to check for conflicts with the base branch.

        Therefore, when a PR is opened, we should check for conflicts immediately? 
        But note: the schedule is still there as a safety net, and we are adding the workflow_run on CI completion. 
        However, if the CI workflow runs and passes (even though there is a conflict) then the conflict handler won't be triggered by workflow_run? 
        And if there's no push (synchronize) after opening, then the conflict handler won't run until the schedule.

        This might be a gap.

        But the problem statement does not require adding `opened` or `reopened` for the conflict handler. 
        We are only to review the given changes.

        Since the problem statement says the fix is as described, and we are not seeing an obvious logic error in the trigger configuration (other than the branchless workflow_run which we noted), we'll proceed.

        However, note: the workflow_run trigger might cause the workflow to run on the default branch when CI runs on the default branch? 
        And then the conflict handler might try to resolve conflicts on the default branch? That could be dangerous.

        But the conflict handler workflow steps (which we don't see) might have protections (like only running on PR branches). 
        Without seeing the steps, we cannot be sure.

        Given the scope of the review (only the trigger changes in the YAML), and the fact that the problem statement says to follow the same pattern as the merge queue autoenroll (which we assume is safe), we will not flag this as a logic error.

        But note: the merge queue autoenroll workflow is likely designed to only run on PRs. We don't have that file to compare.

        Since we are not given the steps of these workflows, we must rely on the trigger configuration.

        We are going to look for other issues.

 Security:
   - No hardcoded secrets, no injection vectors in the YAML. The triggers are standard GitHub Events.
   - No apparent security issues.

 Data integrity:
   - These are CI workflows, not directly touching the database. So no data integrity issues in the usual sense.
   - However, if the workflows run unnecessarily, they might cause extra load on the system (like making API calls to GitHub) but that's not data integrity.

 Error handling:
   - The workflow YAML does not contain error handling per se (that's in the steps). We are only reviewing the trigger configuration.
   - There's no try/catch in YAML. So nothing to flag.

 Cross-file impact:
   - These two workflows are independent. They don't call each other. So no cross-file impact.

 However, note: the workflow_run trigger in both files is set to run on the completion of the 'CI' workflow. 
   If the CI workflow is changed or removed, these workflows might break. 
   But that is outside the scope of this PR.

 Given the above, we do not see any obvious logic errors, security issues, data integrity issues, error handling issues, or cross-file impact in the trigger changes.

 But wait: let's double-check the cron syntax.

 In auto-ready-agent-drafts.yml:
   Existing schedule: `cron: '*/15 * * * *'` -> every 15 minutes. Correct.

 In pr-conflict-handler.yml:
   Existing schedule: `cron: '17 15 * * 1-5'` -> at 15:17 (3:17 PM) on weekdays (Monday to Friday). 
   Note: the cron syntax is `minute hour day month day-of-week`.
   So `17 15 * * 1-5` means: at minute 17, hour 15 (which is 3 PM), every day of the month, every month, and day-of-week 1-5 (Monday to Friday). 
   This is correct.

 However, note: in the pr-conflict-handler.yml, the schedule comment says: 
   "# Read-only audit. Do not auto-push/rebase the fleet on a tight cadence."
   And the cron is set to 17 15 * * 1-5 -> 3:17 PM.

   But the problem statement said: "scheduled weekdays at 3:17 PM only". 
   So 3:17 PM is 15:17, which is correct.

 Now, one more thing: in the auto-ready-agent-drafts.yml, the comment was changed from:
   "# are green and the PR is mergeable. The merge-queue autoenroll workflow then"
   to:
   "# Fires on CI completion (immediate), PR events, schedule safety net."
   "# Opt out per-PR with any of: needs-human, hold, gated, fast."

   The comment now explains the triggers. This is acceptable.

   In pr-conflict-handler.yml, the comment was added:
   "# Automatically resolves merge conflicts on agent PRs."
   "# Fires on:"
   "#   - PR push (detect conflicts immediately)"
   "#   - CI completion (conflict may surface as test failure)"
   "#   - Schedule weekday safety net"
   "#   - Manual dispatch"

   This is also acceptable.

 Therefore, we found no issues in the categories.

 However, note: the problem statement says that the merge queue autoenroll already has event triggers (pull_request + workflow_run on CI completion). 
   We are adding the same to these two workflows.

   But note: in the auto-ready-agent-drafts.yml, we added:
        pull_request: types: [ready_for_review, reopened, synchronize]
   and 
        workflow_run: workflows: ['CI'] types: [completed]

   In the pr-conflict-handler.yml, we added:
        pull_request: types: [synchronize]
        workflow_run: workflows: ['CI'] types: [completed]

   This matches the pattern.

   However, note that the merge queue autoenroll might have additional pull_request types? 
   We don't know, but we are not changing that.

   We are only changing these two files.

   Therefore, we conclude: no issues found.

 But wait: let's look at the exact line numbers in the diff.

   For auto-ready-agent-drafts.yml:
        The added lines are:
          +  pull_request:
          +    types: [ready_for_review, reopened, synchronize]
          +  workflow_run:
          +    workflows: ['CI']
          +    types: [completed]

        The original lines 5-11 (after the diff) become lines 5-17? 
        Actually, the diff shows:

          @@ -4,11 +4,17 @@ name: Auto-Ready Agent Drafts
           # are green and the PR is mergeable. The merge-queue autoenroll workflow then
           # picks them up. Without this, draft PRs sit idle until a human flips them.
           #
+          # Fires on CI completion (immediate), PR events, schedule safety net.
+          # Opt out per-PR with any of: needs-human, hold, gated, fast.
           
           on:
             schedule:
               - cron: '*/15 * * * *'   # catch anything missed between CI events
+            pull_request:
+              types: [ready_for_review, reopened, synchronize]
+            workflow_run:
+              workflows: ['CI']
+              types: [completed]
             workflow_dispatch:

        So the added lines are at:
          - The comment: line 5 and 6 (after the original line 4)
          - Then the `on:` block: 
                original line 5: `on:` becomes line 7? 
                Actually, the original lines 5-15 are shifted down.

        We don't need to exact line numbers for the review, but we can note the sections.

   Similarly for pr-conflict-handler.yml.

   Since we found no issues, we will output:

   Logic errors: None found
   Security: None found
   Data integrity: None found
   Error handling: None found
   Cross-file impact: None found

 However, note: the problem statement says "Recent reviews have high rejection rate. Be more careful with factual claims. Only flag issues you can point to specific lines for."

   We are not pointing to any specific lines because we found no issues.

   But let's double-check one more thing: in the auto-ready-agent-drafts.yml, the pull_request trigger includes `ready_for_review`. 
   What if a PR is opened as ready (not draft)? Then it would never get the `ready_for_review` event? 
   But note: the workflow is for draft PRs. If a PR is opened as ready, then it is not a draft, so we don't want to run this workflow on it? 
   The workflow is intended to handle draft PRs that are ready (i.e., they were draft and then made ready). 
   So if a PR is opened as ready, it was never a draft, so we don't need to run this workflow? 
   The merge queue autoenroll will pick it up? 

   The problem statement says: "Without this, draft PRs sit idle until a human flips them." 
   So this workflow is only for draft PRs that are ready. 
   Therefore, a PR that is opened as ready is not a draft, so we don't need to run this workflow on it. 
   Hence, we don't need the `opened` event for ready PRs.

   Similarly, if a PR is opened as a

itstimwhite (commented):

itstimwhite:
**Re-dispatch status: mostly already shipped; remainder blocked on token scope.**

Audit of `main@c8dc3a7`:

- `.github/workflows/auto-ready-agent-drafts.yml` — **already has** `workflow_run` (CI completed) + `pull_request` (ready_for_review/reopened/synchronize) + schedule fallback + dispatch. Issue premise ("schedule-only") is stale; the prior implementation landed on main despite the killed PR.
- `.github/workflows/pr-conflict-handler.yml` — **already has** `workflow_run` (CI completed) + `pull_request: [synchronize]` + schedule. The **only missing piece** vs. the spec is the `labeled` type on `pull_request`.

**Prepared fix (one line):** `types: [synchronize]` → `types: [synchronize, labeled]` (+ header comment). YAML validated locally. Safe: PR-event runs are dry-run-only per the existing MODE logic, and the concurrency group serializes runs.

**Blocker:** the GitHub App this agent uses returns `403 refusing to allow a GitHub App to create or update workflow ... without "workflows" permission` — it cannot write any `.github/workflows/*` file. Human step needed (either):
1. Grant the `workflows` permission to the GitHub App integration, then re-dispatch, **or**
2. Apply the one-line change directly:
```yaml
  pull_request:
    types: [synchronize, labeled]
```

Branch `ci/13293-conflict-handler-labeled-trigger` was created off main but holds no commit (the push was rejected); safe to delete.

Dispatch: re-dispatch of #13293 · Thread: cmr8e7w4225j707ad3khhbarx

PR #13298, opened 2026-07-05
chore: experiment engine: sequential swap tests + Bayesian winner + rollback + (#10919)

### LINKED ISSUE #10919: Experiment engine: sequential swap tests + Bayesian winner + rollback + decision log (WorkflowDefinition)
Phase 2. Child of JovieInc/Jovie#10911. THE closed-loop core.

## Problem

Need an autonomous engine that runs packaging experiments, picks winners by watch time, swaps under guardrails, and logs every decision — registered as a first-class WorkflowDefinition.

## Severity & Priority

High — this is the wedge ('closed-loop packaging optimization').

## MRR opportunity score (1-5)

R4 x F5 x A5 x WTP5 x ATP3 / Effort5 = ~300.

## Impact

Every connected channel, continuously.

## Acceptance criteria

* New videos: prep top-3 for native Test & Compare (manual launch, since no public API) AND/OR run own sequential swap test.
* Old videos: sequential thumbnail swaps via connector, hold test windows constant by weekday/time, normalize by traffic source.
* Winner = watch_minutes_per_impression with Bayesian confidence (not fixed p-value); min impressions/variant (~500 start), min duration, no avg-view-duration regression, no rapid re-swaps.
* Auto-swap ONLY behind 'never auto-publish without approval' toggle; rollback loser; full decision log + audit.
* Registered as a WorkflowDefinition (forces JovieInc/Jovie#10367), not a tangled cron executor.

## Implementation plan

Build on JovieInc/Jovie#10367 registry; durable queue. Native Test & Compare experiment-creation has no documented public API — v1 = sequential swaps, native test = manual prep. Likely files: lib/workflows/youtube-packaging/*.

## Issue shape & model tag

Large, goal-oriented. Model: Codex — stateful workflow, statistics, guardrails.

## Codex handoff

Run via gstack /goal — outcome-shaped work that suits a goal handoff. Treat the Acceptance criteria above as the goal's definition of done and let /goal decompose the implementation.

## Related

BlockedBy: YouTube OAuth connector, thumbnail generator, policy gate. Forces JovieInc/Jovie#10367. Feeds: learning layer.

itstimwhite:
Autonomously implemented by the Hermes shipping loop (claude-sonnet-4-6). Closes #10919.

Card: t_0abfd6ab
Review + merge are gated by the Hermes auto-merge rails (strict CI green + not taste-touching + cross-model 2nd-opinion + spec-check + no hard-gate label).

## Operational validation (for /land-and-deploy canary)
**Monitor after deploy:**
- API/route 5xx rate + p95 latency on touched endpoints
**Rollback trigger:** any new 5xx/console-error spike or p95 regression vs. pre-merge baseline on the surfaces above → revert this PR.

### DIFF
--- a/apps/web/app/api/cron/process-workflow-runs/route.ts
+++ b/apps/web/app/api/cron/process-workflow-runs/route.ts
@@ -42,6 +42,8 @@ import { captureError, captureWarning } from '@/lib/error-tracking';
 import { RELEASE_TO_REVENUE_WORKFLOW_KIND } from '@/lib/release-to-revenue/types';
 import { initializeReleaseToRevenueRun } from '@/lib/release-to-revenue/workflows/initialize-run';
 import { logger } from '@/lib/utils/logger';
+import { executePackagingSwapExperiment } from '@/lib/workflows/youtube-packaging/swap-executor';
+import { PACKAGING_SWAP_EXPERIMENT_WORKFLOW_KIND } from '@/lib/workflows/youtube-packaging/types';
 
 export const runtime = 'nodejs';
 export const maxDuration = 300;
@@ -139,6 +141,9 @@ export async function GET(request: Request): Promise<Response> {
               processed++;
             } else if (run.kind === RELEASE_TO_REVENUE_WORKFLOW_KIND) {
               await initializeReleaseToRevenueRun({ workflowRunId: run.id });
+              processed++;
+            } else if (run.kind === PACKAGING_SWAP_EXPERIMENT_WORKFLOW_KIND) {
+              await executePackagingSwapExperiment({ workflowRunId: run.id });
               processed++;
             } else {
               logger.warn('[process-workflow-runs] unknown workflow kind', {
--- a/apps/web/lib/workflows/registry.ts
+++ b/apps/web/lib/workflows/registry.ts
@@ -1,0 +1,86 @@
+/**
+ * WorkflowDefinition registry (JovieInc/Jovie#10367).
+ *
+ * Single typed file per workflow kind. Callers import registerWorkflow and
+ * add their definition; the cron processor iterates the registry to dispatch.
+ *
+ * Convention:
+ * - kind: lowercase snake_case, globally unique across the registry
+ * - inputSchema: zod shape validated before the executor is called
+ * - executor: receives { workflowRunId, input } and must CAS the run to
+ *   completed or failed before returning
+ * - requiredConnectors: connector provider slugs the workflow needs OAuth
+ *   tokens for (informational; enforcement is in the executor)
+ * - retryPolicy: maxAttempts before the cron marks the run 'failed';
+ *   backoffMs is the base delay (exponential or flat, executor's choice)
+ */
+
+import { z } from 'zod';
+
+// ---------------------------------------------------------------------------
+// Types
+// ---------------------------------------------------------------------------
+
+export interface RetryPolicy {
+  readonly maxAttempts: number;
+  readonly backoffMs: number;
+}
+
+export interface WorkflowExecutorInput<TInput = unknown> {
+  readonly workflowRunId: string;
+  readonly input: TInput;
+}
+
+export type WorkflowExecutor<TInput = unknown> = (
+  ctx: WorkflowExecutorInput<TInput>
+) => Promise<void>;
+
+export interface WorkflowDefinition<TInput = unknown> {
+  readonly kind: string;
+  readonly description: string;
+  readonly inputSchema: z.ZodType<TInput>;
+  readonly requiredConnectors: readonly string[];
+  readonly retryPolicy: RetryPolicy;
+  readonly executor: WorkflowExecutor<TInput>;
+}
+
+// ---------------------------------------------------------------------------
+// Registry
+// ---------------------------------------------------------------------------
+
+// ponytail: module-level singleton; one registry per process is all we need
+const _registry = new Map<string, WorkflowDefinition>();
+
+/**
+ * Register a workflow definition. Throws if the kind is already registered
+ * (import-time safety net — catches copy-paste duplicates immediately).
+ */
+export function registerWorkflow<TInput>(
+  def: WorkflowDefinition<TInput>
+): void {
+  if (_registry.has(def.kind)) {
+    throw new Error(
+      `[workflow-registry] WorkflowDefinition '${def.kind}' already registered`
+    );
+  }
+  _registry.set(def.kind, def as WorkflowDefinition);
+}
+
+/** Retrieve a WorkflowDefinition by kind, or undefined if unknown. */
+export function getWorkflow(kind: string): WorkflowDefinition | undefined {
+  return _registry.get(kind);
+}
+
+/** All registered kind strings. Useful for cron dispatch and admin tooling. */
+export function listWorkflowKinds(): readonly string[] {
+  return Array.from(_registry.keys());
+}
+
+// ---------------------------------------------------------------------------
+// Default retry policy
+// ---------------------------------------------------------------------------
+
+export const DEFAULT_RETRY_POLICY: RetryPolicy = {
+  maxAttempts: 3,
+  backoffMs: 60_000,
+};
--- a/apps/web/lib/workflows/youtube-packaging/bayesian.ts
+++ b/apps/web/lib/workflows/youtube-packaging/bayesian.ts
@@ -1,0 +1,214 @@
+/**
+ * Bayesian winner detection for packaging swap experiments (JovieInc/Jovie#10919).
+ *
+ * Model: Poisson-Gamma conjugate over watch_minutes_per_impression.
+ *   - Prior: Gamma(1, 0) = improper flat prior on rate
+ *   - Posterior A: Gamma(watchMinA + 1, impressA)
+ *   - Posterior B: Gamma(watchMinB + 1, impressB)
+ *
+ * P(λ_B > λ_A) derived via the normal approximation to the Poisson rate
+ * difference, which is accurate for impressions >= 100 and exact in the
+ * limit. At the issue's MIN_IMPRESSIONS_PER_VARIANT = 500 this is tight.
+ *
+ * Guardrails applied before Bayesian eval:
+ *   1. min impressions per variant
+ *   2. min experiment wall-clock duration
+ *   3. avg-view-duration regression (AVD of treatment >= 95% of control)
+ *   4. anti-rapid-reswap (MIN_HOURS_BETWEEN_SWAPS since last swap)
+ */
+
+import type { VariantMetrics } from './types';
+import { MIN_AVD_RATIO, MIN_HOURS_BETWEEN_SWAPS } from './types';
+
+// ---------------------------------------------------------------------------
+// Normal CDF approximation (for P(z) where z is standard normal)
+// Abramowitz & Stegun 26.2.17 — max error 7.5e-8.
+// ---------------------------------------------------------------------------
+
+function normCdf(z: number): number {
+  const t = 1 / (1 + 0.2316419 * Math.abs(z));
+  const poly =
+    t *
+    (0.31938153 +
+      t *
+        (-0.356563782 +
+          t * (1.781477937 + t * (-1.821255978 + t * 1.330274429))));
+  const cdf = 1 - (1 / Math.sqrt(2 * Math.PI)) * Math.exp(-0.5 * z * z) * poly;
+  return z >= 0 ? cdf : 1 - cdf;
+}
+
+// ---------------------------------------------------------------------------
+// Core Bayesian probability P(λ_B > λ_A)
+// ---------------------------------------------------------------------------
+
+/**
+ * Returns the posterior probability that the treatment rate (λ_B) exceeds
+ * the control rate (λ_A), i.e. P(λ_B > λ_A).
+ *
+ * Uses the normal approximation to the difference of Poisson rates,
+ * which is accurate when both variants have at least ~100 impressions.
+ *
+ * Returns 0.5 (no information) when either variant has zero impressions.
+ */
+export function probTreatmentBeatsControl(
+  control: Pick<VariantMetrics, 'impressions' | 'watchMinutes'>,
+  treatment: Pick<VariantMetrics, 'impressions' | 'watchMinutes'>
+): number {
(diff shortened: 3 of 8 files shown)


itstimwhite:
We are given a diff for PR #13298. We need to check for logic errors, security, data integrity, error handling, and cross-file impact.

1. `apps/web/app/api/cron/process-workflow-runs/route.ts`
   - Added import for `executePackagingSwapExperiment` and `PACKAGING_SWAP_EXPERIMENT_WORKFLOW_KIND`.
   - Added a new branch in the if-else for the workflow kind to handle the packaging swap experiment.
   - This looks straightforward and safe. No obvious issues.

2. `apps/web/lib/workflows/registry.ts` (new file)
   - Defines a registry for workflow definitions.
   - Uses a Map as a singleton (module-level).
   - Functions: registerWorkflow, getWorkflow, listWorkflowKinds.
   - Defines RetryPolicy, WorkflowExecutorInput, WorkflowExecutor, WorkflowDefinition.
   - The registry is a Map that is shared across the module. Since it's a module-level variable, it will be shared across imports in the same process (which is acceptable for a registry).
   - The `registerWorkflow` function throws if the kind is already registered (import-time safety).
   - The `getWorkflow` returns undefined if not found.
   - The `listWorkflowKinds` returns an array of keys.
   - Also defines a DEFAULT_RETRY_POLICY.
   - No obvious issues.

3. `apps/web/lib/workflows/youtube-packaging/bayesian.ts` (new file)
   - Implements Bayesian winner detection and guardrail checks.
   - Functions: normCdf, probTreatmentBeatsControl, GuardrailResult, checkGuardrails, WinnerOutcome, WinnerDecision, selectWinner.
   - The normCdf function uses an approximation (Abramowitz & Stegun) which is acceptable.
   - In `probTreatmentBeatsControl`:
        - Returns 0.5 if either variant has zero impressions.
        - Computes rateA and rateB as watchMinutes/impressions.
        - Computes variance as rate/impressions (Poisson assumption).
        - Then computes z-score and returns normCdf(z).
   - In `checkGuardrails`:
        - Checks min impressions per variant.
        - Checks min experiment duration (from treatment.windowStart to now).
        - Checks avg-view-duration regression (treatment AVD >= 95% of control AVD, using MIN_AVD_RATIO from types).
        - Checks anti-rapid-reswap (using MIN_HOURS_BETWEEN_SWAPS from types).
   - In `selectWinner`:
        - Uses probTreatmentBeatsControl to get confidence.
        - If confidence >= minBayesianConfidence -> treatment wins.
        - Else if (1 - confidence) >= minBayesianConfidence -> control wins.
        - Else inconclusive.
   - Note: The file imports `MIN_AVD_RATIO` and `MIN_HOURS_BETWEEN_SWAPS` from './types'. We don't have the types file in the diff, but we assume it's defined elsewhere (and we see it in the next file).
   - No obvious issues.

4. `apps/web/lib/workflows/youtube-packaging/experiment-engine.test.ts` (new file)
   - This is a test file for the experiment engine.
   - It tests normalCdf, watchMinutesPerImpression, computeBayesianProbBOverA, evaluatePackagingExperiment (guardrails and Bayesian decisions), buildDecisionLogEntry, and workflow registration.
   - The tests look comprehensive and pass (we assume). No issues in the test file itself.

5. `apps/web/lib/workflows/youtube-packaging/experiment-engine.ts` (new file)
   - This is the main experiment engine.
   - Defines constants, types, and the main function `evaluatePackagingExperiment`.
   - Also registers the workflow via `registerWorkflow` at the bottom.
   - We note that the file is long (605 lines). We must check for logic errors.

   Let's break down the key parts:

   - Types: VariantMetrics, ExperimentConfig, ExperimentState, ExperimentDecisionKind, ExperimentGuardrailViolation, ExperimentDecision, ExperimentDecisionLogEntry.

   - Constants: 
        YOUTUBE_PACKAGING_EXPERIMENT_KIND
        DEFAULT_MIN_IMPRESSIONS = 500
        DEFAULT_WIN_THRESHOLD = 0.9
        DEFAULT_LOSE_THRESHOLD = 0.1
        DEFAULT_MAX_REGRESSION_PCT = 0.05
        MS_PER_HOUR
        DEFAULT_MIN_TEST_DURATION_MS = 72 hours
        DEFAULT_MAX_TEST_DURATION_MS = 30 days
        DEFAULT_MIN_SWAP_COOLDOWN_MS = 7 days
        DEFAULT_EXPERIMENT_CONFIG

   - Helper functions:
        normalCdf (same as in bayesian.ts? Actually, note: this file has its own normalCdf and computeBayesianProbBOverA, which are similar to the ones in bayesian.ts but not identical. This is a duplication? But note: the bayesian.ts file is also new and might be used elsewhere? However, in this file we are defining our own. We should check if they are the same.

        In experiment-engine.ts:
          export function normalCdf(z: number): number {
            // Reflect for negative z: Φ(-z) = 1 - Φ(z)
            const sign = z >= 0 ? 1 : -1;
            const t = 1 / (1 + 0.2316419 * Math.abs(z));
            const d = 0.398942282 * Math.exp(-0.5 * z * z);
            const poly =
              t *
              (0.3193815 +
                t * (-0.3565638 + t * (1.7814779 + t * (-1.821256 + t * 1.3302744))));
            const tail = d * poly;
            return sign === 1 ? 1 - tail : tail;
          }

        In bayesian.ts:
          function normCdf(z: number): number {
            const t = 1 / (1 + 0.2316419 * Math.abs(z));
            const poly =
              t *
              (0.31938153 +
                t *
                  (-0.356563782 +
                    t * (1.781477937 + t * (-1.821255978 + t * 1.330274429))));
            const cdf = 1 - (1 / Math.sqrt(2 * Math.PI)) * Math.exp(-0.5 * z * z) * poly;
            return z >= 0 ? cdf : 1 - cdf;
          }

        They are different approximations. The one in experiment-engine.ts uses a polynomial without the sqrt(2*pi) factor? Actually, note that the formula in experiment-engine.ts is:

          Φ(z) = 1 - φ(z) * (b1*t + b2*t^2 + b3*t^3 + b4*t^4 + b5*t^5)   for z>=0
          and for z<0: Φ(z) = φ(z) * (b1*t + ...)

        where φ(z) = exp(-z^2/2)/sqrt(2*pi)

        But in the experiment-engine.ts code, we see:
          const d = 0.398942282 * Math.exp(-0.5 * z * z);
          0.398942282 is 1/sqrt(2*pi) (approximately).

        So it is the same as: d * poly, and then for z>=0: 1 - d*poly, for z<0: d*poly.

        The bayesian.ts version does:
          const cdf = 1 - (1 / Math.sqrt(2 * Math.PI)) * Math.exp(-0.5 * z * z) * poly;
          and then returns z>=0 ? cdf : 1 - cdf.

        Since 1/sqrt(2*pi) is approximately 0.39894228, they are the same.

        However, the polynomials are slightly different. The experiment-engine.ts uses:
          0.3193815, -0.3565638, 1.7814779, -1.821256, 1.3302744
        The bayesian.ts uses:
          0.31938153, -0.356563782, 1.781477937, -1.821255978, 1.330274429

        The difference is in the last decimal places. This is acceptable for an approximation, but note: we have two different approximations in the codebase. This might lead to inconsistencies if both are used. However, in the experiment-engine.ts we are using our own normalCdf and computeBayesianProbBOverA, and the bayesian.ts is not imported in this file. So it's isolated.

        But note: the experiment-engine.ts file does not import from bayesian.ts. It re-implements the normalCdf and computeBayesianProbBOverA. This is code duplication. However, the PR is adding new files and we are reviewing for correctness. We must check if the implementation in experiment-engine.ts is correct.

        Let's check the computeBayesianProbBOverA in experiment-engine.ts:

          export function computeBayesianProbBOverA(
            control: VariantMetrics,
            treatment: VariantMetrics
          ): number {
            if (control.impressions === 0 || treatment.impressions === 0) return 0.5;

            const rateA = watchMinutesPerImpression(control);
            const rateB = watchMinutesPerImpression(treatment);

            // Poisson SE²: guard against zero rates to avoid NaN
            const seA2 =
              rateA > 0 ? rateA / control.impressions : 1 / control.impressions;
            const seB2 =
              rateB > 0 ? rateB / treatment.impressions : 1 / treatment.impressions;

            const pooledSE = Math.sqrt(seA2 + seB2);
            if (pooledSE === 0) return rateB > rateA ? 1 : rateB < rateA ? 0 : 0.5;

            return normalCdf((rateB - rateA) / pooledSE);
          }

        This is similar to the one in bayesian.ts, but note: in bayesian.ts they did:

          const varA = rateA / control.impressions;
          const varB = rateB / treatment.impressions;
          const se = Math.sqrt(varA + varB);

        And then z = (rateB - rateA) / se.

        Here, they are doing the same but with a guard for zero rates (to avoid division by zero in the variance? Actually, if rateA is 0 then seA2 becomes 1/control.impressions, which is not zero). This is a different approach.

        Why the guard? The comment says: "guard against zero rates to avoid NaN". But note: if rateA is 0, then the variance of the rate should be 0? Actually, the Poisson rate variance is rate / n. If the rate is 0, then the variance is 0. However, if we have zero watch minutes, then the rate is 0 and the variance is 0. But if we have zero impressions, we already returned 0.5.

        However, note: we have already checked that impressions are not zero (because we returned 0.5 if either is zero). So control.impressions and treatment.impressions are at least 1.

        Now, if rateA is 0 (i.e., watchMinutes=0), then the variance should be 0. But in the code, they set seA2 to 1/control.impressions when rateA==0. That is not correct.

        Example: control: impressions=100, watchMinutes=0 -> rateA=0.
          The variance of the rate (watchMinutesPerImpression) should be 0 because the rate is exactly 0 (no variance). But they set seA2 = 1/100 = 0.01.

        This would overestimate the variance and make the confidence interval wider, leading to a more conservative test (closer to 0.5). This is not correct.

        The correct variance for the rate (when the rate is estimated as watchMinutes/impressions) is (watchMinutes) / (impressions^2) ??? 

        Actually, note: the watchMinutes is a sum of independent Poisson? The model in the bayesian.ts file assumes that the watchMinutes is Poisson distributed with rate λ * impressions? Then the MLE for λ is watchMinutes/impressions, and the variance of the MLE is λ/impressions. But since we don't know λ, we use the MLE: so variance = (watchMinutes/impressions) / impressions = watchMinutes/(impressions^2).

        However, in the experiment-engine.ts, they are using:
          seA2 = rateA > 0 ? rateA / control.impressions : 1 / control.impressions;

        This is equivalent to: (rateA) / control.impressions when rateA>0, and 1/control.impressions when rateA==0.

        But note: rateA = watchMinutes/control.impressions, so:
          rateA / control.impressions = (watchMinutes/control.impressions) / control.impressions = watchMinutes/(control.impressions^2)

        And when rateA==0, they use 1/control.impressions, which is not equal to 0.

        This is a bug.

        The correct approach is to use the estimated variance: rateA / control.impressions, and if rateA is 0 then the variance is 0. However, note that if we have zero watch minutes, we cannot be sure the rate is exactly zero? But the MLE is zero and the variance of the MLE is zero? Actually, the variance of the MLE for a Poisson rate is λ/n, and if λ=0 then the variance is 0.

        Therefore, we should set seA2 = rateA / control.impressions, and if rateA is 0 then it becomes 0.

        But note: if we set it to 0, then we might get division by zero in the pooledSE? However, we have a check for pooledSE===0.

        So the fix is to remove the guard and just do:
          const seA2 = rateA / control.impressions;
          const seB2 = rateB / treatment.impressions;

        However, note that if rateA is 0, then seA2=0, which is correct.

        But wait: what if we have zero impressions? We already returned 0.5 at the top, so we are safe.

        Therefore, the current code in experiment-engine.ts for computeBayesianProbBOverA is incorrect when the rate is zero (because it uses 1/impressions instead of 0). This would lead to an incorrect confidence.

        We should fix it to use the same as in bayesian.ts (without the guard) or at least use the correct variance.

        Alternatively, we can note that the bayesian.ts version does not have this guard and is correct. Since we are duplicating code, we should use the same implementation.

        However, note: the experiment-engine.ts file is the main one and the bayesian.ts is a separate file. We are not importing bayesian.ts in experiment-engine.ts. We have two options:
          Option 1: Fix the variance calculation in experiment-engine.ts to be: rateA / control.impressions (without guard) and similarly for treatment.
          Option 2: Remove the duplication and use the bayesian.ts functions.

        But note: the PR is adding both files. We are reviewing the experiment-engine.ts file. We must fix the variance calculation.

        Let's look at the bayesian.ts version again: it does not have the guard. It simply does:
          const varA = rateA / control.impressions;
          const varB = rateB / treatment.impressions;

        So we should change the experiment-engine.ts to match.

        However, note: the experiment-engine.ts also has a normalCdf function that is slightly different (different polynomial coefficients). We should also consider using the same normalCdf as in bayesian.ts? Or at least make sure they are consistent.

        But note: the normalCdf function in experiment-engine.ts is used only in this file, and the one in bayesian.ts is used only in bayesian.ts. Since they are not shared, it's acceptable to have two different approximations as long as they are both accurate enough. However, for consistency and to avoid confusion, we might want to use the same.

        Given the time, we focus on the variance bug.

        Therefore, in experiment-engine.ts, in the function computeBayesianProbBOverA, we should change:

          const seA2 =
            rateA > 0 ? rateA / control.impressions : 1 / control.impressions;
          const seB2 =
            rateB > 0 ? rateB / treatment.impressions : 1 / treatment.impressions;

        to:

          const seA2 = rateA / control.impressions;
          const seB2 = rateB / treatment.impressions;

        And remove the guard.

        Also note: the check for pooledSE===0: if both seA2 and seB2 are zero, then pooledSE=0. This can happen if both rates are zero. Then we return:
          return rateB > rateA ? 1 : rateB < rateA ? 0 : 0.5;

        But if both rates are zero, then rateB==rateA, so we return 0.5. That is correct.

        However, note: if one rate is zero and the other is positive, then we don't hit the pooledSE===0 branch because one of the variances is positive. So we are safe.

        Let's update the code accordingly.

   - Now, let's check the guardrail checks in experiment-engine.ts (the function `checkGuardrails` inside the file):

        We see:
          function checkGuardrails(
            state: ExperimentState,
            config: ExperimentConfig,
            now: Date
          ): ExperimentGuardrailViolation[] {
            const violations: ExperimentGuardrailViolation[] = [];

            // 1. Min impressions per variant
            if (state.control.impressions < config.minImpressionsPerVariant) {
              violations.push({
                rule: 'min_impressions

itstimwhite (commented):

PR #13379, opened 2026-07-06
ci: wire 7 more jobs to self-hosted runner (10 total)

itstimwhite:
## Problem

Only 3/47 CI jobs use our self-hosted OrbStack arm64 runner pool (5 runners, 10 vCPUs, 23 Gi RAM). Two problems:

**1. Speed:** 44 jobs queue on GitHub's shared `ubuntu-latest` pool, adding 4+ min of queue wait per job. Worst bottleneck: `Path Changes` is the gatekeeper for ***every*** CI run — can't start downstream jobs until it completes.

**2. Resilience:** If the MBP goes down (sleep/reboot/travel), all jobs targeting `jovie-runner` queue forever with no automatic fallback.

### Part A — More jobs on self-hosted runner

Wired 8 more jobs to `CI_FAST_RUNNER`:

| Job | Why safe |
|-----|----------|
| **`ci-path-changes`** | JS GitHub Action, no Docker |
| `ci-biome` | Node.js (setup-node-pnpm) |
| `ci-guardrails` | Node.js (setup-node) |
| `ci-structural-contract` | Node.js+Python |
| `ci-risk-classifier` | Node.js (setup-node-pnpm) |
| `ci-env-example-guard` | bash/grep only |
| `ci-fast` | Aggregator |
| `ci-integration-ready` | Aggregator |

**Before:** 3 jobs on self-hosted  
**After: 11 jobs on self-hosted**  
36 jobs remain on `ubuntu-latest` (Playwright, Lighthouse, Vercel deploy, neon-db).

### Part B — Auto-fallback when MBP is down

`runner-health-monitor.yml` — scheduled every 5 minutes:

- Checks if any `jovie-runner` runner is online
- If all 5 offline for **2+ consecutive checks (~10 min)** → flips `CI_FAST_RUNNER` to `ubuntu-latest`
- When runners come back → restores to `jovie-runner`
- Anti-flap: 2-check debounce in both directions

## Impact

| | Before | After |
|---|---|---|
| Self-hosted jobs | 3 (6%) | **11 (24%)** |
| MBP goes down | jobs queue forever | **auto-fallback in ~10 min** |
| Path Changes queue | 4+ min waiting | **instant pickup**

### DIFF
--- a/.github/actions/setup-node-pnpm/action.yml
+++ b/.github/actions/setup-node-pnpm/action.yml
@@ -13,9 +13,13 @@ runs:
     - name: Trust workspace for git (self-hosted ownership)
       shell: bash
       run: |
-        git config --global --get-all safe.directory \
-          | grep -qxF "$GITHUB_WORKSPACE" \
-          || git config --global --add safe.directory "$GITHUB_WORKSPACE"
+        # No-op when git is not in PATH (minimal self-hosted runner environments).
+        # The safe.directory concern only applies when git is actually used.
+        if command -v git &>/dev/null; then
+          git config --global --get-all safe.directory \
+            | grep -qxF "$GITHUB_WORKSPACE" \
+            || git config --global --add safe.directory "$GITHUB_WORKSPACE"
+        fi
 
     - name: Setup pnpm
       uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4.2.0
--- a/.github/actions/setup-node-pnpm/action.yml
+++ b/.github/actions/setup-node-pnpm/action.yml
@@ -20,4 +20,4 @@ runs:
 
     - name: Install dependencies
       shell: bash
-      run: pnpm install --offline --frozen-lockfile
+      run: pnpm install --frozen-lockfile
--- a/.github/actions/setup-node-pnpm/action.yml
+++ b/.github/actions/setup-node-pnpm/action.yml
@@ -4,6 +4,19 @@ description: 'Sets up Node.js, pnpm, and installs dependencies with optimal cach
 runs:
   using: 'composite'
   steps:
+    # Self-hosted runners reuse a workspace whose owner can differ from the job
+    # user, tripping git's dubious-ownership guard. That makes turbo's SCM probe
+    # report the checkout as "not part of a Git repository" and breaks
+    # `turbo --affected` (typecheck/build/unit tests). Trust the workspace so SCM
+    # detection works. No-op on hosted runners. Idempotent to avoid bloating the
+    # self-hosted ~/.gitconfig across runs.
+    - name: Trust workspace for git (self-hosted ownership)
+      shell: bash
+      run: |
+        git config --global --get-all safe.directory \
+          | grep -qxF "$GITHUB_WORKSPACE" \
+          || git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
     - name: Setup pnpm
       uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4.2.0
       
--- a/.github/workflows/runner-health-monitor.yml
+++ b/.github/workflows/runner-health-monitor.yml
@@ -1,0 +1,129 @@
+name: Runner Health Monitor
+
+# Heartbeat fallback: when the self-hosted OrbStack runner pool goes offline
+# (MBP sleep/reboot/travel), auto-flip CI_FAST_RUNNER from jovie-runner to
+# ubuntu-latest so CI doesn't queue forever. Flips back when runners return.
+#
+# Detects 2+ consecutive lost heartbeats before flipping to avoid flapping
+# on brief network blips.
+
+on:
+  schedule:
+    - cron: '*/5 * * * *'   # every 5 minutes
+  workflow_dispatch:         # manual trigger for testing
+
+permissions:
+  contents: read
+  actions: write             # needed to read runners + update repo variables
+
+jobs:
+  monitor:
+    name: Check runner health
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    env:
+      GH_TOKEN: ${{ github.token }}
+      TARGET_VARIABLE: CI_FAST_RUNNER
+      OFFLINE_TARGET: ubuntu-latest
+      ONLINE_TARGET: jovie-runner
+    steps:
+      - name: Query runner status
+        id: runners
+        run: |
+          set -euo pipefail
+
+          STATE_FILE="${{ runner.temp }}/runner-health-state"
+          PREV="${{ runner.temp }}/runner-health-prev"
+
+          # Fetch runners (online, offline — just not removed)
+          RUNNERS=$(gh api repos/JovieInc/Jovie/actions/runners --jq \
+            '[.runners[] | select(.labels[].name == "jovie-runner") | {name: .name, status: .status, busy: .busy}]')
+
+          TOTAL=$(echo "$RUNNERS" | jq length)
+          ONLINE=$(echo "$RUNNERS" | jq '[.[] | select(.status == "online")] | length')
+          OFFLINE=$(echo "$RUNNERS" | jq '[.[] | select(.status == "offline")] | length')
+
+          echo "total=$TOTAL"
+          echo "online=$ONLINE"
+          echo "offline=$OFFLINE"
+          echo "runner_list=$RUNNERS"
+
+          echo "total=$TOTAL" >> "$GITHUB_OUTPUT"
+          echo "online=$ONLINE" >> "$GITHUB_OUTPUT"
+
+          # Determine current health state
+          if [ "$ONLINE" -eq 0 ]; then
+            echo "health=down" >> "$GITHUB_OUTPUT"
+          elif [ "$ONLINE" -ge 1 ]; then
+            echo "health=up" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Flip to GitHub-hosted fallback if runners are down
+        if: steps.runners.outputs.health == 'down'
+        env:
+          PREV_FLIP_FILE: ${{ runner.temp }}/last-flipped
+        run: |
+          set -euo pipefail
+
+          CURRENT=$(gh api repos/JovieInc/Jovie/actions/variables/CI_FAST_RUNNER --jq '.value' 2>/dev/null || echo "")
+
+          if [ "$CURRENT" = "ubuntu-latest" ]; then
+            echo "Already on fallback (ubuntu-latest). No action needed."
+            exit 0
+          fi
+
+          # Require 2+ consecutive detections before flipping (anti-flap)
+          CONSECUTIVE_FILE="${{ runner.temp }}/runner-down-count"
+          COUNT=0
+          if [ -f "$CONSECUTIVE_FILE" ]; then
+            COUNT=$(cat "$CONSECUTIVE_FILE")
+          fi
+          COUNT=$((COUNT + 1))
+          echo "$COUNT" > "$CONSECUTIVE_FILE"
+
+          FLAP_THRESHOLD=2  # require 2+ consecutive checks (~10 min) before flipping
+          if [ "$COUNT" -lt "$FLAP_THRESHOLD" ]; then
+            echo "First detection of runner outage. Need $((FLAP_THRESHOLD - COUNT)) more. (count=$COUNT)"
+            exit 0
+          fi
+
+          echo "Runners offline for $COUNT consecutive checks. Flipping CI_FAST_RUNNER → ubuntu-latest..."
+          gh api -X PATCH repos/JovieInc/Jovie/actions/variables/CI_FAST_RUNNER \
+            -f name=CI_FAST_RUNNER -f value=ubuntu-latest 2>&1
+
+          echo "last_flip=down" > "$PREV_FLIP_FILE"
+
+      - name: Restore self-hosted runners when they come back
+        if: steps.runners.outputs.health == 'up'
+        run: |
+          set -euo pipefail
+
+          CURRENT=$(gh api repos/JovieInc/Jovie/actions/variables/CI_FAST_RUNNER --jq '.value' 2>/dev/null || echo "")
+
+          if [ "$CURRENT" = "jovie-runner" ]; then
+            echo "Already on self-hosted (jovie-runner). No action needed."
+            exit 0
+          fi
+
+          # Only restore if we were previously on fallback (debounce: 2 checks)
+          CONSECUTIVE_FILE="${{ runner.temp }}/runner-up-count"
+          COUNT=0
+          if [ -f "$CONSECUTIVE_FILE" ]; then
+            COUNT=$(cat "$CONSECUTIVE_FILE")
+          fi
+          COUNT=$((COUNT + 1))
+          echo "$COUNT" > "$CONSECUTIVE_FILE"
+
+          FLAP_THRESHOLD=2
+          if [ "$COUNT" -lt "$FLAP_THRESHOLD" ]; then
+            echo "First detection of runner recovery. Need $((FLAP_THRESHOLD - COUNT)) more. (count=$COUNT)"
+            exit 0
+          fi
+
+          echo "Runners back for $COUNT consecutive checks. Restoring CI_FAST_RUNNER → jovie-runner..."
+          gh api -X PATCH repos/JovieInc/Jovie/actions/variables/CI_FAST_RUNNER \
+            -f name=CI_FAST_RUNNER -f value=jovie-runner 2>&1
+
+      - name: Log health state
+        run: |
+          echo "::notice::Runner health: online=${{ steps.runners.outputs.online }}/${{ steps.runners.outputs.total }}"


itstimwhite:
### Merge activity

* **Jul 6, 3:26 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 6, 3:52 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 6, 3:55 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 6, 3:56 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 6, 5:06 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 6, 5:30 AM UTC**: The merge label 'merge-queue' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) once it meets the requirements.
* **Jul 6, 5:30 AM UTC**: `itstimwhite` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie).
* **Jul 6, 5:31 AM UTC**: CI is running for this pull request on a draft pull request ([#13381](https://app.graphite.com/github/pr/JovieInc/Jovie/13381)) due to your merge queue CI optimization settings.
* **Jul 6, 5:35 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=JovieInc&repo=Jovie) via draft PR: [#13381](https://app.graphite.com/github/pr/JovieInc/Jovie/13381).

itstimwhite:
We are given a diff for a new workflow file: .github/workflows/runner-health-monitor.yml

We are to check for:
1. Logic errors
2. Security
3. Data integrity
4. Error handling
5. Cross-file impact

Let's go through the workflow step by step.

First, note that the workflow is designed to monitor the health of self-hosted runners labeled "jovie-runner" and automatically flip a repository variable (CI_FAST_RUNNER) between "jovie-runner" (self-hosted) and "ubuntu-latest" (GitHub-hosted) based on the health of the runners.

The workflow runs every 5 minutes and has three main steps:
1. Query runner status (to get the count of online/offline runners with label "jovie-runner")
2. Flip to GitHub-hosted fallback if runners are down (with anti-flap: requires 2 consecutive detections of being down)
3. Restore self-hosted runners when they come back (with anti-flap: requires 2 consecutive detections of being up)
4. Log health state

Now, let's check for issues:

### 1. Logic errors

We'll look for:
- Race conditions: The workflow uses temporary files in the runner's temp directory. Since each job runs in a fresh runner, there is no shared state between runs of this workflow. However, note that the workflow is scheduled and may have multiple runs overlapping? But the schedule is every 5 minutes and the job times out after 2 minutes, so it's unlikely to overlap. Even if it did, each run uses its own temporary files (because each run gets a fresh runner and its own temp directory). So no race condition between runs of this workflow.

But note: the workflow is updating a repository variable (CI_FAST_RUNNER). If two runs of this workflow happen at the same time, they might both read the same current value and then both try to update. However, the schedule is every 5 minutes and the job takes less than 2 minutes, so it's safe. But to be absolutely safe, we might consider using a lock? However, the problem of overlapping runs is low and the consequence is that we might flip twice (which is harmless because flipping to the same value twice is idempotent). So not a critical issue.

- Null dereferences: We are using `set -euo pipefail` so the script will exit on any error or unset variable. We are using `gh api` commands and then processing the output with `jq`. We check for the existence of files before reading them.

- Off-by-one: The anti-flap threshold is set to 2, meaning we require 2 consecutive detections (so 2 checks, which is 10 minutes) before flipping. This is as intended.

- Incorrect conditionals: 
  - In the "Flip to GitHub-hosted fallback" step: 
      if [ "$ONLINE" -eq 0 ]; then
        echo "health=down" >> "$GITHUB_OUTPUT"
      elif [ "$ONLINE" -ge 1 ]; then
        echo "health=up" >> "$GITHUB_OUTPUT"
      fi
    This covers all cases (ONLINE is either 0 or >=1). Note: ONLINE cannot be negative.

  - In the flip step: 
      We check if the current value of CI_FAST_RUNNER is already "ubuntu-latest", then we exit.
      Then we count consecutive downs and only flip when we hit the threshold.

  - In the restore step: 
      We check if the current value is already "jovie-runner", then exit.
      Then we count consecutive ups and only restore when we hit the threshold.

  This seems correct.

However, note that the workflow uses two separate counters for down and up. This is correct because we want to debounce in both directions.

But there is a potential issue: what if the runners go down and then up again before we have flipped? The down counter would be incremented and then reset when we see an up? Actually, no: because when the runners are up, we are in the restore step and we increment the up counter. The down counter is only incremented when we see down. So if we have a pattern: down, up, down, up, ... then the down counter would be 1 on the first down, then reset to 0 on the up (because we don't touch the down counter on up) and then 1 again on the next down. So we would never flip. This is the intended behavior (anti-flap).

But note: the down counter is stored in a file that is specific to the runner's temp directory. Since each run of the workflow uses a fresh runner, the file does not persist between runs. Therefore, the counter is reset every time the workflow runs? 

Wait: the workflow runs every 5 minutes, but each run is on a fresh runner. Therefore, the temporary files (like `${{ runner.temp }}/runner-down-count`) are not preserved between runs. This breaks the anti-flap mechanism because we need to remember the count across runs.

This is a critical logic error.

Explanation:
- The workflow runs every 5 minutes on a fresh runner (ubuntu-latest).
- In each run, we check the runner status.
- We then try to read a counter from a file in the runner's temp directory (which is unique to that run) to see how many consecutive downs (or ups) we have seen.
- But since the runner is fresh, the file does not exist from a previous run. Therefore, the counter always starts at 0 in every run.

So the anti-flap mechanism (requiring 2 consecutive checks) will never work because we never have a count of 1 from the previous run to build upon.

We need to persist the state between runs. Since we are allowed to use repository variables or secrets, we could store the counter in a repository variable. Alternatively, we could use the repository's environment or a file in the repository (but that would require committing, which is not ideal for a counter that changes frequently).

Alternatively, we can store the state in two repository variables: one for the down count and one for the up count? But note: we only need to remember the consecutive count for the current state (either we are counting downs or ups, not both at the same time). Actually, we are counting consecutive downs when we are in a down state and consecutive ups when we are in an up state. And we reset the opposite counter when the state changes.

However, note that the workflow does not know the previous state from the last run because it runs on a fresh runner. Therefore, we must store the state (the current health and the consecutive count) in a persistent way.

We are already storing the last flip in a file (PREV_FLIP_FILE) but that is also in the temporary directory and not persistent.

Therefore, the entire anti-flap mechanism is broken because it relies on temporary files that do not persist between workflow runs.

How to fix?
We can store the consecutive count and the last known state (or the last flip direction) in repository variables. We have permissions to write to repository variables (we have `actions: write` permission).

We already are reading and writing the variable `CI_FAST_RUNNER`. We could use two additional variables, for example:
- `CI_FAST_RUNNER_DOWN_COUNT`: to count consecutive downs
- `CI_FAST_RUNNER_UP_COUNT`: to count consecutive ups

But note: we only need one counter at a time. Actually, we can store:
- The last known health state (from the perspective of the monitor) and the consecutive count for that state.

Alternatively, we can store:
- The last flip direction (or the last known state of the runners as perceived by the monitor) and the consecutive count for the current streak.

However, note that the monitor's perception of health (online>=1 -> up, online==0 -> down) is what we are using.

We can do:

Let's define two repository variables:
- `RUNNER_HEALTH_STATE`: either "up" or "down" (the last state we detected that was sustained for at least the threshold? Actually, we want to remember the state we are currently in for the purpose of counting consecutive same states)
- `RUNNER_HEALTH_COUNT`: the number of consecutive times we have seen that state.

But note: we start with an unknown state. We can initialize the state to "up" (assuming runners are up) and count=0? Or we can read the current value of CI_FAST_RUNNER to infer the state? 

Alternatively, we can avoid storing state by using the current value of CI_FAST_RUNNER and the last flip time? But that might be complex.

Given the constraints, I propose:

We store two repository variables:
- `RUNNER_HEALTH_CONSECUTIVE_DOWN`: integer, the number of consecutive times we have seen the runners down (reset to 0 when we see up)
- `RUNNER_HEALTH_CONSECUTIVE_UP`: integer, the number of consecutive times we have seen the runners up (reset to 0 when we see down)

But note: we don't need both at the same time. We can store one variable for the current streak and a flag for what the streak is about? Or we can store:

- `RUNNER_HEALTH_STREAK`: integer (the current consecutive count of the same state)
- `RUNNER_HEALTH_STATE`: string ("up" or "down") - the state that the streak is counting

Then, in each run:
1. Determine current state: if ONLINE>=1 -> state="up", else state="down"
2. Read the stored state and streak from the variables.
3. If the stored state equals the current state, then increment the streak.
   Else, reset the streak to 1 and update the stored state to the current state.
4. Then, if the current state is "down" and the streak >= FLAP_THRESHOLD (2) and the current CI_FAST_RUNNER is not already set to fallback, then flip to fallback.
5. If the current state is "up" and the streak >= FLAP_THRESHOLD (2) and the current CI_FAST_RUNNER is not already set to self-hosted, then flip to self-hosted.

But note: we must avoid flipping multiple times. We only want to flip when we cross the threshold.

Alternatively, we can stick to the original plan but store the counters in repository variables.

Given that we already have the ability to write repository variables, let's change the workflow to use two repository variables for the counters: one for down and one for up.

We'll have:
- `CI_FAST_RUNNER_DOWN_COUNT`: the number of consecutive times we have seen the runners down (reset to 0 when we see up)
- `CI_FAST_RUNNER_UP_COUNT`: the number of consecutive times we have seen the runners up (reset to 0 when we see down)

Steps in the workflow:

1. Query runner status -> get ONLINE count -> determine current state (up if ONLINE>=1, down if ONLINE==0)

2. Read the current down and up counts from the repository variables (if they don't exist, treat as 0).

3. If current state is down:
      new_down_count = old_down_count + 1
      new_up_count = 0   // because we broke the up streak
   Else (current state is up):
      new_up_count = old_up_count + 1
      new_down_count = 0

4. Write the new counts back to the repository variables.

5. Then, for flipping to fallback (down state):
      If current state is down and new_down_count >= FLAP_THRESHOLD (2) and the current CI_FAST_RUNNER is not "ubuntu-latest", then flip to ubuntu-latest.

6. For restoring (up state):
      If current state is up and new_up_count >= FLAP_THRESHOLD (2) and the current CI_FAST_RUNNER is not "jovie-runner", then flip to jovie-runner.

But note: we must read the current value of CI_FAST_RUNNER to avoid flipping if we are already in the target.

However, note that we are updating the counts in every run, so we have the streak.

This approach uses persistent storage (repository variables) and therefore survives between runs.

But note: we have to be cautious about concurrent runs. If two runs happen at the same time, they might both read the same old counts, then both update and we lose one increment. However, the job runs for less than 2 minutes and is scheduled every 5 minutes, so the chance of overlap is low. Moreover, even if we lose an update, it just means we might be one count behind, which is acceptable because we require 2 consecutive. We might flip one run later than intended, but it's safe.

Alternatively, we could use a lock via a repository variable? But that complicates things.

Given the low frequency and the fact that the consequence of a missed flip is just a bit longer queue time (which is acceptable), we can proceed without a lock.

Now, let's see if the current workflow has this issue: yes, because it uses temporary files that are not persistent.

Therefore, we have a logic error: the anti-flap mechanism does not work because the counters are not persisted between workflow runs.

### 2. Security

We are using:
- `GH_TOKEN: ${{ github.token }}` - this is the default token, which is scoped to the repository and has the permissions we set (contents: read, actions: write). This is acceptable.

We are making API calls to:
- `gh api repos/JovieInc/Jovie/actions/runners` - to list runners (we have contents: read? Actually, the token we are using is the default GITHUB_TOKEN, and we have set `permissions: contents: read, actions: write`. The `actions/runners` endpoint requires `actions: read` for public repos? But note: we are in a private repo? The repo is JovieInc/Jovie, which is likely private. However, the permission we set is `actions: write` which includes read? Actually, the permission `actions: write` allows both read and write for actions. So it's okay.

- `gh api repos/JovieInc/Jovie/actions/variables/CI_FAST_RUNNER` - to read a variable (requires `actions: read` for variables? Actually, the variables endpoint: to read a variable we need `actions: read` and to update we need `actions: write`. We have `actions: write` so we can read and write.

We are not exposing any secrets in logs because we are not printing the token or any variable values that are secrets. We are only printing counts and status.

We are using `set -euo pipefail` which is good for error handling.

No obvious security issues.

### 3. Data integrity

We are updating a repository variable (CI_FAST_RUNNER) that is used by other workflows to determine which runner to use. We are changing it between two known values: "jovie-runner" and "ubuntu-latest". We are reading the current value before updating to avoid unnecessary updates.

We are also storing two new counters (if we were to fix the logic) in repository variables. We would need to create them if they don't exist.

But note: the current workflow does not create any new variables, it only reads and writes CI_FAST_RUNNER. So no data integrity issue there.

However, the current workflow does not persist the counters, so the anti-flap mechanism is broken. That is a logic issue, not data integrity.

### 4. Error handling

We are using `set -euo pipefail` so the script will exit on any error, unset variable, or failed pipe.

We are checking the exit ofile the existence of files before reading them (for the temporary counters) but as we noted, those are not persistent so the mechanism fails.

We are using `2>/dev/null` when reading the variable to avoid an error if the variable doesn't exist, and then defaulting to empty string.

We are checking the current value of the variable to avoid updating if it's already set to the target.

We are not handling the case where the `gh api` call fails (e.g., network error, rate limiting, etc.). However, the `set -e` will cause the step to fail if the `gh api` command fails. And the job has a timeout of 2 minutes.

But note: if the `gh api` call to list runners fails, then we set RUNNERS to an empty string? Actually, no: because we are running in a pipe and we have `set -euo pipefail`, if the `gh api` command fails, the entire step will fail and we won't get to set the output variables.

We might want to handle transient errors by retrying? But the job is short and runs every 5 minutes, so a single failure is not critical. We can leave it as is.

However, note that we are using the temporary files to store state between steps in the same job. Since the job runs in a single runner and we are using the same temporary directory for all steps, the files are shared between steps in the same job. That is acceptable.

But the problem is that the temporary directory is not shared between different runs of the workflow (because each run is on a fresh runner). So the anti-flap mechanism that relies on these temporary files is broken.

### 5. Cross-file impact

This workflow updates the repository variable `CI_FAST_RUNNER`. Other workflows that use this variable to set `runs-on` will be affected.

We are changing the value of this variable based on the health of the self-hosted runners.

We are told that 8 more jobs are being wired to use `CI_FAST_RUNNER`. So if we flip the variable to `ubuntu-latest`, those jobs will run on GitHub-hosted runners, and if we flip to `jovie-runner`, they will run on the self-hosted runners.

This is the intended behavior.

We must ensure that the variable `CI_FAST_RUNNER` is set to one of the two allowed values: either "jovie-runner" or "ubuntu-latest". Our workflow only sets it to these two values, so it's safe.

However, note that if the variable is not set at all, then the workflows that use it might fail. But we are reading it and if it's not set we treat it as empty string. Then, in the flip step, if it's empty and we are down, we will flip to ubuntu-latest. Similarly, in the restore step, if it's empty and we are up, we will flip to jovie-runner.

But note: the workflows that use `CI_FAST_RUNNER` must have a default? Actually, in the workflow syntax, if the variable is not set, it will be an empty string and then the `runs-on` will be invalid. So we must ensure the variable is always set.

We are initializing it by flipping to one of the two values when we detect a state that has persisted for 2 checks. But what if the workflow has never run? Then the variable might be unset.

We should consider initializing the variable to a safe default

itstimwhite (commented):

itstimwhite:
**Typecheck failure triage — dispatched fix (`node-version: "20"`) is a false premise; not pushed.**

Evidence:
- Every ci.yml job (incl. Typecheck) uses `./.github/actions/setup-node-pnpm`, which already pins `node-version: '22'` on `actions/setup-node` v4.4.0. There are no un-pinned setup-node steps to fix; the one direct usage (ci.yml L453) is v6.4.0 pinned to '22'. `pnpm/action-setup` is already v4 (v4.2.0).
- Pinning Node **20** would violate `package.json` engines (`node >=22.13.0 <23`) and `.nvmrc` (22.13.0) — it would break CI on all runners, not fix the self-hosted one.
- The failing job (check-run 85290718018) ran on `orbstack-arm64-5` (`jovie-runner`) and failed at `pnpm turbo typecheck --affected --force` (exit 1) **after** setup-node succeeded — so the job was on Node 22, not the runner's default 24. The "Node.js 20 deprecated… forced to run on Node.js 24" annotation is a warning about the *action runtime* (GitHub is deprecating Node 20 action runtimes), not the job's Node.
- Because a workflow-file change is a Turbo global-hash change, `--affected` ran the full typecheck on the arm64 runner. Failure is likely arm64/OrbStack-environment-specific.

Next step (needs repo admin): pull the job logs for run 28766089977 / job 85290718018 — the actual tsc output is 403-gated for non-admins. That will show whether it's a real type error, an arm64 native-binary gap, or an OOM on the runner.

Dispatch ID: n/a (thread cmr8po4vg2e7y07ad9gsp8d53)

itstimwhite:
We are given two files changed in the PR:

1. .github/actions/setup-node-pnpm/action.yml
2. .github/workflows/runner-health-monitor.yml

We are to focus on:
1. Logic errors
2. Security
3. Data integrity
4. Error handling
5. Cross-file impact

Let's examine each file.

First, .github/actions/setup-node-pnpm/action.yml:

The changes are:
- Added a step to trust the workspace for git (self-hosted ownership) to avoid issues with turbo's SCM probe.
- Changed the install command from `pnpm install --offline --frozen-lockfile` to `pnpm install --frozen-lockfile` (removing the `--offline` flag).

Analysis:

1. Logic errors:
   - The added step: 
        git config --global --get-all safe.directory \
          | grep -qxF "$GITHUB_WORKSPACE" \
          || git config --global --add safe.directory "$GITHUB_WORKSPACE"
     This is safe and idempotent. It checks if the workspace is already in the safe.directory list and adds it if not.
     However, note that the step uses `--global` which affects the entire runner. Since the runner is self-hosted and dedicated to Jovie, this is acceptable.
     But note: the step is added in a composite action that might be used in multiple jobs. If the same runner runs multiple jobs, the global git config might be modified multiple times (but it's idempotent so it's okay).

   - Removing `--offline` from `pnpm install`:
        Before: `pnpm install --offline --frozen-lockfile`
        After:  `pnpm install --frozen-lockfile`
     The `--offline` flag tells pnpm to not attempt to fetch from the network and rely solely on the local cache.
     Without `--offline`, if the lockfile is frozen and the dependencies are not in the cache, pnpm will try to fetch from the registry.
     However, note that the action is named "setup-node-pnpm" and it is expected to set up the environment and install dependencies.
     The removal of `--offline` might be intentional to allow installation if the cache is missing (for example, on a fresh self-hosted runner).
     But note: the action also has a step that sets up pnpm (using pnpm/action-setup) and then installs dependencies.
     The `--frozen-lockfile` flag ensures that the lockfile is not updated and that the exact versions are installed.

     Why was `--offline` there? Probably to speed up installation by avoiding network calls and relying on the cache.
     However, if the cache is not populated (e.g., first run on a runner), then the installation would fail without `--offline`? 
     Actually, without `--offline`, it would try to download from the registry. So removing `--offline` makes the step more robust for first-time runs.

     But note: the action is used in the context of a CI workflow that likely has caching set up (e.g., using actions/cache for pnpm store).
     However, the PR does not show the workflows that use this action, so we cannot be sure.

     However, the problem statement says that the self-hosted runner is being used for more jobs. The change to remove `--offline` might be to avoid failures when the cache is not present.

     This change does not seem to introduce a logic error. It might be a trade-off between speed and robustness.

     But note: the action is called `setup-node-pnpm` and it is expected to set up the environment. The removal of `--offline` is acceptable.

2. Security:
   - No hardcoded secrets.
   - The step that modifies git config is safe because it only adds a directory to the safe list, which is a configuration change and not a security vulnerability.
   - No injection vulnerabilities because the `$GITHUB_WORKSPACE` is used in a grep pattern with `-F` (fixed string) and then passed to `git config` as an argument (so it's not interpreted as code).

3. Data integrity:
   - Not applicable (no database changes).

4. Error handling:
   - The added step uses `set -euo pipefail` implicitly? Actually, the step is a shell script and the action does not set `set -euo pipefail` for the step, but the commands are:
        git config ... | grep ... || git config ...
     If the first `git config` fails, the pipe to grep might fail, but the `||` will run the second git config. However, note that the first `git config` (--get-all) should not fail unless there's an internal git error. Similarly, the second `git config` (--add) should not fail unless there's a permission issue or invalid input.

     But note: the step does not have explicit error handling. However, the action as a whole does not have error handling for this step. If the git config command fails (for example, if the user doesn't have permission to write to the global git config), then the step will fail and the job will fail.

     Is that acceptable? The step is meant to fix a known issue on self-hosted runners. If it fails, then the turbo SCM probe might break. So we want the job to fail if we cannot set up the git safe directory.

     Therefore, no additional error handling is needed.

   - The change in the install command: removing `--offline` might cause the step to take longer if the cache is cold, but it will not fail due to network issues (unless there's no network, which is a separate issue). However, if there is no network and the cache is missing, then the installation will fail. But that is the same as before: without `--offline` and without cache, it would fail because it tries to download. With `--offline` and without cache, it would fail immediately. So the behavior in the case of missing cache and no network is the same (failure). The difference is that with `--offline` removed, if there is a network and the cache is missing, it will download and succeed (whereas before it would fail). So it's more robust.

     However, note that the action is running on a self-hosted runner that is expected to have network access (since it's used for CI). So this change is acceptable.

5. Cross-file impact:
   - This action is used in workflows that run on the self-hosted runner. The change to the install command (removing `--offline`) might affect the speed of installation (slightly slower if cache is warm, but more robust if cache is cold). However, the problem statement does not indicate any issue with the installation step.

   - The added step for git config is necessary for the turbo SCM probe to work on self-hosted runners. Without it, jobs that use turbo (like typecheck, build, unit tests) might fail because they think the checkout is not part of a Git repository. Therefore, this change is critical for the jobs that are being moved to the self-hosted runner (such as `ci-biome`, `ci-guardrails`, etc.) to function correctly.

     So the cross-file impact is positive: it enables the additional jobs to run correctly on the self-hosted runner.

Now, let's examine the second file: .github/workflows/runner-health-monitor.yml

This is a new workflow file.

We'll break it down:

1. Logic errors:
   - The workflow runs on `ubuntu-latest` (so it's not affected by the self-hosted runner status).
   - It checks the status of runners with the label "jovie-runner".
   - It counts total, online, and offline runners.
   - It sets an output `health` to "down" if ONLINE==0, otherwise "up".
   - Then, if health is down, it tries to flip the variable `CI_FAST_RUNNER` to `ubuntu-latest` (the fallback) but only after 2 consecutive detections (to avoid flapping).
   - If health is up, it tries to flip back to `jovie-runner` after 2 consecutive detections of being up.

   Potential issues:

   a) The workflow uses `gh api` to interact with GitHub's API. It assumes that the `gh` CLI is available and authenticated. The workflow sets `env: GH_TOKEN: ${{ github.token }}` which provides a token for the `gh` CLI. This is standard.

   b) The workflow reads the variable `CI_FAST_RUNNER` and updates it. It uses:
        gh api repos/JovieInc/Jovie/actions/variables/CI_FAST_RUNNER --jq '.value'
     This requires that the variable exists. What if it doesn't? The command is followed by `|| echo ""` so if the variable doesn't exist, it returns an empty string.

     Then, in the "flip to fallback" step:
        if [ "$CURRENT" = "ubuntu-latest" ]; then ... exit 0
     If the variable doesn't exist, then CURRENT is empty string, so it won't match "ubuntu-latest", and it will proceed to try to flip.

     Similarly, in the "restore" step, if the variable doesn't exist, it won't match "jovie-runner", so it will try to flip to `jovie-runner`.

     However, note that the variable `CI_FAST_RUNNER` is expected to be set by the repository (as per the problem description). It should exist. But to be safe, we might want to check for existence.

     However, the problem says that the variable is used to control which runner the jobs use. It is set in the repository settings. So it should exist.

     But what if it is deleted? Then the workflow would try to set it. Setting a variable that doesn't exist is allowed by the GitHub API (it will create it). So it's acceptable.

   c) The anti-flap mechanism:
        For flipping to fallback (when runners are down):
          - It uses a file to count consecutive downs.
          - It requires 2 consecutive downs to flip.

        For restoring (when runners are up):
          - It uses a separate file to count consecutive ups.
          - It requires 2 consecutive ups to restore.

     However, note that the two counters are stored in different files:
        - For down: CONSECUTIVE_FILE="${{ runner.temp }}/runner-down-count"
        - For up:   CONSECUTIVE_FILE="${{ runner.temp }}/runner-up-count"

     This is correct because we want to count consecutive downs and consecutive ups separately.

     But note: the state of the runner health can change from down to up and vice versa. The counters are reset when the health changes? Actually, no: 
        - When health is down, we update the down counter and leave the up counter untouched (but the up counter step doesn't run because the condition is `if: steps.runners.outputs.health == 'up'`).
        - Similarly, when health is up, we update the up counter and leave the down counter untouched.

     However, what if the health goes from down to up? Then the down counter file is left with the last count (say, 2) and then when we get an up, we start counting up from 0 (if the up counter file doesn't exist) or from the last up count.

     This is acceptable because we only care about consecutive same states.

   d) The workflow uses `runner.temp` for temporary files. This is unique to each job run, so it's safe.

   e) The workflow has a timeout of 2 minutes. The steps are quick (API calls and shell commands) so it should be fine.

   f) The workflow uses `gh api` to patch the variable. It does not check the response for errors. If the API call fails (for example, due to rate limits, or insufficient permissions), then the step will fail and the job will fail. However, note that the workflow has the `actions: write` permission, which is required to update variables.

     But what if the API call fails? The step does not have error handling. We see:
        gh api -X PATCH ... 2>&1
     and then the step continues. However, if the `gh api` command fails (non-zero exit code), then because of `set -euo pipefail` at the top of the run block, the step will exit with an error and the job will fail.

     Is that acceptable? We want to know if the flip failed. However, note that the workflow is designed to run every 5 minutes. If one flip attempt fails, the next run (5 minutes later) will try again.

     But consider: if the variable update fails due to a transient error, we might want to retry. However, the workflow does not retry. But since it runs every 5 minutes, it will eventually succeed.

     However, note that the anti-flap mechanism relies on consecutive counts. If the flip attempt fails, then the variable is not updated, so the next run will see the same state (still down) and increment the counter again. Then it will try again. So it will keep trying every 5 minutes until it succeeds.

     This is acceptable.

   g) The workflow logs the health state at the end with a notice. This is fine.

2. Security:
   - The workflow uses the `github.token` which is a token with the permissions granted to the workflow (here: contents: read, actions: write). This is the least privilege required.
   - No hardcoded secrets.
   - The API calls are to the same repository (JovieInc/Jovie) so no cross-repository leakage.

3. Data integrity:
   - Not applicable (no database).

4. Error handling:
   - As noted, the `gh api` calls are run with `set -euo pipefail` so any error in the command will cause the step to fail.
   - However, we might want to handle specific errors (like network errors) by retrying within the step? But the workflow runs every 5 minutes, so a transient error will be retried soon.

   - One potential issue: the workflow uses `jq` to parse JSON. If the API response is not valid JSON, then `jq` will fail and the step will fail. But the GitHub API should return valid JSON.

   - Another issue: the workflow assumes that the `gh` CLI is installed. It is, because the runner has the `gh` CLI pre-installed (as part of the GitHub Actions runner image).

5. Cross-file impact:
   - This workflow updates the repository variable `CI_FAST_RUNNER`. This variable is used in other workflows to determine which runner to use for jobs (by setting `runs-on: ${{ vars.CI_FAST_RUNNER }}`).
   - Therefore, it has a direct impact on the CI jobs that are wired to use `CI_FAST_RUNNER`.
   - The change is safe because it only flips between two known values: `jovie-runner` and `ubuntu-latest`.
   - The workflow includes anti-flap to prevent rapid toggling.

In the setup-node-pnpm action:
  - The step that adds the safe directory: 
        git config --global --get-all safe.directory \
          | grep -qxF "$GITHUB_WORKSPACE" \
          || git config --global --add safe.directory "$GITHUB_WORKSPACE"
    This is safe, but note: the `--global` flag. If the runner is used by multiple users or multiple jobs that run as different users, then the global git config might be shared. However, the self-hosted runners are dedicated to Jovie and likely run as the same user (the runner user). So it's acceptable.

    However, note: the action is composite and might be used in a job that runs as a different user? In GitHub Actions, each job runs in a fresh container or VM, but for self-hosted runners, the job runs as the runner user (the user that started the runner service). So all jobs on the same runner run as the same user. Therefore, the global git config is shared among jobs on the same runner, but that's okay because we want the same setting for all jobs.

In the runner-health-monitor workflow:
  - The workflow uses `gh api` to get the list of runners. It filters by the label "jovie-runner". However, note that the runner might have multiple labels. The query:
        [.runners[] | select(.labels[].name == "jovie-runner") | {name: .name, status: .status, busy: .busy}]
    This is correct: it selects any runner that has at least one label named "jovie-runner".

  - The workflow then counts the total, online, and offline. Note: the GitHub API for runners returns a list of runners that are not removed. The status can be "online", "offline", or "idle"? Actually, the documentation says: 
        status: The status of a runner. Can be one of online, offline, or idle.
    But note: the API endpoint used is `repos/{owner}/{repo}/actions/runners` and the status field can be "online", "offline", or "idle". However, in the workflow, we are only counting "online" as up. What about "idle"?

    Looking at the GitHub documentation: 
        https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#list-self-hosted-runners-for-an-organization
        The status can be: online, offline, or idle.

    However, note that the workflow considers a runner as "online" only if the status is exactly "online". What about "idle"? 

    In the context of GitHub Actions, a runner that is idle is still available to pick up jobs. So we should consider both "online" and "idle" as available.

    But wait: the API documentation for the endpoint says:
        status	string
        The status of a runner. Can be one of online, offline, or idle.

    And in the workflow, we have:
        ONLINE=$(echo "$RUNNERS" | jq '[.[] | select(.status == "online")] | length')

    This misses the "idle" state.

    This is a logic error: a runner that is idle is ready to run jobs, so it should be considered as available.

    Therefore, we should change the condition to include both "online" and "idle".

    Alternatively, we can consider any runner that is not "offline" as available.

    Let's change the logic for ONLINE to count runners that are not offline.

    We can do:
        ONLINE=$(echo "$RUNNERS" | jq '[.[] | select(.status != "offline")] | length')

    But note: the API might return other statuses? According to the documentation, only online, offline, idle.

    Alternatively, we can count the offline and then subtract from total? But we already have OFFLINE

itstimwhite (commented):

Issue #12315, opened 2026-06-28
Replace the "create share drop" form with a share menu (links already exist)


itstimwhite:
Parent: JovieInc/Jovie#12313. Source: Tim #product 2026-06-28.

## Problem

The right rail adds a multi-field "Create share drop" FORM (layout picker, title, comment, passphrase, expiry, allow-downloads) nested in a collapsible card inside a scrolling rail. Tim: excessive slop; you should not have to fill a form to get a link — every entity already HAS a link.

## Direction (decided by Tim)

Every entity has a public link and a private/secret link that already exist. Replace the form with a share MENU (reuse the sidebar share-dropdown pattern + UTM builder, JovieInc/Jovie#5371):

* Open link · Copy link
* Share privately → copy the secret link (no passphrase — just a secret URL; toggle off anytime)
* Share publicly → pick destination → auto-generate UTMs (Jovie picks secret-vs-public + UTMs by audience: e.g. supervisor=secret, fan blast=public)<br>Remove: title (defaults to release title), comment, passphrase, expiry (it “expires” when the song releases / via the off toggle), allow-downloads (defer; supervisor-only, later).

## Acceptance

* No share form; sharing is a menu reachable from the entity context/ellipsis menu.
* Secret link + public link both available; no required text input to share.
* Public share builds UTMs (reuse JovieInc/Jovie#4020 passthrough + JovieInc/Jovie#5371 builder).

## Related / regression

Reopen JovieInc/Jovie#10812 (its "share-card redesign" shipped this form). Reuse JovieInc/Jovie#5371, JovieInc/Jovie#4020.

## Shape & model

Medium–large, decided direction → Codex. HOLD ready-for-intake until a quick share-menu comp is approved (design gate).

itstimwhite:
Jovie agent (codex issue shipper) claimed this issue.

Branch: `codex/gh-12315-replace-the-create-share-drop-form-with-a-share`
Risk: `high`
Model route: `escalation` using `opus`
Integration branch: not used for this issue

itstimwhite:
Jovie agent (codex issue shipper) stopped on a real blocker.

Agent exited without successfully shipping.

Status: `143`
Signal: `null`
Log: `/Users/timwhite/.hermes/logs/codex-issue-shipper/github-12315-2026-06-29T09-07-53-860z.log`
[...]

PR #13400, opened 2026-07-06
profile: UTM Builder — enforce required Source, canonical Input, copy-state guard

### LINKED ISSUE #13223: UTM Builder dialog polish: enforce required Source, use canonical input primitive, copy-state guard
**Problem**: UtmBuilderDialog marks Source as required but never enforces it — 'Copy link' happily copies the bare profile URL with zero UTM params (silent no-op for the user's tracking intent). Inputs are raw hand-rolled `<input>` elements with inline utility classes instead of the canonical Input primitive (duplication class tracked by JovieInc/Jovie#12013).

**Where**: apps/web/components/features/profile/UtmBuilderDialog.tsx.

**Fix**: (1) Disable Copy/Open (or inline-name the failing rule per Best-UX guardrails) until `utm_source` is non-empty; keep the live URL preview as-is (it works). (2) Swap raw inputs for the canonical Input primitive with tokenized focus states. (3) Keep the dialog otherwise — the live preview + Open/Copy actions are right. Note: manual UTM Builder remains the power-tool path; audience-based auto-UTMs arrive with the canonical share menu (JovieInc/Jovie#12315) — no conflict, don't expand scope here.

**Acceptance**: Copy with empty Source is prevented with a plain-language inline message naming the rule; inputs are the canonical primitive; result URL preview updates live; copy toast unchanged.

**MRR score**: Medium-low — correctness guard on a tracking feature artists rely on for campaign attribution; effort=XS-S.<br>**Shape/model**: small well-spec'd; Kimi/DeepSeek.

itstimwhite:
## Problem
UTM Builder marks Source as required but never enforces it — Copy/Open happily emit the bare profile URL with zero UTM params, and the inputs are hand-rolled `<input>`s instead of the canonical primitive.

## What changed
- **Copy-state guard:** `Copy link` and `Open` are disabled until `utm_source` is non-empty; `handleCopy` also early-returns defensively. A plain-language inline rule ("Add a source (like instagram) to create your tracking link.") renders in a fixed-height `aria-live` slot so appearing/clearing never reflows the dialog (layout-shift rule).
- **Canonical input primitive:** raw `<input>` elements swapped for `Input` from `@jovie/ui` (`packages/ui/atoms/input.tsx`), dropping the inline utility-class soup. Explicit `id` is preserved so the existing `Label htmlFor` association (and `getByLabelText` tests) keep working.
- **Regression test** added to `apps/web/tests/unit/profile/UtmBuilderDialog.test.tsx`: buttons disabled + rule named with empty Source, copy blocked, both re-enabled and message cleared once a source is entered.
- Live URL preview, copy toast, and Open behavior otherwise unchanged. No scope expansion into the share-menu auto-UTM work (#12315).

## Assumptions
- "Disable Copy/Open" is read as gating **both** actions (issue title says copy-state guard; body says "Disable Copy/Open"), since opening a source-less tracking link is the same silent no-op.
- Kept the existing app-level `Label` atom rather than the Input's built-in `label` prop — minimal diff, preserves the required-asterisk treatment.
- Input primitive's default `md` size replaces the previous hand-rolled `h-9` styling — that token drift toward the canonical atom is the point of the issue.

## Verification
Local gate could not run: sandbox egress to registry.npmjs.org is 403-blocked (network-access request filed and waited on; not granted within the task window), so `pnpm install` is impossible in this environment.

```
$ curl -s -o /dev/null -w "%{http_code}" https://registry.npmjs.org/pnpm
403
```

**Verification deferred to CI** — required lanes cover the full local gate: `ci-fast` (typecheck+biome), `Unit Tests` (`pnpm --filter=@jovie/web run test:fast`, includes the updated UtmBuilderDialog.test.tsx), `Build (public routes)`. Merge-queue only lands on green.

bug-to-test: regression test included (see above).

## Links
- Fixes #13223
- Dispatch: hyperagent thread cmr8yfd6c2doc06adxqfbkn9s


itstimwhite:

PR #13510, opened 2026-07-07
ci: retry production Lighthouse on transient DevTools protocol timeouts

itstimwhite:
## Problem

The `ci-lighthouse-production` job fails intermittently with `PROTOCOL_TIMEOUT` — Chrome's DevTools protocol doesn't respond within the default timeout during performance trace collection against live production (`jov.ie`).

The same flake was already mitigated for PR Lighthouse runs via `run-public-lighthouse.ts` which retries up to 3 times with a 5s cooldown. But the production job called `lhci autorun` directly with no retry.

## Fix

Wrap the production `lhci autorun` in a bash retry loop (3 attempts, 10s cooldown). Perf regressions are deterministic — they fail every attempt — so retrying doesn't mask real issues. Only transient protocol timeouts get absorbed.

## Risk

Low. Job already uses `continue-on-error: true` (informational only). Retry adds at most 30s of wall-clock time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5258,4 +5258,21 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: ./.github/actions/setup-node-pnpm
       - name: Run Lighthouse CI against production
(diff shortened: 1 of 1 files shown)


itstimwhite:
Superseded by #13513 (combined fix)

PR #14081, opened 2026-07-12
test(connectors): real-route coverage for suggested-action approve (JOV-4220)

itstimwhite:
## Summary

Adversarial audit found the existing "integration test" for the suggested-actions approve route **reimplements the CAS logic inline and asserts against its own reimplementation** — it would not fail if the real route diverged. Worse: `tests/integration/**` is excluded from every vitest config wired into package scripts, so it **never runs in CI at all** (JOV-4226 tracks retirement/rewiring).

New suite imports the **real** `POST` handler (9 tests):
- 401 with zero downstream calls
- CAS approve: exact `set`/`where` (id + userId + `status='pending'` predicate) + `returning` shape; workflow enqueued with approvalId + payload; inbox decision recorded; `Cache-Control: no-store`
- CAS miss → orphan-recovery mapping pinned: `enqueued`→`approved-recovered`, `already-queued`→`approved-pending-enqueue`, `not-found`→404, `not-accepted`→409
- Enqueue-throws-after-CAS: fail-closed 500 with `captureError`; the approved-but-not-enqueued row is intentionally left for retry/nightly reconcile (matches the route's design note — now tested)
- CAS update throws: 500, recovery/enqueue never invoked

## Verification
- 9/9 green (plus the legacy integration test unchanged); typecheck + biome clean
- **Frontier gate 6/6 killed**: dropped pending-predicate, dropped user scoping, skipped enqueue, swapped recovery mapping, 404-for-409, recovery-on-CAS-hit — each by its named assertion

## Notes
- Per loop policy, no CHANGELOG entry. bug-to-test: N/A — tests-only. Cost impact: none. No UI change.


itstimwhite (commented):

PR #14356, opened 2026-07-14
fix(test): honor profile card content box

itstimwhite:
## Summary

Correct the public profile mobile release-card regression fixture so it measures artwork against the padded content box that EntityCard actually defines.

## Exact failure and root cause

- Affected check: Public Profile Mock Home Release Card Layout, test "renders a stable bento release card".
- Failure: the fixture required the artwork image to be at least outer card width minus 2px. On the 224px pearl card, the rendered image is 196px.
- Root cause: the fixture compared the image to the 224px outer card instead of the intentional padded content box. EntityCard contributes 12px horizontal padding plus borders; the artwork frame is 198px and its image content is 196px.
- Classification: broken test or fixture contract. This is not a PR-specific product regression, flaky CI, dependency/environment drift, merge conflict, or recurring Gem failure.

## Fix

Capture the artwork parent content width and compare the image width to that computed content box with a 2px border tolerance. Existing assertions continue to cover horizontal overflow, hero/card separation, title containment, and compact cover height.

## Bug-to-Test

bug-to-test: satisfied

Regression test: apps/web/tests/e2e/profile-mobile-viewport-stability.spec.ts

## Test Coverage

Test-only diff. No application code paths changed.

## Verification

- Focused Playwright, iPhone SE and iPhone 13/14: 2 passed.
- EntityCard Vitest suite: 9 passed.
- Biome on the changed fixture: passed, no fixes.
- Web typecheck with Node 22.22.1 and pnpm 9.15.4: passed.
- Bug-to-test gate: satisfied.
- Repository pre-push hook: proxy guard, Tailwind guard, typecheck, affected verify, lint, CI harness validation all passed.
- Diff check: clean.
- Commit: signed SSH commit 719946ef514c6a605bac28f08d1f19399e3f883d.

## Pre-Landing Review

Primary review approved the one-file fixture correction. No findings.

## Design Review

design-canonical: not applicable. This changes test measurement only, not rendered UI.

## Eval Results

No prompt-related files changed; evals skipped.

## Scope Drift

Scope Check: CLEAN. One test fixture file changed.

## Plan Completion

No separate plan file. The assigned fixture correction, regression proof, rebase, signing, and draft PR handoff are complete.

## Linear follow-ups

Linear follow-ups: none.

## Verification Results

11 assertions passed across the focused Playwright and EntityCard suites. Required static and repository hooks passed.

## Test plan

- [x] Reproduce the old 196px versus 224px failure.
[...]

PR #14369, opened 2026-07-14
fix(e2e): provision persisted auth actors

itstimwhite:
## Root cause

PR #14356 Mobile Overflow Guard (320px), job 87011291751, failed because the E2E helper wrote synthetic Better Auth labels such as user_test and ba_dev_creator-ready directly into bypass cookies. When no persisted app user was resolved, the fallback session exposed that label as the app user ID; downstream users.id UUID queries failed with PostgreSQL 22P02. All seven authenticated ready selectors then timed out.

Classification: broken E2E fixture and environment drift. This is deterministic and non-retryable until the actor is provisioned; it is not PR-specific and not flaky CI.

## Smallest safe fix

- Provision persona sessions through POST /api/dev/test-auth/session instead of writing arbitrary identity cookies.
- Preserve the persona-less legacy path only for server-validated persisted Better Auth users and fail closed with 404 when absent.
- Add Gem diagnosis for 22P02 plus user_test or ba_dev_* as a non-retryable broken E2E fixture, with persisted-session remediation.
- Map this exact repair signature to focused mandatory pre-push coverage.

## Verification

- Auth helper, route, and persisted-actor tests: 47/47 passed.
- Gem classifier and affected-selector tests: 72/72 passed.
- Normal selected pre-push lane: 48/48 passed.
- Exact authenticated app-home 320px Playwright path: 1/1 passed.
- The initial seven-route local run showed every session POST returning 200 and no 22P02; local cold Turbopack compilation exceeded the route timeout before surfaces completed.
- Web typecheck, Biome, ESLint, Tailwind guard, CI harness, signed commit, diff check, and bug-to-test passed.

bug-to-test: satisfied

### DIFF
--- a/apps/web/app/api/dev/test-auth/session/route.ts
+++ b/apps/web/app/api/dev/test-auth/session/route.ts
@@ -5,6 +5,7 @@ import {
   buildDevTestAuthCookieDescriptors,
   DEV_TEST_AUTH_COOKIE_NAMES,
   ensureDevTestAuthActor,
+  ensureExistingDevTestAuthActor,
   getCachedDevTestAuthSession,
   getDevTestAuthAvailability,
   parseDevTestAuthPersona,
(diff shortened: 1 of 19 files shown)


itstimwhite (commented):

